PR #71, opened 2026-07-14
jnigen: tag-bit handle lifecycle — immutable lock-order key + closed-handle guards

milyin:
Fixes #34, #35 — the two P0s from the 2026-07-14 JniGen API review, as one lifecycle redesign.

## Design: tag-bit closed state

Closed state moves into **bit 0 of `NativeHandle.ptr`**: `close()`, `take()`, and every consuming wrapper's `finally` set `ptr = p or 1L` instead of zeroing. The address bits are therefore **write-once** for the object's lifetime:

- **#35 (lock-order inversion)**: the N-ary lock ordering key becomes the masked address `ptr and -2L` — immutable per object, in all three lock paths (list `sortedBy`, 2-arity, 3-arity network). A concurrent `close()` only sets bit 0 and can't move any handle across the sort order. (The original issue's premise needed one correction: the flaw was the key's *mutability*, not pointer-ness — and the mutation fired on **every consuming call** via `ParamMode::Consume`, not just close/take. Pointer-derived ordering survives once the key is frozen.)
- **#34 (closed-race UB)**: a handle closed between the pre-lock guard and the native call now carries an *odd* pointer instead of a 0 that the nullable niche can't distinguish from "absent". Every Rust path that dereferenced a raw `jlong` — required borrow converter, by-value consume prelude, owned-handle expansion leaves, required data-class handle fields (the last one also fixes a pre-existing `Box::from_raw(0)` null-deref) — rejects `*v == 0 || (*v & 1) == 1` and routes `"Operation on a closed native handle."` through the normal `onError` channel. `Option`-consume keeps `0` = `None` but treats tagged as present-but-closed = error. `freePtr` ignores tagged pointers.

Bit 0 is guaranteed free by a generated per-opaque compile-time check `align_of::<T>() >= 2`, spelled as `const _: () = { if … { panic!() } }` rather than `assert!` — the `qualify_item` pass doesn't descend into macro token streams, so a type inside `assert!` would stay unqualified.

Kotlin surface: `isClosed()` = `ptr == 0L || (ptr and 1L) != 0L`; `peek()` normalizes closed → `0` (single volatile read); pre-lock guards use `isClosed()`; kdocs state the real contract (the "callers re-read and null-check inside body" claim from #34's evidence is gone).

Documented residual caveat (in the `NativeHandle` kdoc): sort-key ties exist only between same-address twins (`take()`, allocator address reuse); the older twin is durably closed before the newer exists, so tied pairs can't be locked together in correctly synchronized programs. A construction-counter id would make ties structurally impossible at the cost of an extra field per handle; rejected in favor of zero-cost pointer semantics.

## Tests

- **covertest-kotlin storm section** (acceptance for both issues): 4 reader threads hammering the 3-handle `storageTotalLen` over a shared `AtomicReferenceArray` pool while 2 stormers `close()` / `take().close()` / double-`close()` and swap in fresh handles, 60 s watchdog. No deadlock, no crash; closed-during-call races observed (asserted > 0) and all surfaced via `onError`. `take().close()` deliberately manufactures masked-key ties.
- Generator snapshot tests pin the tagged close/take, masked comparators, `isClosed()` guards, Rust converter guards (`if*v==0||(*v&1)==1`), the align check, and tagged-aware `freePtr`.
- `cargo test --all --all-features` green (237 lib tests); regeneration byte-deterministic; committed goldens regenerated (covertest + perftest; cbindgen output untouched).

### DIFF
--- a/examples/covertest-kotlin/kotlin/generated/io/prebindgen/covertest.kt
+++ b/examples/covertest-kotlin/kotlin/generated/io/prebindgen/covertest.kt
@@ -11,23 +11,39 @@ import io.prebindgen.covertest.model.Stamp
  * Base class for every typed native handle: owns the raw `Box<T>` pointer
  * slot and its monitor. Subclasses add their type-specific `close()` /
  * `take()` / `freePtr`.
+ *
+ * Lifecycle is a tag bit: `Box` pointers are at least 2-aligned (asserted
+ * on the Rust side), so bit 0 is free — closing/consuming sets `ptr = p or 1`
+ * instead of zeroing. The address bits (`ptr and -2`) are therefore
+ * write-once for the object's whole lifetime, which is what makes them a
+ * sound lock-ordering key (a mutable key could reorder concurrent lock
+ * acquisition and deadlock). All `ptr` writes happen under this handle's
+ * monitor.
  */
 public abstract class NativeHandle(initialPtr: Long) : AutoCloseable {
     @Volatile internal var ptr: Long = initialPtr
 
-    public fun peek(): Long = ptr
+    /** The live pointer, or `0` if this handle is closed. */
+    public fun peek(): Long {
+        val p = ptr
+        return if (p == 0L || (p and 1L) != 0L) 0L else p
+    }
 
-    public fun isClosed(): Boolean = ptr == 0L
+    public fun isClosed(): Boolean = ptr == 0L || (ptr and 1L) != 0L
 }
 
 /**
- * Acquire every handle's monitor in one global order (sorted by raw
- * pointer) so concurrent calls touching the same handles can't deadlock,
- * then run [body]. Closed handles (`ptr == 0`) are still locked; callers
- * re-read and null-check each pointer inside [body]. Scales to any arity.
+ * Acquire every handle's monitor in one global order — sorted by the
+ * immutable address bits (`ptr and -2`; bit 0 is the closed tag and
+ * never participates) — so concurrent calls touching the same handles
+ * can't deadlock, then run [body]. The key never changes after
+ * construction: closing only sets bit 0, so a concurrent `close()`
+ * can't reorder anyone's acquisition. Closed handles are still locked;
+ * their tagged pointers are rejected by the Rust-side converter guard
+ * inside the native call. Scales to any arity.
  */
 internal fun <R> withSortedHandleLocks(handles: List<NativeHandle>, body: () -> R): R {
-    val sorted = handles.sortedBy { it.ptr }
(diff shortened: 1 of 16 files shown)


milyin:
## Performance: fix vs pre-fix (perftest-kotlin)

Measured the JNI wrapper overhead of the tag-bit lifecycle with `perftest-kotlin` (N = 5,000,000, VEC_N = 16, macOS/aarch64). Protocol: **ABBA interleaving** — main×2 → fix×4 → main×2, each block preceded by a full rebuild plus settle pause, median of 4 runs per side. (A naive block comparison — 3 fix runs then 3 main runs — showed a phantom 20–35% "speedup" from machine-state drift between time blocks; interleaving cancels it.)

| op | variant | main med ns/op [min..max] | fix med ns/op [min..max] | Δ median |
|---|---|---|---|---|
| put | native.null | 38.5 [33..46] | 34.5 [33..36] | −10.4% |
| put | native.str | 672.0 [634..897] | 664.0 [652..679] | −1.2% |
| get | native.null | 278.0 [233..324] | 250.0 [235..253] | −10.1% |
| get | native.str | 641.5 [552..721] | 575.0 [547..766] | −10.4% |
| callback | native.null | 352.5 [323..403] | 367.0 [338..384] | +4.1% |
| callback | native.str | 666.5 [599..895] | 654.0 [601..699] | −1.9% |
| put_vec | batch.null | 265.5 [243..457] | 243.5 [236..347] | −8.3% |
| put_vec | batch.str | 10860.5 [10238..12074] | 11881.0 [10835..12017] | +9.4% |
| get_vec | batch.null | 4145.5 [3902..7427] | 3900.0 [3645..4227] | −5.9% |
| get_vec | batch.str | 10187.5 [8854..13149] | 9466.0 [9058..10081] | −7.1% |
| callback_vec | batch.null | 7194.0 [6665..9233] | 6842.0 [6606..7182] | −4.9% |
| callback_vec | batch.str | 12153.5 [11575..12811] | 11103.0 [10810..11532] | −8.6% |

**Conclusion: no measurable regression.** Deltas scatter −10%…+9% and the per-side min..max ranges overlap on every row — run-to-run noise (JVM + macOS P/E-core scheduling) dominates. That matches expectation: the fix adds per call at most one AND+compare in the pre-lock guard (`isClosed()`), one AND per lock-order comparison (multi-handle calls only), and two integer compares in the Rust converter per handle parameter — single-digit nanoseconds against a ≥35 ns JNI floor. The compile-time `align_of` check and the tag writes in `close()`/`take()`/consume are off the hot path.

PR #77, opened 2026-07-14
jnigen: generated <Name>Api interface — checked injection hatch on all class kinds (#54)

milyin:
Fixes #54 (P2 from the 2026-07-14 JniGen API review).

## The mechanism: a generated `<Name>Api` interface + orthogonal `.implements`

Every class decl — `ptr_class!` / `data_class!` / `value_class!` / `enum_class!` — gets two independent switches:

- **`.interface()`** emits a generated `<Name>Api` interface mirroring the class's public surface (inherited `NativeHandle` members, ctor properties, and declared `.fun` members) and makes the class implement it, applying `override` automatically to class-body members. Implied by `.interface_name(...)`.
- **`.implements(x)`** makes the class nominally implement the user interface `x` (dotted FQNs imported + shortened; same-package names work bare). It does **not** enable `.interface()` — technically orthogonal.

Used together they form the **checked-injection** story: the user's interface *extends* the generated one, so its default members call the class's real generated members under full compiler checking — no guessed prototypes, no hand-replicated signatures, no editing of generated code.

```rust
ptr_class!(Storage).interface().implements("io.prebindgen.covertest.CovResource")
data_class!(Payload).interface().implements("io.prebindgen.covertest.Timestamped")
enum_class!(Priority).interface_name("PriorityKind").implements("io.prebindgen.covertest.Ranked")
```

```kotlin
// generated
public interface StorageApi : AutoCloseable { fun peek(): Long; fun isClosed(): Boolean; fun take(): Storage; fun len(…): Long; … }
public class Storage(initialPtr: Long) : NativeHandle(initialPtr), StorageApi, CovResource { public override fun len(…) … }

// hand-written SDK — extends the generated interface, defaults over real members
interface CovResource : StorageApi {
    fun isEmpty(): Boolean = len(boom) == 0L      // class-specific generated member
    val live: Boolean get() = !isClosed() && peek() != 0L
}
```

## Interface naming

Per-decl `.interface_name("StorageContract")` (relative, mangle-bypass, like `.name()`), or the global `set_interface_name_mangle(|class_name| …)` hook (receives the **final class name**; **identity is a hard error** — a class and its interface can't share a name in one package; unset default appends `"Api"`).

## Design changes vs the interim sketch

- **`.overrides()` member marker reverted** — it forced consumers to guess generated prototypes and inverted the hierarchy. The generated-interface approach makes the class the source of truth and the compiler the checker.
- **`.kotlin_type()` removed entirely** (all four decls) — its replace-the-class hatch is subsumed by augment-only `.interface()`. No consumer used it. With its only producer gone, `NameSpec` collapses from an enum to a struct (the `Verbatim` arm had no other caller). `NameSpec::Verbatim` for scalar/generic wrappers as mentioned in docs did not actually exist in code.
- The issue's other option — a configurable base *between* `NativeHandle` and the generated class — remains deliberately NOT added: it threads consumer code into the lifecycle chain. Interface defaults cover the need without that risk.

## Acceptance

- covertest exercises the hatch on **three kinds** on the JVM: `Storage` (ptr) via `CovResource : StorageApi`, `Payload` (data) via `Timestamped : PayloadApi`, `Priority` (enum) via `Ranked : PriorityKind` — each SDK interface extends the generated `<Name>Api`, defaults call class-specific generated members, used polymorphically. **`./gradlew run`: PASS, 26 sections.**
- The fourth path (value_class interface) is covered by a generator unit test (`value_class_interface_emits_generated_api`).
- Generator unit tests: implements-only supertypes (no generated interface, no `override`); `.interface()` generated-interface emission + auto-`override`; per-decl `.interface_name` + `set_interface_name_mangle` hook (asserts input = class name); identity-result rejection; duplicate-implements rejection.
- `cargo test -p prebindgen --lib`: **253 passed.** CI pre-flight clean (fmt CI config, clippy `--deny warnings`, example-cbindgen goldens restored). `regen-check.sh`: participating classes changed as designed, non-participating byte-identical, Rust side unchanged.

### DIFF
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -254,17 +254,13 @@ fn main() {
                 )
                 .class(
                     ptr_class!(Storage)
-                        // #54: `.overrides()` — the member implements the
-                        // matching CovResource abstract (Kotlin needs the
-                        // `override` modifier on class-body members).
-                        .fun(fun!(storage_len).overrides())
+                        .fun(fun!(storage_len))
                         .fun(fun!(storage_contains))
                         .constructor(fun!(storage_with_payload))
                         // #54 integration hatch: the generated class joins a
-                        // hand-written consumer interface — abstracts satisfied
-                        // by NativeHandle's inherited peek()/isClosed() AND the
-                        // `.overrides()`-marked len() — no hand-editing of
-                        // generated code.
+                        // hand-written consumer interface (satisfied by
+                        // NativeHandle's public peek()/isClosed()) — no
+                        // hand-editing of generated code.
                         .implements("io.prebindgen.covertest.CovResource"),
                 )
                 // The callback-handler handles (single payload / whole batch / owned
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -254,13 +254,17 @@ fn main() {
                 )
                 .class(
                     ptr_class!(Storage)
-                        .fun(fun!(storage_len))
+                        // #54: `.overrides()` — the member implements the
+                        // matching CovResource abstract (Kotlin needs the
+                        // `override` modifier on class-body members).
+                        .fun(fun!(storage_len).overrides())
                         .fun(fun!(storage_contains))
                         .constructor(fun!(storage_with_payload))
                         // #54 integration hatch: the generated class joins a
-                        // hand-written consumer interface (satisfied by
-                        // NativeHandle's public peek()/isClosed()) — no
-                        // hand-editing of generated code.
+                        // hand-written consumer interface — abstracts satisfied
+                        // by NativeHandle's inherited peek()/isClosed() AND the
+                        // `.overrides()`-marked len() — no hand-editing of
+                        // generated code.
                         .implements("io.prebindgen.covertest.CovResource"),
                 )
                 // The callback-handler handles (single payload / whole batch / owned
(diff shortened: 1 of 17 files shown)


milyin:
Added in 2af942f: **`.overrides()`** — the marker that makes the hatch work for **class-specific** interfaces.

Kotlin accepts an inherited base-class member (`NativeHandle`'s `peek()`/`isClosed()`) as an interface implementation implicitly, but requires the `override` modifier on **class-body** members — verified empirically: an interface declaring `fun len(onError: JniErrorHandler<Long>): Long` failed against the generated class with *"'len' hides member of supertype 'CovSized' and needs 'override' modifier"*. Without a marker, injected interfaces could only abstract over the shared `NativeHandle` surface, which makes them near-useless — the point of injection is reaching the class's own API.

```rust
ptr_class!(Storage)
    .fun(fun!(storage_len).overrides())   // emits: public override fun len(…)
    .implements("io.prebindgen.covertest.CovResource")
```

- Valid only on `ptr_class` instance methods (the marker pairs with `.implements(...)`): free package functions, companion constructors, and data/value-class members hard-error, as do all five non-surfacing decoration seams (ignore / variant / field / const source / convert source) — the #36 one-policy rule.
- covertest acceptance extended: `CovResource` now declares the class-specific `len(onError)` abstract (satisfied by the marked member) plus an `isEmpty()` default built on it, driven through a polymorphic interface reference on the JVM — **PASS, 26 sections**. Golden diff: `public fun len` → `public override fun len`, one line.
- 253 lib tests; CI pre-flight clean; regen-check PASS.

PR #83, opened 2026-07-17
perftest: twin-handle lifecycle benchmark — plain vs .gc_managed()

milyin:
## What

Measures the real cost of the `.gc_managed()` machinery (#82) with a twin minimal handle pair — `Token` (plain) vs `TokenGc` (`.gc_managed()`), identical Rust shape — plus six new perftest rows.

## Results (medians of 3 runs, N=5M classic / 2M handle rows, Apple Silicon)

| row | plain | gc | delta |
|---|---|---|---|
| `handle_close` — create + explicit close | 34 ns | 119 ns | **+85 ns** |
| `handle_drop` — create + drop, no close | 20 ns* | 171 ns | **+151 ns** |
| `handle_call` — method on a live handle | 18.9 ns | 18.9 ns | **0** |

\* plain drop leaks natively (allocation cost only — today's leak-on-drop behavior); the gc number includes the Cleaner-driven free and PhantomRef/GC churn.

**Existing hot paths are unaffected**: pre-#82 (`598f740`) vs this branch, medians of the untouched `put`/`get`/`callback`(+`_vec`) rows differ only within noise (±3%, mixed signs) — the `NativeHandle.ptr` open-property change costs nothing for plain classes, and the atomic-cell getter matches the plain field on the call path.

## Interpretation for the zenoh hot path

A received zenoh sample materializes 2 handles (KeyExpr + payload ZBytes). GC-managing them would add roughly 85–150 ns × 2 ≈ **0.2–0.3 µs per message** of allocation/registration/cleanup work against the ~1 µs/msg subscriber floor — i.e. a potential double-digit percentage throughput cost, to be confirmed end-to-end with `run_thr_java.sh` (follow-up measurement). The per-class opt-out for KeyExpr/ZBytes in zenoh-flat-jni stays justified; the zero call-path cost confirms the backstop classes (Session, Publisher, …) pay only at create/close time.

265 lib tests green; fmt + clippy clean; covertest untouched.

### DIFF
--- a/examples/perftest-flat/src/lib.rs
+++ b/examples/perftest-flat/src/lib.rs
@@ -273,6 +273,45 @@ pub fn string_len(s: &String) -> usize {
     s.len()
 }
 
+/// Minimal opaque handle for LIFECYCLE benchmarking: the plain variant of a
+/// twin pair whose only difference is the Kotlin-side `.gc_managed()` flag —
+/// benchmarked head-to-head to price the GC-cleanup machinery (atomic cell,
+/// Cleaner registration, CAS release ticket) per handle. A boxed handle like
+/// [`Storage`] (not `#[prebindgen]`, not `#[repr(C)]`).
+pub struct Token {
+    value: i64,
+}
+
+/// Create a plain benchmark token.
+#[prebindgen]
+pub fn token_new(value: i64) -> Token {
+    Token { value }
+}
+
+/// Read a plain benchmark token.
+#[prebindgen]
+pub fn token_value(t: &Token) -> i64 {
+    t.value
+}
+
+/// GC-managed twin of [`Token`] — identical shape and cost on the Rust side;
+/// the Kotlin binding declares this one `.gc_managed()`.
+pub struct TokenGc {
+    value: i64,
+}
+
+/// Create a gc-managed benchmark token.
+#[prebindgen]
+pub fn token_gc_new(value: i64) -> TokenGc {
+    TokenGc { value }
+}
+
+/// Read a gc-managed benchmark token.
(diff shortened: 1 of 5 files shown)


milyin:
## Follow-up: end-to-end zenoh throughput measurement

`run_thr_java.sh` (`CROSS=0`, 3 passes × 8 rounds × 100k msgs, loopback peer-peer), flat×flat subscriber medians, with `KeyExpr` + `ZBytes` temporarily flipped to `.gc_managed()` in zenoh-flat-jni (the two per-message handle classes — one of each is materialized per received sample; the payload handle is explicitly closed after its lazy read, the keyexpr handle is dropped and Cleaner-freed):

| payload | baseline (plain) | gc_managed | delta | per-msg cost |
|---|---|---|---|---|
| 8 B | 891,880 msg/s | 685,624 msg/s | **−23.1%** | +337 ns |
| 1 KB | 652,189 msg/s | 562,413 msg/s | **−13.8%** | +245 ns |

(main×main control column stayed at ~490k/~430k msg/s across both runs — machine stable.)

The observed +245–337 ns/msg matches the microbench prediction (~85 ns close-path + ~151 ns drop-path ≈ 236 ns, plus GC-pressure amplification from the PhantomRef churn at ~700k registrations/s). Conclusion: **the per-class opt-out for hot-path per-message handles (KeyExpr, ZBytes) is justified** — gc-managing them costs a double-digit throughput percentage; the backstop classes (Session, Publisher, …, Encoding) pay only at create/close time, and nothing on the call path. The temporary flip was reverted; zenoh-flat-jni is unchanged.

PR #84, opened 2026-07-17
jnigen: binding-local functions — one vocabulary for custom locally-defined members

milyin:
## What

**Binding-local functions**: any fn the *binding crate* defines can back generated output fields, free functions, instance methods, companion constructors, and `convert!` conversion sources — no `#[prebindgen]` source-crate item needed. One vocabulary for all of them:

```rust
// A fn is declared by a ≥2-segment PATH plus its stated signature —
// named params become the foreign-side parameter names, and a
// Result<_, E> return IS the error channel (Err routes to onError):
fun!(crate::summary_describe).sig(sig!((s: &Summary, verbose: bool) -> String))

// usable in every position:
.fun(fun!(crate::summary_describe).sig(sig!((s: &Summary, verbose: bool) -> String)))
.method(fun!(crate::summary_mean).sig(sig!((s: &Summary) -> f64)))
.constructor(fun!(crate::summary_from_mean)
    .sig(sig!((count: i64, mean: f64) -> Result<Summary, String>)))

// output field (the receiver explicit, checked against the expanded type):
.expand(expand_return!(Encoding)
    .field(fun!(encoding_get_id))
    .field(fun!(crate::encoding_if_schema)
        .sig(sig!((e: &Encoding) -> Option<&Encoding>))
        .name("handle")))

// convert! source (trait macros from!/try_from!/into!/try_into! stay TRAIT-only):
.convert(convert!(Label)
    .input(fun!(crate::label_in).sig(sig!((s: String) -> Result<Label, String>)))
    .output(fun!(crate::label_out).sig(sig!((l: Label) -> String))))
```

A bare-ident `fun!(f)` still declares a `#[prebindgen]` registry fn, unchanged. Conditional delivery — an `Option<&Self>` field that is null when a binding-side predicate declines (zenoh's "Encoding handle only when schema-carrying", ZettaScaleLabs/zenoh-flat-jni#6) — is one use among many.

## Mechanism

A new `Prebindgen::local_functions()` hook feeds a synthesis pre-pass at the top of `Registry::resolve`: stated signatures become ordinary registry entries (body never emitted; origin = the path's module prefix, driving the existing `origin_module` call qualification — the generated file compiles inside the binding crate, so `crate::…` resolves). After synthesis a binding-local fn **is** a registry fn: converters, expansion defaults, the method receiver rule, `.split_on_param`, `Result` fallibility, and `expand_param!` variant arms (referenced by ident) all apply unchanged.

**Naming is fully uniform** — `.name()` is never obligatory: the default derivation feeds the mangle hooks the camel-cased **last path segment** with the same package/class context as a registry fn; fields follow the field precedence (override → class-member inheritance → camel-cased segment). Emission detail: an `Option<&T>` field flattens as a **nullable identity leaf** (final-Option unwrap → handle-projection clone → boxed `Long?`/null wire), and a self-typed field degrades to a leaf instead of the splice-cycle error — which is what enables conditional delivery.

Hard errors: single-segment paths, missing `.sig`, `.sig` on a registry fn, ident collisions with `#[prebindgen]` items or between locals with different signatures, path-built `fun!` in variant position (declare first, reference by ident).

## Deleted (0.5 no-shims policy)

Two predecessor vocabularies for the same concept:
- `field!("name").with(ty!, path!)` + `FieldDecl`/`FieldSourceKind`;
- `convert!` sources' `.with(path!)`/`.error(ty!)` + `ConvertSpec::LocalFn` (the sig's `Result` replaces `.error` — it existed only because a bare path carried no signature).

## Tests

- Lib (**272 green**): fields (root, spliced, wire-shape boxed/null), functions (all positions, path-qualified calls, expansion default on a local param, local ctor as variant arm), convert sources (fallible + infallible via sig), the mangling proof (custom hooks receive the camel-cased last segment), and all the hard errors.
- Covertest (**29/29 JVM sections**, every-feature rule): conditional-handle field, `describeSummary`, `Summary.mean()`/`Summary.fromMean` — both **without `.name()`** (the strip-class-prefix hook derives the names automatically, generated output byte-identical), the fallible `fromMean(-1, …)` → `onError` arm, and the Label convert round-trip on the unified form.
- fmt + clippy clean on stable and MSRV 1.85.

Companion PR: ZettaScaleLabs/zenoh-flat-jni#6 (consumes the conditional-handle field; migrated to the final syntax, regen byte-identical; green once this merges).

### DIFF
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -46,7 +46,9 @@
 //! | per-fn `.expand_return(…)` identity-only | `storage_summary_handle` / `archive_latest` (raw handle return) |
 //! | per-fn `.expand_param(name, …)` variants | `storage_expect_summary` |
 //! | per-fn `.expand_return(…)` fields+self | `storage_summary_full` |
-//! | binding-local field `field!` `.with(ty!, path!)` | `storage_summary_probe` — conditional handle via `crate::summary_if_nonempty` |
+//! | binding-local field `field!` `.with(ty!, path!)` | `storage_summary_probe` — custom field, here a conditional handle via `crate::summary_if_nonempty` |
+//! | binding-local fn `fun!(crate::…)` `.sig(sig!)` as free fn | `describeSummary` ← `crate::summary_describe` |
+//! | binding-local fn as `.method()` / `.constructor()` | `Summary.mean()` ← `crate::summary_mean`; `Summary.fromMean` ← `crate::summary_from_mean` |
 //! | `Result<_, E>` → `onError`           | `storage_try_with_label` |
 //! | `Option<T>`                          | `Option<Payload>` (in + out) / `Option<Vec>` / `Option<i64>` / `Option<enum>` (param + return + field) |
 //! | `impl Fn` callbacks (single + slice) | `payload_handler_new` / `payload_vec_handler_new` |
@@ -88,7 +90,7 @@
 
 use prebindgen::{
     constant, convert, core::Registry, data_class, enum_class, expand_param, expand_return, expr,
-    field, from, fun, into, lang::JniGen, matching, package, path, ptr_class, try_from, ty,
+    field, from, fun, into, lang::JniGen, matching, package, path, ptr_class, sig, try_from, ty,
     value_class,
 };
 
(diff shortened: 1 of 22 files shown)


milyin:
Added the covertest exercise per the every-feature-demonstrated rule (`a19d091`): `storage_summary_probe` carries a per-fn field set of `(count, total)` plus `field!("handle").with(ty!(Option<&Summary>), path!(crate::summary_if_nonempty))` — the binding-side predicate lives in covertest-kotlin's `src/lib.rs` next to the `label_in`/`label_out` convert!-with precedent. The new JVM harness section asserts both arms end to end: an empty summary arrives **value-only** (null handle, no native clone), a non-empty one delivers a **live owned handle** alongside the decomposed values. 28/28 sections pass; feature table updated.

milyin:
Third commit (`fd84866`): **convert! unified onto the one binding-local vocabulary.** `try_from!(…).with(path!(…)).error(ty!(…))` and `ConvertSpec::LocalFn` are deleted (0.5 no-shims policy) — a binding-local conversion source is now the same decl as everywhere else:

```rust
.convert(convert!(Label)
    .input(fun!(crate::label_in).sig(sig!((s: String) -> Result<Label, String>)))
    .output(fun!(crate::label_out).sig(sig!((l: Label) -> String))))
```

The sig's `Result<_, E>` **is** the error channel (`.error` existed only because a bare path carried no signature); the source joins the `local_functions()` synthesis pre-pass and lowers through the ordinary registry-fn path. `from!`/`try_from!`/`into!`/`try_into!` remain for trait sources only.

Also: synthesized local fns now join the helper set (the spurious `skipping undeclared … summary_if_nonempty` warning is gone); a path-built `fun!` in variant/field positions is a hard error with guidance; covertest migrates Label to the unified form (Kotlin surface byte-identical) and demonstrates **Result-in-sig!** with a now-fallible `Summary.fromMean` (negative count → `onError`). 29/29 sections; 271 lib tests; MSRV clippy clean.

PR #207, opened 2026-07-27
jnigen: give byte-backed Kotlin values content equality

milyin:
Reported on [ZettaScaleLabs/zenoh-flat-jni#16](https://github.com/ZettaScaleLabs/zenoh-flat-jni/pull/16): generated values with equal contents compare unequal.

```kotlin
Timestamp(1uL, byteArrayOf(1, 2, 3)) == Timestamp(1uL, byteArrayOf(1, 2, 3))  // false
EntityGlobalId(ZenohId(byteArrayOf(1)), 7) == EntityGlobalId(ZenohId(byteArrayOf(1)), 7)  // false
```

A generated class mirrors a Rust type deriving `PartialEq`/`Eq`, so this is the generator contradicting the type it mirrors — not a binding bug.

## What Kotlin actually does

Measured on Kotlin 1.9.0 before writing any code:

| | `==` | `hashCode` equal |
|---|---|---|
| `@JvmInline value class V(val b: ByteArray)` | ❌ false | ✅ true |
| `data class D(val n: ULong, val id: ByteArray)` | ❌ false | ✅ true |
| `data class` holding the value class | ❌ false | ✅ true |
| plain class + explicit content ops | ✅ true | ✅ true |

Kotlin is **inconsistent**, not uniformly identity-based: a data class's generated `hashCode`/`toString` *do* special-case arrays (`contentHashCode`/`contentToString`) while its `equals` does not. That is why the report saw `eq=false hashEq=true` — a broken value finds the right `HashMap` bucket and is then rejected. `==` is the observable defect; a hashCode check alone would not have found it. (`hashSetOf(t1, t2).size == 2` confirms the practical damage.)

## Why value blobs lose `@JvmInline`

Kotlin 1.9 rejects the members outright:

```
e: Member with the name 'equals' is reserved for future releases
e: The feature "custom equals in value classes" is experimental and should be enabled explicitly
```

So an inline value class **cannot** be given value equality at this language level, and the experimental typed-equals replacement would force an opt-in flag on every consumer of a shared tier. Value blobs are therefore emitted as a plain `data class`.

Cost: one small allocation per crossing at the wrapper tier. **The JNI ABI is untouched** — externs already declare `ByteArray` (`external fun zenohIdToString(z: ByteArray, …)`) and the wrapper passes `.bytes` explicitly, so the erasure was never load-bearing. Value blobs are also documented as off-hot-path.

## The change

`equality::content_equality_members` emits `equals`/`hashCode`/`toString` for any class with an array-backed constructor property and **nothing** for the rest, so existing classes keep the compiler's own generation and the diff stays confined. Three emitters feed it: `data_class!` (`render.rs`), value blobs and `sealed_class!` variant payloads (`kotlin_emit.rs`).

## Coverage

Nothing exercised the broken shapes — covertest had the `Stamp` value blob but **no data class with a `ByteArray` field**, which is precisely why this reached a consumer. `BlobValue` adds both (a `Vec<u8>` field beside a scalar, plus a nested value blob) and the new section asserts content equality, hash equality, `HashSet` de-duplication, `toString`, and that *each* component participates (a comparison ignoring one would otherwise still pass).

Verified the section fails without the fix:

```
IllegalStateException: value blob must compare by content:
  Stamp(bytes=[7, 0, …]) vs Stamp(bytes=[7, 0, …])
```

`PASS - 46 sections`, 334 lib tests, fmt/clippy clean.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/equality.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/equality.rs
@@ -33,21 +33,101 @@
 
 use super::*;
 
-/// A property whose Kotlin type compares by identity and therefore needs
-/// content-aware operators.
-enum ArrayProp {
-    /// `ByteArray` / `ByteArray?`.
-    Bytes { nullable: bool },
+/// Whether a Kotlin type carries an array anywhere inside it, and therefore
+/// compares by identity unless the generated operators dig in.
+///
+/// Recursive, because a container of arrays is just as identity-compared as a
+/// bare one: `List<ByteArray>` (from `Vec<Vec<u8>>`) inherits `ByteArray`'s
+/// `equals`, so two lists of equal chunks are unequal and `toString` renders
+/// `[[B@3830f1c0]`. A container of *classes* is fine — those already compare
+/// by value — so only an array at the bottom makes a property array-bearing.
+fn array_bearing(ty: &kt::KtType) -> bool {
+    match ty {
+        kt::KtType::Named { fqn, args, .. } => {
+            fqn.rsplit('.').next().unwrap_or(fqn) == "ByteArray" || args.iter().any(array_bearing)
+        }
+        kt::KtType::Function { .. } => false,
+    }
 }
 
-/// Classify one constructor property. `None` for everything that already
-/// compares by value (scalars, `String`, enums, nested generated classes).
-fn array_prop(ty: &kt::KtType) -> Option<ArrayProp> {
-    match ty.simple_name() {
-        Some("ByteArray") => Some(ArrayProp::Bytes {
-            nullable: ty.is_nullable(),
-        }),
+/// The element type of a single-argument container (`List<T>` -> `T`).
+fn element_of(ty: &kt::KtType) -> Option<&kt::KtType> {
+    match ty {
+        kt::KtType::Named { args, .. } if args.len() == 1 => Some(&args[0]),
         _ => None,
+    }
+}
+
+/// `a == b` for one value of type `ty`, digging through containers.
+///
+/// `a`/`b` are Kotlin expressions. Nullability is handled per level: a
+/// `ByteArray?` rides `contentEquals`'s nullable-receiver overload, while a
+/// nullable container needs an explicit both-null / both-present test.
+fn eq_expr(a: &str, b: &str, ty: &kt::KtType) -> String {
+    if !array_bearing(ty) {
+        return format!("{a} == {b}");
+    }
+    if element_of(ty).is_none() {
+        // The array itself.
+        return format!("{a}.contentEquals({b})");
+    }
+    let elem = element_of(ty).expect("checked above");
+    let inner = eq_expr("__x", "__y", elem);
+    let cmp = format!(
+        "{a}.size == {b}.size && {a}.indices.all {{ __i -> \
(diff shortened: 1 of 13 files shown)


prebindgen/src/api/lang/jnigen/jni/kotlin_emit.rs
```diff
@@ -421,20 +421,33 @@ impl JniGen {
...
+                 Not a `@JvmInline value class`: Kotlin 1.9 reserves `equals`/`hashCode`\n\
+                 members on a value class, so one cannot be given the value equality this\n\
+                 type has in Rust — arrays would compare by identity.",
                 key.as_str()
             );
             let class_kdoc = crate::api::lang::jnigen::jni::source_item_doc(registry, key)
                 .map(|d| format!("{d}\n\n{framework_line}"))
                 .unwrap_or(framework_line);
-            let mut class = KtClass::new(ClassKind::ValueInline, &class_name)
+            let mut class = KtClass::new(ClassKind::Data, &class_name)
```
milyin, line 433: Changing the wrapper to a reference class also changes its JVM representation when it is stored in another generated object, but the whole-object input decoder still assumes value-class erasure. In this PR's generated `JObject_to_BlobValue`, native code does:

```rust
env.get_field(v, "stamp", "[B")
```

while the generated Kotlin field is now `val stamp: Stamp`, whose descriptor is `Lio/prebindgen/covertest/model/Stamp;`.

I reproduced this by marking the new `BlobValue` declaration `.jobject_input()`, adding `blob_value_echo(BlobValue) -> BlobValue`, and round-tripping the value. The JVM fails with:

```text
java.lang.NoSuchFieldError: io.prebindgen.covertest.model.BlobValue.stamp [B
```

Please update the `ProjectionKind::ValueBlob` branch in `emit/flat_input.rs` to read the `Stamp` object and then its `bytes` (including nullable handling), and add this jobject-input round trip to the covertest. The old `[B` lookup was valid only while the field was inline-erased.

prebindgen/src/api/lang/jnigen/jni/equality.rs
```diff
@@ -0,0 +1,138 @@
...
+enum ArrayProp {
+    /// `ByteArray` / `ByteArray?`.
+    Bytes { nullable: bool },
+}
+
+/// Classify one constructor property. `None` for everything that already
+/// compares by value (scalars, `String`, enums, nested generated classes).
+fn array_prop(ty: &kt::KtType) -> Option<ArrayProp> {
+    match ty.simple_name() {
+        Some("ByteArray") => Some(ArrayProp::Bytes {
```
milyin: This classification fixes only a directly declared `ByteArray`, but supported container shapes can still contain arrays and retain identity equality. For example, adding `chunks: Vec<Vec<u8>>` to the new `BlobValue` generates:

```kotlin
val chunks: List<ByteArray>
// ...
chunks == other.chunks
```

Two independently returned values with equal chunks then fail the PR's own equality assertion, and `toString()` renders the nested element as `[B@...`:

```text
BlobValue(... chunks=[[B@3830f1c0], ...) !=
BlobValue(... chunks=[[B@39ed3c8d], ...)
```

`Vec<Vec<u8>>` resolves and the binding compiles, so this is not an unsupported source shape. Please make the content operations recursive through array-bearing containers (and cover this case), or explicitly reject such properties instead of silently emitting identity semantics.

PR #209, opened 2026-07-27
Remove the value blob; fixed-size arrays cross as Kotlin primitive arrays

milyin:
**Supersedes #207**, whose equality work is carried forward here (commits preserved).

## Why

A `value_class!` type crossed as the **raw memory image** of the Rust struct. That representation produced three defects in one review cycle and carried a latent unsoundness:

- **Unequal** — it surfaced as a `ByteArray`, which compares by identity, so `ZenohId(b) != ZenohId(b)` (the original report).
- **Unfixable as a value class** — Kotlin reserves `equals`/`hashCode` members on a value class through **2.2.21** (measured, including `-Xvalue-classes`), so the wrapper had to become a real class — losing the zero-allocation property that was the entire point.
- **Unsound** — input `read_unaligned::<T>()`-ed caller-supplied bytes after a *length* check only. `Copy` never implied every bit pattern was a valid `T`, and no trait asserts that: `struct S { flag: bool }` plus a byte of `2` was UB.
- **Layout-dependent** — the JVM held a `repr(Rust)` struct's memory image, padding included, for a layout with no stability guarantee.

The blob existed for exactly one reason: the resolver could not handle a fixed-size array field. `ZenohId { bytes: [u8; ZENOH_ID_MAX_SIZE] }` failed with `Unresolved { key: TypeKey("[u8 ; 16]") }` in both directions.

## What

**1. `[T; N]` crosses as the matching Kotlin primitive array.** Table-driven off the existing `JniPrim` scalars — `[u8; N]` → `ByteArray`, `[i64; N]` → `LongArray`, and so on for all eight — bulk-copied through `set_*_array_region` / `get_*_array_region`, boxing nothing. Wider unsigned elements carry raw bits in the signed array, matching the existing scalar rule (`u64` already crosses as a raw `jlong`); Kotlin's `ULongArray` is `@ExperimentalUnsignedTypes` and would push an opt-in onto every consumer of a shared tier.

`N` is never needed at generation time — it is often a const path, not a literal — so the decode leans on `TryFrom<&[T]> for [T; N]`. **That `try_into` is the length check.** Elements convert by cast, never transmute, which is what removes the UB.

**2. The blob is deleted** — declarator, `ProjectionKind::ValueBlob`, `WrapKind::Blob`, `write_value_blobs`, the `__assert_copy` guard, ~50 arms across 17 files. `ValueClassDecl` was a strict subset of `DataClassDecl`, so both call sites migrated by changing the declarator alone.

**3. Content equality** (from #207) is carried forward and **widened** — every Kotlin primitive array compares by identity, not just `ByteArray`.

## Found by doing it, not assumed

- **Consts in type position were never qualified.** `QualifyEmittedTypes` implemented only `visit_type_path_mut`, but an array length is an *expression* path — so `[u8; ZENOH_ID_MAX_SIZE]` emitted a const that wasn't in scope. The new `visit_expr_path_mut` is restricted to **registry-indexed const idents**; that restriction is load-bearing, since expression paths also cover every local in a converter body (`v`, `env`).
- **Four tables** independently encoded "which wires are object-shaped" / "which Kotlin types are arrays". Adding seven wires needed all four, and I found that by breaking three in turn (`missing lifetime specifier` → `unsupported wire JShortArray` → `NoSuchMethodError: run`). Now one table each in `wire_access`.
- **`is_leaf_vec_element`'s doc contradicted its body** (claimed opaque handles excluded; `cfg.is_opaque()` includes them). Pre-existing, corrected in passing.

## Coverage

`Arrays` carries one field per primitive family member plus `[u64; 2]` pinning the raw-bits rule, round-tripped both directions with **per-element value checks** — an equality-only check between two echoes would survive a cast bug. `BlobValue` keeps the array-backed equality cases. `Stamp` becomes the *negative* case: a scalar-only data class the content operators must leave alone.

The two binding-error sections that sourced their failure from the blob's byte-length guard are re-pointed at the fixed-size-array length guard, its direct successor.

## Cost, measured

`ZenohId`'s generated Kotlin is **byte-identical** apart from the now-false blob KDoc — the check that this removal is API-neutral.

One cold-path regression, tracked in **#208**: a bare `ZenohId` return now goes through the `fromParts` builder upcall where the blob returned its wire directly. Nested uses (`EntityGlobalId.zid` inside `SourceInfo` — the sample callback's hot path) ride their parent's single `fromParts` and are unaffected. #208 also covers packing a small array into scalar slots, which would retire the equality machinery for such types entirely.

## Verification

334 lib tests · `PASS - 46 sections` · fmt + clippy clean. Downstream ZettaScaleLabs/zenoh-flat-jni#16 builds against this with 18/18 JVM tests.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/names.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/names.rs
@@ -75,7 +75,7 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
     /// length cannot contain a local, which is what makes this scope safe.
     fn visit_type_array_mut(&mut self, arr: &mut syn::TypeArray) {
         syn::visit_mut::visit_type_mut(self, &mut arr.elem);
-        reject_scoped_array_length(arr);
+        reject_unsupported_array_length(arr);
         let mut lengths = QualifyLengthPaths {
             length_names: self.length_names,
         };
@@ -83,59 +83,69 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
     }
 }
 
-/// Reject an array length that opens a SCOPE.
+/// Refuse an array length built from anything but a small, closed set of
+/// expression forms.
 ///
 /// [`QualifyLengthPaths`] rewrites a bare path to its origin module, which is
-/// sound only while every path in the length names a source item. An inline
-/// const block breaks that — it may bind locals, and a local shadowing a source
-/// item would be rewritten into it:
+/// sound only while every path in the length names a source ITEM. Anything that
+/// can bind a name breaks that premise — a local shadowing a source item gets
+/// rewritten into it:
 ///
 /// ```ignore
 /// [u8; const { let array_len = 3; array_len }]   // `array_len` is a LOCAL
+/// [u8; match 3 { array_len => array_len }]       // ...so is this one
 /// ```
 ///
-/// Qualifying that yields `myflat::array_len`, a function item where a `usize`
-/// was meant. Scope tracking would be the general answer; it is a lot of
-/// machinery for a shape that has no place in an FFI boundary type, so this
-/// rejects the whole family instead — including a block that binds nothing,
-/// since "binds nothing" is exactly the judgement that needs the scope tracking
-/// to be trustworthy. Hoist the value to a named `const` and the length becomes
-/// an ordinary path.
+/// Qualifying either yields `myflat::array_len`, a function item where a
+/// `usize` was meant; a same-typed collision would compile and silently change
+/// the length.
 ///
-/// Rejecting is deliberate over silently mis-qualifying: the failure a user
-/// sees names the type and the fix.
-fn reject_scoped_array_length(arr: &mut syn::TypeArray) {
+/// This is a WHITELIST on purpose. Listing the binding forms instead means
+/// every omitted or newly added `syn::Expr` variant silently reopens the hole —
+/// which is exactly how `match` and `if let` slipped past the first attempt.
+/// Inverting it makes the failure mode "a legitimate length is refused", which
+/// is loud and trivially worked around by hoisting the value into a named
+/// `const`.
+fn reject_unsupported_array_length(arr: &mut syn::TypeArray) {
     // Rendered before the mutable walk below borrows the length.
     let rendered = quote::ToTokens::to_token_stream(&*arr).to_string();
-    struct FindScope(bool);
+    struct Check(Option<&'static str>);
     // `VisitMut` rather than `Visit`: syn's immutable visitor is behind a
     // feature this crate does not enable, and the walk mutates nothing.
-    impl syn::visit_mut::VisitMut for FindScope {
+    impl syn::visit_mut::VisitMut for Check {
         fn visit_expr_mut(&mut self, e: &mut syn::Expr) {
-            // Every expression form that can introduce a binding.
-            if matches!(
+            let ok = matches!(
                 e,
-                syn::Expr::Const(_)
-                    | syn::Expr::Block(_)
-                    | syn::Expr::Unsafe(_)
-                    | syn::Expr::Closure(_)
-                    | syn::Expr::Async(_)
-            ) {
-                self.0 = true;
+                // A literal length, `[u8; 4]`.
+                syn::Expr::Lit(_)
+                    // The names this pass exists to qualify: `MAX`,
+                    // `Holder::N`, and the callee of `array_len()`.
+                    | syn::Expr::Path(_)
+                    // Const arithmetic over those: `A + 1`, `-1`, `(A) * 2`,
+                    // `A as usize`, `array_len()`.
+                    | syn::Expr::Binary(_)
+                    | syn::Expr::Unary(_)
+                    | syn::Expr::Paren(_)
+                    | syn::Expr::Group(_)
+                    | syn::Expr::Cast(_)
+                    | syn::Expr::Call(_)
+            );
+            if !ok && self.0.is_none() {
+                self.0 = Some("an unsupported expression form");
             }
             syn::visit_mut::visit_expr_mut(self, e);
         }
     }
-    let mut find = FindScope(false);
-    syn::visit_mut::VisitMut::visit_expr_mut(&mut find, &mut arr.len);
-    if find.0 {
+    let mut check = Check(None);
+    syn::visit_mut::VisitMut::visit_expr_mut(&mut check, &mut arr.len);
+    if let Some(what) = check.0 {
         panic!(
-            "fixed-size array `{}`: a length that opens a scope (an inline `const {{ … }}` \
-             block, or a closure) is not supported — such a block can bind locals, and this \
-             generator qualifies the length's paths against their source module, which would \
-             rewrite a local into a source item. Hoist the value into a named `const` and use \
-             that as the length.",
-            rendered
+            "fixed-size array `{rendered}`: the length uses {what}. Only a literal, a path, a \
+             call, and const arithmetic over those are supported — anything that can bind a name \
+             (`const {{ … }}`, `match`, `if let`, a closure, a loop) would let a LOCAL be \
+             mistaken for a source item, because this generator qualifies the length's paths \
+             against their source module. Hoist the value into a named `const` and use that as \
+             the length."
         );
     }
 }
--- a/prebindgen/src/api/lang/jnigen/jni/emit/names.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/names.rs
@@ -45,6 +45,10 @@ pub(crate) struct QualifyEmittedTypes<'a> {
     /// Bare type name → the module it is reachable under (the item's origin
     /// crate, or the registry's default module).
     pub(crate) source_names: &'a std::collections::HashMap<String, syn::Path>,
+    /// Bare **const** name → the same. Kept separate from `source_names`
+    /// because it qualifies EXPRESSION paths, where a blanket rename would
+    /// also hit every local variable — see [`Self::visit_expr_path_mut`].
+    pub(crate) const_names: &'a std::collections::HashMap<String, syn::Path>,
 }
 
 impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
@@ -58,6 +62,26 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
             }
         }
         syn::visit_mut::visit_type_path_mut(self, tp);
+    }
+
+    /// Qualify a bare `#[prebindgen]` **const** reference.
+    ///
+    /// A const can appear inside a TYPE — an array length, `[u8; MAX_SIZE]` —
+    /// which `syn` models as an expression, so `visit_type_path_mut` never sees
+    /// it and the generated file references a const that is not in scope. Only
+    /// idents the registry indexes as consts are rewritten: expression paths
+    /// are also every local variable and function name in a converter body, and
+    /// qualifying those indiscriminately would rewrite `v` or `env`.
+    fn visit_expr_path_mut(&mut self, ep: &mut syn::ExprPath) {
+        if ep.qself.is_none() && ep.path.leading_colon.is_none() && ep.path.segments.len() == 1 {
+            let ident = ep.path.segments[0].ident.to_string();
+            if let Some(module) = self.const_names.get(&ident) {
+                let mut qualified = module.clone();
+                qualified.segments.push(ep.path.segments[0].clone());
+                ep.path = qualified;
+            }
+        }
+        syn::visit_mut::visit_expr_path_mut(self, ep);
     }
 }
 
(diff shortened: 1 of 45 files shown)


prebindgen/src/api/lang/jnigen/jni/emit/names.rs
```diff
@@ -59,6 +63,26 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
...
+
+    /// Qualify a bare `#[prebindgen]` **const** reference.
+    ///
+    /// A const can appear inside a TYPE — an array length, `[u8; MAX_SIZE]` —
+    /// which `syn` models as an expression, so `visit_type_path_mut` never sees
+    /// it and the generated file references a const that is not in scope. Only
+    /// idents the registry indexes as consts are rewritten: expression paths
+    /// are also every local variable and function name in a converter body, and
+    /// qualifying those indiscriminately would rewrite `v` or `env`.
+    fn visit_expr_path_mut(&mut self, ep: &mut syn::ExprPath) {
```
milyin, line 174: This visitor rewrites expression paths in the entire generated item, not only const expressions in type positions, so a captured const can collide with generator-owned locals. I reproduced it with the legal source:

```rust
#[allow(non_upper_case_globals)]
#[prebindgen]
pub const env: usize = 4;

#[prebindgen]
pub struct Arrays {
    pub bytes: [u8; env],
}
```

The const need not even be declared to JniGen (`prebindgen` reports it as skipped). Because `registry.consts` still places `env` in `const_names`, generation rewrites every JNI local use:

```rust
perftest_flat::env.get_java_vm()
converter(&mut perftest_flat::env, value)
```

The isolated covertest build then fails with roughly 1,750 errors such as `no method named get_java_vm found for type usize`.

Please scope this qualification to `syn::TypeArray::len` rather than installing a global `visit_expr_path_mut` pass. A `visit_type_array_mut` implementation can visit the element normally and run a small const-path visitor only over `arr.len`; local variables cannot occur in that type-level expression. A regression using a lowercase const named `env` would pin the distinction.

prebindgen/src/api/lang/jnigen/jni/emit/names.rs
```diff
@@ -59,6 +62,45 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
...
+/// Qualifies bare const paths, run ONLY over an array's length expression by
+/// [`QualifyEmittedTypes::visit_type_array_mut`]. Separate from the type
+/// visitor so it can never reach a converter body's locals.
+struct QualifyConstPaths<'a> {
+    const_names: &'a std::collections::HashMap<String, syn::Path>,
+}
+
+impl syn::visit_mut::VisitMut for QualifyConstPaths<'_> {
+    fn visit_expr_path_mut(&mut self, ep: &mut syn::ExprPath) {
+        if ep.qself.is_none() && ep.path.leading_colon.is_none() && ep.path.segments.len() == 1 {
```
milyin: This only qualifies bare, single-segment constants, so an associated const used as the length remains in the generated crate scope. I reproduced it with both types declared to JniGen:

```rust
#[prebindgen]
pub struct ArrayLen { pub marker: u8 }

impl ArrayLen { pub const N: usize = 4; }

#[prebindgen]
pub struct Arrays { pub bytes: [u8; ArrayLen::N] }
```

`cargo build -p covertest-kotlin --offline` then emits `[u8; ArrayLen::N]` in three converter signatures/ascriptions and fails each with `E0433: use of undeclared type ArrayLen`; the expressions need to be `perftest_flat::ArrayLen::N`. Since the fixed-array feature accepts ordinary Rust length expressions, please also qualify the leading source-type segment of multi-segment paths inside `arr.len` (the existing `source_names` map has the origin when that type is declared), and add an associated-const regression.

prebindgen/src/api/lang/jnigen/jni/emit/names.rs
```diff
@@ -59,6 +62,65 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
...
+        if ep.qself.is_none() && ep.path.leading_colon.is_none() {
+            // The map to consult is decided by shape: one segment names a const
+            // directly, more than one means the leading segment is the type
+            // that owns the const.
+            let head = &ep.path.segments[0];
+            let ident = head.ident.to_string();
+            let module = if ep.path.segments.len() == 1 {
+                self.const_names.get(&ident)
+            } else {
+                self.source_names.get(&ident)
```
milyin: This lookup still makes qualification depend on the associated-const owner being declared as part of the JniGen surface. I reproduced the remaining case by removing only `.class(data_class!(Holder))` from the new regression while keeping the captured `Holder` item and declared `Blob`: the targeted test then generates bare `[u8; Holder::N]` in the converter return/parameter types and fails (a compiled consumer hits the same `E0433` as before).

`Holder` is not a boundary field type here; it is only the Rust namespace for a compile-time length. Requiring its declaration emits an otherwise-unused Kotlin class merely to make generated Rust compile, and is asymmetric with the free-const path above, which qualifies indexed consts even when they are skipped by JniGen. Please derive the associated-owner lookup from all indexed source structs/enums plus `origin_module` (or an equivalent registry-wide map), and leave `Holder` undeclared in the regression so this opt-in boundary is pinned.

prebindgen/src/api/lang/jnigen/jni/trait_impl.rs
```diff
@@ -372,11 +372,34 @@ impl JniGen {
...
+                .unwrap_or_else(|| self.default_module(registry));
+            length_names.insert(ident.to_string(), module);
+        };
+        for ident in registry.consts.keys() {
+            add_length_name(ident);
+        }
+        for ident in registry.structs.keys() {
+            add_length_name(ident);
+        }
+        for ident in registry.enums.keys() {
```
milyin: The registry-wide map still omits functions, but an array length can call a source-defined `const fn`. I reproduced this on `c103958` by adding an indexed-but-undeclared item and one field to the new regression:

```rust
pub const fn array_len() -> usize { 4 }
// in Blob
pub called: [u8; array_len()],
```

The focused test reports `skipping undeclared #[prebindgen] fn array_len` and generates bare `[u8; array_len()]` in the input/output converter signatures and ascription; a consumer build therefore fails with `E0425` instead of resolving `myflat::array_len()`. This is the same compile-time-namespace case as the skipped free const and `Holder`: the function does not belong in the Kotlin surface merely because its result determines an array length. Please include `registry.functions` in `length_names` and extend the regression with an undeclared const-fn call.

prebindgen/src/api/lang/jnigen/jni/trait_impl.rs
```diff
@@ -372,11 +372,33 @@ impl JniGen {
...
+        // asymmetric with consts, which qualify whether or not JniGen declared
+        // them.
+        // EVERY indexed item, keyed off the same origin map `origin_module`
+        // reads. A length is an arbitrary const expression, so it can name a
+        // const, the type owning an associated const, or a `const fn` — and
+        // enumerating item KINDS here has already missed one of those three
+        // twice. Iterating the origin map cannot drift from `origin_module`
+        // and needs no revisiting when a new item kind is indexed.
+        let length_names: std::collections::HashMap<String, syn::Path> = registry
+            .item_origins
```
milyin: `item_origins` is not actually the set of every indexed item: `Registry::from_items` only inserts into it when `SourceLocation::crate_name` is `Some`. Core explicitly supports origin-less hand-built streams and documents that adapters fall back to `crate` for them; the previous item-map loops retained those names and applied that fallback.

I reproduced the regression on `227d32b` by changing only the new test fixture to `let loc = SourceLocation::default()` and expecting the documented `crate::` prefix. The focused test then emits bare `[u8; env]`, `[u8; Holder::N]`, and `[u8; array_len()]` in converter code (while declared `Blob` correctly becomes `crate::Blob`) and fails all three qualification assertions; this also breaks if the generated file is included below the crate root. Please derive the name set from all indexed named items independently of whether an origin stamp exists, then use `origin_module(...).unwrap_or_else(default_module)` for the value. A registry-level named-item iterator would avoid returning to per-adapter kind enumeration; an origin-less variant of this regression would pin the fallback.

prebindgen/src/api/lang/jnigen/jni/emit/names.rs
```diff
@@ -59,6 +62,61 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
...
+    /// Qualify a `#[prebindgen]` const used as an ARRAY LENGTH (`[u8; MAX]`).
+    ///
+    /// `syn` models the length as an expression, so `visit_type_path_mut` never
+    /// sees it and the generated file would reference a const that is not in
+    /// scope. The rewrite is confined to `arr.len` on purpose: a generated
+    /// converter body is full of expression paths that are LOCALS (`v`, `env`),
+    /// and a source crate may legally declare `pub const env: usize` — so a
+    /// whole-item expression pass would rewrite those locals to
+    /// `mycrate::env` even when restricted to registered const idents. An array
+    /// length cannot contain a local, which is what makes this scope safe.
```
milyin, line 75: This safety premise is not true once the length contains an inline const block: those blocks may bind locals, and the length visitor currently has no scope tracking. This is valid Rust (verified with `rustc`):

```rust
pub const fn array_len() -> usize { 4 }

pub struct Blob {
    pub local: [u8; const {
        let array_len = 3;
        array_len
    }],
}
```

With `array_len` indexed but undeclared, adding this field to the new regression on `58e4d13` rewrites the tail expression to `myflat::array_len`, producing converter types like:

```rust
[u8; const { let array_len = 3; myflat::array_len }]
```

That is a function item rather than the local `usize`, so the generated consumer no longer compiles (and even a same-typed collision could silently change the length). Please make the length visitor scope-aware for bindings introduced inside inline const/block expressions and leave locally bound paths untouched. Skipping inline consts wholesale would also miss legitimate source references inside them, so a regression should cover both a shadowing local and a source item used in the same kind of block.

prebindgen/src/api/lang/jnigen/jni/emit/names.rs
```diff
@@ -59,6 +62,119 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
...
+/// sees names the type and the fix.
+fn reject_scoped_array_length(arr: &mut syn::TypeArray) {
+    // Rendered before the mutable walk below borrows the length.
+    let rendered = quote::ToTokens::to_token_stream(&*arr).to_string();
+    struct FindScope(bool);
+    // `VisitMut` rather than `Visit`: syn's immutable visitor is behind a
+    // feature this crate does not enable, and the walk mutates nothing.
+    impl syn::visit_mut::VisitMut for FindScope {
+        fn visit_expr_mut(&mut self, e: &mut syn::Expr) {
+            // Every expression form that can introduce a binding.
```
milyin: This list is not every expression form that can introduce a binding. `match` arms bind patterns directly, and `if let`/`while let` do likewise; their nested bodies are not guaranteed to appear as an `Expr::Block` node seen by this matcher.

I verified this with valid Rust and then replaced only the new regression field on `0820084`:

```rust
pub local: [u8; match 3 {
    array_len => array_len,
}],
```

where the indexed `const fn array_len()` already exists. `cargo test -p prebindgen --lib array_length_opening_a_scope_is_rejected --offline` then fails with `test did not panic as expected`. Generation proceeds and the arm-body path is subject to the same rewrite from local `array_len` to the source function.

Please reject pattern-binding forms too (`Match`, `If`/`Let`, and any supported loop forms). Given the intentionally conservative policy here, a whitelist of expression shapes proven not to introduce bindings may be safer than another blacklist: otherwise each omitted/new `syn::Expr` variant recreates the silent-misqualification risk. Add at least the match-arm case to the rejection regression.

prebindgen/src/api/lang/jnigen/jni/emit/names.rs
```diff
@@ -59,6 +62,129 @@ impl syn::visit_mut::VisitMut for QualifyEmittedTypes<'_> {
...
+/// compile-time namespace does not have to be declared to the binding: forcing
+/// that would emit a dead Kotlin class purely to make the generated Rust
+/// compile.
+struct QualifyLengthPaths<'a> {
+    length_names: &'a std::collections::HashMap<String, syn::Path>,
+}
+
+impl syn::visit_mut::VisitMut for QualifyLengthPaths<'_> {
+    fn visit_expr_path_mut(&mut self, ep: &mut syn::ExprPath) {
+        if ep.qself.is_none() && ep.path.leading_colon.is_none() {
```
milyin, line 175: The whitelist accepts every `Expr::Path`, but this branch explicitly skips paths with `qself`. That leaves a valid associated-const spelling accepted yet unqualified:

```rust
pub struct Holder { pub marker: u8 }
// source impl: `impl Holder { const N: usize = 4; }`
pub qualified_assoc: [u8; <Holder>::N],
```

I verified the syntax with `rustc`, then added only this field to the existing qualification fixture on `cab1390`. The focused test generates `[u8; <Holder>::N]` unchanged in the input/output converter signatures and ascription (rather than `[u8; <myflat::Holder>::N]`), so a consumer build cannot resolve `Holder` in the generated crate.

Please either qualify the type inside `qself` with the same registry-wide origin map, or—consistent with the new conservative policy—reject `Expr::Path` when `qself.is_some()` during whitelist validation. The current combination silently accepts a form the qualifier declines to process. A regression using `<Holder>::N` would pin the chosen behavior; `<Holder as Trait>::N` deserves the same explicit decision.

PR #212, opened 2026-07-28
One frontend for array lengths, and a typed source model that carries extents to C (#210, first step of #211)

milyin:
Closes #210. Stage **F0** of #215 (umbrella for #211), and starts **F2**.

> Updated to describe the final state (`e76cd88`). Eleven commits, each answering a review round — see [History](#history).

## The problem

Array-length handling had **two independent walks** over the same expression, both in `jnigen/jni/emit/names.rs`:

* `reject_unsupported_array_length` — a whitelist deciding what is **accepted**;
* `QualifyLengthPaths` — a rewriter deciding what it can **qualify**.

Nothing tied them together, so they drifted eight times. #8 was still open: `[u8; <Holder>::N]` was accepted and emitted verbatim into a crate where `Holder` is not in scope. Cbindgen, meanwhile, had no array handling at all.

## The grammar

**A length is a number.** A generator runs in `build.rs`, where it cannot evaluate Rust. Two spellings reach one, and nothing else does:

| Form | Status |
|---|---|
| `[u8; 4]`, `[u8; 16usize]` | supported |
| `[u8; N]` — **bare name, same source crate, literal initializer** | supported |
| everything else | **refused** |

Refused includes any longer path (`crate::limits::MAX`, `Holder::N`, `usize::MAX`, `<Holder>::N`), an unmarked const, a computed const, a const from another source crate, const arithmetic, `const fn` calls, and anything that can bind a name.

The restriction is on **lengths**, not consts — a `#[prebindgen] const` may still be computed however you like.

## The mechanism

One fallible lowering in `core::frontend`. `Ok` means the length was fully understood **and** reduced to a number, so acceptance is a *consequence* of lowering rather than a parallel judgement. It runs at **ingest**, as pass 3 of `Registry::from_items`, before any adapter exists. jnigen's `reject_unsupported_array_length`, `QualifyLengthPaths` and `length_names` are deleted.

**Same-source provenance**: uniqueness holds across the *marked* namespace only, so a bare name must be a const marked in the item's own source crate — otherwise one crate's unmarked const silently binds to another's marked one of the same name.

## The typed source model

A C header must be able to emit `uint8_t tag[MARKER_TAG_LEN]`, not `[4]` — a symbolic extent is part of that API's meaning. Carrying that fact by stashing the original `syn::Type` for cbindgen to re-read would violate #211's invariants 5 and 6, so the fact goes **in the IR** instead, beginning `SourceModel`:

* `SourceType` — closed and language-neutral; lowering is **total** over the documented grammar.
* `ArrayExtent { value, source }` on the **use site**. `value` keeps `[u8; A]` and `[u8; 4]` one type and one converter; `source` is what C spells.
* The model is the source of truth; the `syn` field types are `to_syn()` of it, never a second representation.

Scope is bounded: **types** complete, **items** structs-only, **adapters** cbindgen's struct-field path. Per-field *classification* still runs on the projection — `SourceType` already answers `is_scalar`/`is_string`/`is_vec`, but deleting those duplicates is F5/F6.

`TypeKey` stays numeric, so jnigen is untouched: covertest PASS, generated Kotlin and Rust byte-identical.

## What the C side needed besides the fact

Both verified with a cbindgen probe rather than assumed:

* **cbindgen had no array support.** `c_field_wire` now accepts an array of non-`bool` scalars — `[u8; N]` is already `#[repr(C)]`-compatible, so it is its own wire and the field copies with no conversion. `[bool; N]` stays refused: its domain is `0`/`1` and a mirror is reinterpreted wholesale with no per-element hook.
* **Consts never reached the header.** They were aliased to `= perftest_flat::ARRAY_BYTES`, which cbindgen cannot evaluate, so it emitted no `#define`. The probe showed that with the alias cbindgen still emits `uint8_t tag[ALIAS_LEN]` — a header referencing an **undefined** symbol. So the literal-value change is load-bearing, not cosmetic. Only extent consts change; the rest keep the alias.

Result:

```c
#define MARKER_TAG_LEN 4
typedef struct marker_t {
  uint8_t tag[MARKER_TAG_LEN];
  uint32_t weight;
} marker_t;
```

## Tests

- **Acceptance matrices** — one for lengths, one for types: source spelling → model, or precise rejection. A new shape is a row. `<Holder>::N` is the #210 regression.
- `a_length_must_name_a_const_from_its_own_crate` — the foreign const has a *usable* value, so it pins provenance rather than incidental unresolvability.
- `equal_lengths_collapse_to_one_type_after_ingest` — inspects state after `from_items` and pins **both** halves: one type-keyed row, and the model still reporting `A`, `B` and a literal per field.
- `data_struct_const_sized_array_keeps_the_const_name` + `..._bool_array_field_is_refused`.
- **smoke.c** round-trips `Marker` under ASan/LSan/UBSan and uses `MARKER_TAG_LEN` as an array bound — the test *compiling* is most of the proof, since a missing `#define` is a compile error.

## History

Eleven commits, each a review round:

1. `0a91388` — frontend seed: one walk, resolution at ingest, jnigen's duplicate machinery deleted
2. `7f1380c` — resolve a path by its item, not its first segment
3. `4880ffe` — a length is a number, not a path; dissolved the multi-source provenance hole and the relative-module ambiguity
4. `961be86` — a type-keyed table stores the number, not the spelling
5. `32370af` — the spelling is discarded, and the model says so
6. `21ab04a` — **reversed**: a C header can spell an extent by name, via the typed source model
7. `13978e7` — the projection preserves type identity (lifetimes, foreign paths; tuples and associated types refused)
8. `a3cf2af` — turbofish and a trailing generic comma normalize away, at the canonicalization boundary
9. `ca21e16` — the callback gate stops dropping the return; refusals say whether they are reserved (#216) or definitive
10. `2b84e02` — normalization is idempotent again; the HRTB refusal is reserved (#222), not "impossible"; `is_unit` lives in core
11. `e76cd88` — the HRTB fix-it states its condition; elision is only exact when each bound lifetime is used once

## Verification

- `cargo test --all --all-features` — 429 lib tests + integration
- clippy `--deny warnings`, fmt — clean
- `examples/regen-check.sh` — byte-identical
- `./examples/smoke-asan.sh` — PASS on Linux CI
- `covertest-kotlin ./gradlew run` — PASS, 46 sections

The `x86_64` goldens are **derived** from the `aarch64` pair (only that arch regenerates locally); the substitution was first verified to reproduce the committed `x86_64` files exactly at HEAD. CI on x86_64 confirms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,3 +189,5 @@ See example projects in the [examples directory](https://github.com/milyin/prebi
 
 - **prebindgen API Reference**: [docs.rs/prebindgen](https://docs.rs/prebindgen)
 - **prebindgen-proc-macro API Reference**: [docs.rs/prebindgen-proc-macro](https://docs.rs/prebindgen-proc-macro)
+- **The source language**: [docs/source-language.md](docs/source-language.md) — which Rust forms a `#[prebindgen]` crate may use
+- **Data-carrying enums**: [docs/sum-types.md](docs/sum-types.md)
--- a/docs/source-language.md
+++ b/docs/source-language.md
@@ -146,14 +146,25 @@ can answer:
 |---|---|---|
 | const index | const name + origin crate | `A` is `4` in crate X |
 | type table (`Registry::array_len`) | `TypeKey` | this array type's length is `4` |
-| per-use source model *(not built yet — #211's `SourceModel`)* | use site | field `S::a` was written `[u8; A]` |
+| per-use source model *(not built — #211's `SourceModel`)* | use site | field `S::a` was written `[u8; A]` |
 
-So `Registry::array_len` returns a bare `usize`. It **cannot** report which
-spelling produced the type, because by the time a `TypeKey` exists that question
-has more than one true answer — and storing one would make it depend on which
-occurrence was seen last. The occurrence model that does know the spelling is
-crate-private for exactly that reason: there is no type-keyed table it could be
-handed out through honestly.
+`Registry::array_len` returns a bare `usize`, and the third row does not exist
+yet. So **the spelling is discarded at the frontend**: after
+`Registry::from_items`, nothing — no type table, no adapter — can tell that
+`S::a` wrote `A`, `S::b` wrote `B`, and `S::literal` wrote `4`.
+
+That is a **policy decision**, not a consequence of Rust's type equality, so it
+is stated rather than assumed. The reasoning: those three are one Rust type, one
+`TypeKey` and one converter, so letting an adapter render them differently would
+mean two destination representations for a single Rust type — which the
+type-keyed converter table cannot express. The capability would be incoherent
+here even if the provenance were carried.
+
+The cost is real and bounded: a C header cannot echo `uint8_t x[MAX_SIZE]`, only
+`uint8_t x[16]`. If that becomes wanted, provenance must arrive on the **use
+site** — field, parameter, return — in the per-use `SourceModel`, and never on a
+type-keyed table, where three occupants of one key cannot be told apart by the
+key.
 
 ### What is refused, and why
 
--- a/docs/source-language.md
+++ b/docs/source-language.md
@@ -56,7 +56,7 @@ Decided by `Registry::scan_fn_signature`
 | a `self` receiver | **refused** — `ScanError::UnsupportedReceiver` |
 | a non-ident parameter pattern (`(a, b): (u8, u8)`) | **refused** — `ScanError::UnsupportedParamPattern` |
 | generic parameters / `where` clauses | unspecified |
-| `impl Fn(T, …) + Send + Sync + 'static` parameter | supported — the callback form |
+| `impl Fn(T, …) + Send + Sync + 'static` parameter | supported — the callback form, see [Callbacks](#callbacks) |
 | any other `impl Trait` | **refused** — `ScanError::DisallowedImplTrait` |
 
 A source crate stays **plain idiomatic Rust**: values are returned by value,
@@ -81,7 +81,7 @@ and canonicalized at ingest by `types_util::normalize_type`.
 | the unit `()` | supported | |
 | a non-empty tuple | **refused** | No adapter has ever lowered one; only `()` is in the language. |
 | raw pointer | walked, adapter-dependent | |
-| `impl Fn(…)` | supported in a parameter position | |
+| `impl Fn(…) + Send + Sync + 'static` | supported in a parameter position | Must return `()` — see [Callbacks](#callbacks). |
 | any other `impl Trait` | refused | |
 | a path with a qualified self (`<T as Trait>::Assoc`) | **refused** | `#[prebindgen]` never captures `impl` blocks, so what it resolves to is unknowable. Outside modeled positions `normalize_type` still leaves it verbatim. |
 
@@ -124,6 +124,40 @@ complete rule set is documented on that function; in summary:
 Because a declaration in `build.rs` is matched against the *normalized*
 spelling, declaring a source item with its crate path
 (`ptr_class!(myflat::Foo)`) is a hard error with a fix-it, not a silent miss.
+
+## Callbacks
+
+**Decided by the frontend** — `core::registry::extract_fn_trait_sig`, the single
+acceptance gate for every callback position.
+
+The accepted form is exactly `impl Fn(T, …) + Send + Sync + 'static`, and the
+`Fn` **must return `()`** (elided or written). Bound order is free and
+canonicalized; so are a trailing input comma and an explicit `-> ()`.
+
+| Form | Status |
+|---|---|
+| `impl Fn(u8) + Send + Sync + 'static` | supported |
+| `impl Fn(u8,) -> () + Sync + Send + 'static` | supported — same type, canonicalized |
+| `impl Fn(u8) -> u16 + …` | **reserved** — issue [#216](https://github.com/milyin/prebindgen/issues/216) |
+| `impl Fn(u8) -> impl Fn(u8) + …` | **reserved** |
+| `impl for<'a> Fn(&'a u8) + …` | **unsupported** |
+
+### Reserved is not the same as unsupported
+
+A refusal says which of two things it is, because they call for opposite
+responses:
+
+* **reserved** — the language intends this and the machinery does not exist
+  yet. Wait, or work around it; the diagnostic names the tracking issue.
+* **unsupported** — it cannot work here and will not. Redesign.
+
+A callback returning a value is *reserved*: every callback wire is void-shaped
+today — C's `call` has no return, jnigen's `run` returns void — so the result
+would be dropped. It used to be **accepted** and silently dropped, which is the
+bug this refusal replaces.
+
+A higher-ranked binder is *unsupported*: no FFI boundary can be generic over a
+lifetime, so no adapter could ever carry one.
 
 ## Array lengths
 
--- a/docs/source-language.md
+++ b/docs/source-language.md
@@ -8,12 +8,16 @@ It is the contract between the source crate and every generator: what a form
 *means* is decided once, by the frontend; how it *crosses a boundary* is decided
 per adapter.
 
-> **Read this first:** only the [array-length subgrammar](#array-lengths) has
-> actually moved into `core::frontend` so far. Every other row below records
-> where the decision is made *today*, which is often inside an adapter. #211
-> tracks moving them. Rows marked **frontend** are decided once and are
-> identical for every generator; rows marked with a site are not, and may differ
-> between C and JNI until they migrate.
+> **Read this first:** the [type grammar](#types) and the
+> [array-length subgrammar](#array-lengths) have moved into `core::frontend`,
+> but they are enforced there **only in modeled positions** — today, the fields
+> of a `#[prebindgen]` struct. Item kinds, function signatures and every other
+> type position are still decided at their use sites, often inside an adapter,
+> and #211 tracks moving them.
+>
+> So a row below can be refused by the frontend in a struct field and reach an
+> adapter unexamined in a function parameter. Where that distinction matters it
+> is stated on the row.
 
 Anything not listed here is **unspecified**. A form that happens to work is not
 thereby part of the language, and the frontend may start refusing it.
@@ -62,21 +66,43 @@ to JNI is the adapter's job.
 
 ## Types
 
-Type positions are walked by `immediate_subtype_positions`
-(`prebindgen/src/api/core/registry.rs`) and canonicalized at ingest by
-`types_util::normalize_type`.
+**Decided by the frontend** in modeled positions —
+`core::frontend::model::lower_type` produces a closed `SourceType`, and lowering
+is total: a form with no variant is refused. Elsewhere, type positions are
+walked by `immediate_subtype_positions` (`prebindgen/src/api/core/registry.rs`)
+and canonicalized at ingest by `types_util::normalize_type`.
 
 | Form | Status | Notes |
 |---|---|---|
 | path type (`Foo`, `Vec<u8>`, `Option<&T>`) | supported | |
-| reference (`&T`, `&mut T`) | supported | |
+| reference (`&T`, `&'a T`, `&mut T`) | supported | |
 | slice (`&[T]`) | supported | |
 | fixed-size array (`[T; N]`) | supported | Length: see [below](#array-lengths). |
-| tuple | supported | |
+| the unit `()` | supported | |
+| a non-empty tuple | **refused** | No adapter has ever lowered one; only `()` is in the language. |
 | raw pointer | walked, adapter-dependent | |
 | `impl Fn(…)` | supported in a parameter position | |
 | any other `impl Trait` | refused | |
-| a path with a qualified self (`<T as Trait>::Assoc`) | left verbatim | Never normalized; its spelling is its identity. |
+| a path with a qualified self (`<T as Trait>::Assoc`) | **refused** | `#[prebindgen]` never captures `impl` blocks, so what it resolves to is unknowable. Outside modeled positions `normalize_type` still leaves it verbatim. |
+
+### Lifetimes are part of a type's name
+
+A lifetime is preserved exactly — `Foo<'static>` is **not** `Foo`, and
+`&'static T` is not `&T` — but it is never modeled as structure, because it
+means nothing to a destination language. The model keeps it verbatim so the
+projection reconstructs the type it came from.
+
+That matters beyond spelling: declarations are matched against the normalized
+form, so `ptr_class!(ZKeyExpr<'static>)` only ever matches a captured
+`ZKeyExpr<'static>`.
+
+### A builtin must be spelled bare
+
+`String` / `Option` / `Vec` / `Box` / `Result` are recognized only as a bare,
+single-segment path. `normalize_type` has already reduced the genuine std
+spellings at ingest and deliberately leaves unknown crate paths alone, so
+`foreign::Option<u8>` is a **foreign named type** that merely shares a name —
+collapsing it would silently retype the value and select the wrong converter.
 
 ### Path normalization
 
--- a/docs/source-language.md
+++ b/docs/source-language.md
@@ -131,6 +131,30 @@ literally are **one type and one converter** — they always were in Rust, and t
 frontend now agrees. And a changed const shows up in the diff of a committed
 generated artifact, where echoing the name would have shown nothing at all.
 
+### Three identities, kept apart
+
+Normalization is deliberately lossy, and that dictates which question each layer
+can answer:
+
+```text
+[u8; A]  where A = 4
+[u8; B]  where B = 4     ──>   [u8; 4]
+[u8; 4]
+```
+
+| Layer | Keyed by | Answers |
+|---|---|---|
+| const index | const name + origin crate | `A` is `4` in crate X |
+| type table (`Registry::array_len`) | `TypeKey` | this array type's length is `4` |
+| per-use source model *(not built yet — #211's `SourceModel`)* | use site | field `S::a` was written `[u8; A]` |
+
+So `Registry::array_len` returns a bare `usize`. It **cannot** report which
+spelling produced the type, because by the time a `TypeKey` exists that question
+has more than one true answer — and storing one would make it depend on which
+occurrence was seen last. The occurrence model that does know the spelling is
+crate-private for exactly that reason: there is no type-keyed table it could be
+handed out through honestly.
+
 ### What is refused, and why
 
 | Form | Reason |
--- a/docs/source-language.md
+++ b/docs/source-language.md
@@ -104,7 +104,7 @@ spelling, declaring a source item with its crate path
 **Decided by the frontend** — `core::frontend::lower_array_len`. This is the
 only subgrammar with an executable acceptance matrix
 (`prebindgen/src/api/core/frontend/tests.rs`); it is closed, and both adapters
-consume the same `ArrayLen`.
+consume the same `ArrayExtent`.
 
 ### A length is always a number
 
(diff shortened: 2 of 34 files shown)


milyin (commented):
I found one correctness hole in the new path lowering before this lands.

`NameIndex::strip_source_head` removes `crate` / `self` / a source-module head before `lower_array_len` decides whether the path is a `SourceConst`. The lookup then checks only the first remaining segment. That breaks a marked free const reached through an intermediate source module, for example:

```rust
mod limits {
    #[prebindgen]
    pub const MAX: usize = 4;
}

#[prebindgen]
pub struct Blob {
    pub bytes: [u8; crate::limits::MAX],
}
```

The registry indexes `MAX`, but `crate::limits::MAX` is first reduced to `limits::MAX`; because `limits` is not an indexed item, this becomes `ExternalConst { path: limits::MAX }`. The generated crate then emits the unqualified `limits::MAX`, which either fails to resolve or can bind a consumer-side module. The same problem occurs for `myflat::limits::MAX`: despite the `ExternalConst` contract saying it is emitted verbatim, the source head is dropped.

This path is inside the documented literal-or-plain-path grammar, so it should either lower to a source-qualified path while preserving the intermediate modules, or be explicitly rejected. Please add acceptance-matrix rows for `crate::limits::MAX` and `myflat::limits::MAX`; they would pin the distinction from the working associated-owner case `crate::Holder::N`.

Aside from that issue, the single frontend decision, transactional per-item rewrite, diagnostics, and retained end-to-end coverage look coherent. All four CI jobs are green.

milyin (commented):
Rereview of `7f1380c`: the original `crate::limits::MAX` case is fixed and the new matrix covers it well, but the revised rule still has a blocking provenance hole.

### 1. A `crate::...` path can bind to an item from the wrong source

`Registry::resolve_array_lengths` builds one registry-wide `NameIndex` and calls the resolver without the origin of the item currently being lowered. Global uniqueness applies only to **indexed** items; it does not prove that a matching bare name denotes the item referenced by `crate` in another source.

A valid two-source example is:

```rust
// source_a
mod limits {
    pub const MAX: usize = 4; // deliberately not marked
}

#[prebindgen]
pub struct A {
    pub bytes: [u8; crate::limits::MAX],
}

// source_b
#[prebindgen]
pub const MAX: usize = 8;
```

The marked namespace is unique. Nevertheless, while lowering `A`, `crate::limits::MAX` is classified source-relative, the global anchor scan finds the indexed `MAX` from source B, and the type becomes `[u8; source_b::MAX]`. That silently changes the length from 4 to 8. It also violates the new documented rule that a source-relative reference to an unmarked item produces `UnresolvedSourcePath`.

The resolver needs the current item origin. A `crate` / `self` / current-source-module path must only anchor to an item from that source; an explicit `source_b::...` path may select source B. Please add a multi-source regression row or registry test where the same bare name exists unmarked in A and marked in B.

### 2. Relative module paths are still ambiguous

The docs say a source may spell a path through its declaring module, but an ordinary `limits::MAX` with no indexed item named `limits` is classified as `ExternalConst` and emitted unchanged. The new `limits::MAX` matrix row passes only because the fixture also creates an unrelated indexed leaf named `limits`, making the head appear source-relative.

The inverse is also possible: an external crate path `limits::MAX` is misclassified as source-relative whenever a marked const/function named `limits` exists, since value-namespace leaves do not prevent `limits` from naming a module or extern crate in the type namespace.

Without indexing source modules, those two valid Rust meanings are indistinguishable. The grammar therefore needs either:

- an explicit-source-head requirement for intermediate modules, with ambiguous relative module paths rejected/documented; or
- module namespace/provenance in the lowering context.

### 3. `NameIndex::new` is no longer usable through the public facade

Its public signature now requires `ItemRole`, but `prebindgen::core::frontend` re-exports `NameIndex` and omits `ItemRole`; `api` is crate-private. An external caller cannot name the constructor argument type. Re-export `ItemRole` beside `NameIndex`, or make these construction details non-public and expose the intended frontend entry point.

All four updated CI jobs pass, and the first fix itself is coherent. Findings 1 and 2 mean the claim that every accepted path is fully resolved is not established yet.

milyin (commented):
Rereview of `4880ffe`: the three previous findings are addressed. Narrowing the grammar removes the namespace ambiguity, same-source provenance is now explicit and tested, the lowering machinery is consistently crate-private, and all four CI jobs pass.

I found one remaining IR consistency issue.

### `Registry::array_len` cannot preserve the `ArrayLen::Const` name it exposes

The new model intentionally distinguishes:

```rust
ArrayLen::Literal(4)
ArrayLen::Const { name: A, value: 4 }
ArrayLen::Const { name: B, value: 4 }
```

and the docs say the const name is retained for diagnostics or generators that want to echo it. However, ingestion rewrites all three source types to `[u8; 4]`, and `array_lens` is keyed only by the rewritten `TypeKey`. Therefore these valid declarations collide:

```rust
#[prebindgen]
pub const A: usize = 4;
#[prebindgen]
pub const B: usize = 4;

#[prebindgen]
pub struct S {
    pub a: [u8; A],
    pub b: [u8; B],
    pub literal: [u8; 4],
}
```

Every `found` entry becomes the same `[u8; 4]` key, and `lens.extend(...)` silently overwrites the previous `ArrayLen`. A generator querying `registry.array_len(TypeKey("[u8; 4]"))` receives `Literal(4)`, `Const(A)`, or `Const(B)` according to traversal / map iteration, not the spelling of the field it is processing. Across separate items, the winner is also nondeterministic because the item maps are `HashMap`s.

The numeric value is stable, so the clean fix seems to be making the registry-level representation canonical and numeric only: do not promise a source const name through a type-keyed side table. If retaining source spelling is important, it has to be stored per source use rather than per semantic `TypeKey`. Please add a collision test covering two equal-valued consts plus the literal spelling.

### PR description

The PR body still describes the superseded grammar as “literal or plain path,” documents `SourceConst` / `ExternalConst`, and says `crate::MAX` and external paths are supported. The final code now accepts only a literal or a same-source bare marked const with a literal initializer. Please update the body so the review and eventual merge record describe `4880ffe`, not the first commit.

Apart from the side-table provenance mismatch, the numeric lowering is substantially cleaner and closes the prior correctness holes.

milyin (commented):
Rereview of `961be86`: the type-keyed collision is fixed correctly, but tracing the full pipeline changes my conclusion about the “simplest representation” choice. The concern about `[u8; A]` versus `[u8; B]` is valid.

These are two different questions:

1. **Are they different Rust types?** No. If `A == B == 4`, both are `[u8; 4]` and should normally share the same low-level converter.
2. **Are they different source uses that an adapter may render differently?** Potentially yes. `A` and `B` may be domain-significant symbolic constants even though they currently have the same value. A destination adapter may want to retain those names, emit different destination aliases/wrappers, preserve symbolic extents in a C surface, or at least carry different documentation and diagnostics.

The frontend/adapter boundary argues for preserving both facts:

```rust
ResolvedArrayLen {
    value: usize,                         // semantic/type identity
    source: Literal | Const(ConstId),    // per-use provenance
}
```

The canonical type table can still use `[u8; 4]` and `Registry::array_len(TypeKey) -> usize`. The const identity must live on the field/parameter/return occurrence, not on `TypeKey`.

### The latest test does not preserve the distinction through the pipeline

`equal_lengths_collapse_to_one_typed_entry` proves that the internal `found` vector temporarily contains three occurrence-level `ArrayLen`s. But `Registry::resolve_array_lengths` immediately:

- rewrites every source item type to the literal `[u8; 4]`;
- stores only `len.value()` in `array_lens`;
- drops the occurrence-level `ArrayLen` vector.

After `Registry::from_items`, `on_struct`, `on_function`, type resolution, and every language adapter see only the literal spelling. No source model or sidecar retains whether a field used `A`, `B`, or `4`. So the test statement “the occurrences stay distinguishable” is true only inside the lowering call, not at the adapter-facing boundary.

This is not a reason to put `Const { name }` back into the type-keyed map; that would recreate the previous bug. It is a reason to retain the occurrence model separately. Since #211 intends a per-use `SourceModel`, the clean options are:

- add the per-use array-length record now, keyed by a stable field/parameter/return use identity; or
- preserve the original item AST alongside the lowered semantic item until `SourceModel` owns that provenance.

A regression should inspect the state available **after `Registry::from_items`** and prove both halves:

- all three uses resolve to semantic length 4 and may share conversion code;
- an adapter/source-model consumer can still tell that `S::a` used `A`, `S::b` used `B`, and `S::literal` used a literal.

If the intended language contract explicitly declares const spelling semantically irrelevant, then discarding it is consistent—but that is a frontend policy decision that removes an adapter capability, not a consequence of Rust type equality. Given the stated boundary where adapters decide destination representation, I recommend preserving the provenance and letting each adapter ignore or use it.

All four CI jobs pass. The PR body also still describes the superseded plain-path grammar and should be updated once this representation decision is settled.

milyin (commented):
Rereview of `21ab04a`: the array direction is now correct. Keeping
`ArrayExtent { value, source }` on each `SourceField` preserves `A` versus `B`
without contaminating the canonical `[u8; 4]` type key, and the C consumer uses
the source half for exactly the policy reason discussed in the previous review.

The larger rework introduces one blocking issue outside array extents.

### `SourceType::to_syn()` changes accepted Rust type identities

`Registry::resolve_array_lengths` now lowers every named struct field and
unconditionally writes `field.ty = ty.to_syn()`. That makes `to_syn()` a
semantic rewrite in the live pipeline, but the model does not retain enough
information to reconstruct several forms that the documented source language
already distinguishes:

```text
&'static Foo                    -> &Foo
Foo<'static>                    -> Foo
(Foo,)                          -> (Foo)
<T as Trait>::Assoc             -> Trait::Assoc
foreign::Option<u8>             -> Option<u8>
```

I reproduced all five at this head with one projection test.

The causes are separate but related:

- `Ref` has no lifetime.
- `lower_path` skips lifetime arguments and then clears the last segment's
  arguments, despite the comment saying the lifetime survives in `name`.
- `Named` stores no `qself`, and `to_syn` forces `qself: None`.
- tuple reconstruction uses `(#(#ts),*)`, which omits the required trailing
  comma for a one-tuple.
- builtin recognition matches only the final segment name. `normalize_type`
  deliberately preserves unknown foreign paths, so `foreign::Option<T>` must
  remain a foreign named type; it must not become the prelude `Option<T>`.

These are not cosmetic spelling differences. They alter `TypeKey`, converter
selection, and the generated field type. In particular, losing lifetimes
directly contradicts `docs/source-language.md`'s existing contract
(`Foo<'static> != Foo`), which is also relied on by declarations such as
`ZKeyExpr<'static>`.

Please make the model identity-preserving for every accepted non-array form
(including lifetimes and qualified self), and require the normalized builtin
forms to be genuinely bare before classifying them. For tuples, retain the
one-element shape. A regression should compare the normalized original type
with the projection's `TypeKey`; the current `to_syn_round_trips` test only
checks that projecting an already-lossy projection is idempotent, so it cannot
detect this class of loss.

There is also stale contract text from the superseded design:

- `core/frontend.rs` still says only the array-length subgrammar has moved and
  that `[u8; A]` provenance is deliberately not carried.
- the “Read this first” note in `docs/source-language.md` likewise says only
  array lengths are in the frontend.

Both now contradict `frontend::model`, `Registry::source_struct`, and the final
PR body, so they should be updated with the model fix.

All four GitHub checks pass, and the existing focused frontend suite passes; the
gap is specifically that its projection matrix does not cover the identity
forms above.

milyin (commented):
Rereview of `13978e7`: the five cases from the previous review are addressed
coherently. Lifetimes and foreign paths retain identity, tuples and
qualified-self types are now explicit frontend rejections rather than silent
rewrites, the stale contract text is fixed, and the replacement regression
compares the original key instead of proving only that a lossy projection is
idempotent.

I found one remaining instance of the same identity problem.

### Generic-path punctuation is lost before `TypeKey`

`Named.args` is reduced to a `Vec<NamedArg>`, and `to_syn()` constructs a fresh
`AngleBracketedGenericArguments`. That discards two pieces of accepted Rust
syntax which `TypeKey` still treats as identity because its canonical form is
the normalized token string:

```text
Wrapper::<u8>  -> Wrapper<u8>
Wrapper<u8,>   -> Wrapper<u8>
```

Both are valid struct-field types. I added each spelling to
`the_projection_preserves_type_identity` locally at this head, and each fails:

```text
TypeKey("Wrapper :: < u8 >") != TypeKey("Wrapper < u8 >")
TypeKey("Wrapper < u8 , >")  != TypeKey("Wrapper < u8 >")
```

This is especially problematic during the partial migration: modeled struct
fields are rewritten through `to_syn()`, while function signatures are not.
The same Rust type can therefore retain the old key in a function signature and
receive the new key in a field, splitting converter lookup by position.

Either result can be coherent, but the loss must happen at the declared
canonicalization boundary:

- If turbofish and a trailing generic comma are semantically irrelevant—as
  `TypeKey`'s purpose suggests—normalize them in `normalize_type` before any key
  is formed, for every position.
- Otherwise preserve them in `SourceType` and its projection.

Please add both spellings to the normalization/projection matrix. This is a
concrete example of why reconstruction should not silently define type
canonicalization.

The PR body is also one commit behind again: it still calls `21ab04a` the final
state and lists six commits, omitting `13978e7`.

Verification: `cargo test --all --all-features` passes locally (425 library
tests), and all four GitHub checks are green.

milyin (commented):
Rereview of `a3cf2af`: the previous finding is fixed in the right place.
`normalize_type` now removes angle-bracket turbofish/trailing-comma punctuation
before any `TypeKey` is formed, recursively and in every position. Both the
normalization test and the source-model identity matrix pin it, lifetimes remain
identity, the PR body is updated, and all four checks are green.

There is one remaining instance of the same “accepted shape exceeds the IR”
problem in callbacks.

### `Callback { args }` accepts and then drops callback semantics

`lower_type` delegates every `impl Trait` to `extract_fn_trait_args`.
That helper recognizes the bound names but returns only the `Fn` inputs; it
does not inspect the parenthesized output or the trait bound's higher-ranked
lifetimes. `SourceType::Callback` likewise stores only `args`, and `to_syn()`
always reconstructs:

```rust
impl Fn(args...) + Send + Sync + 'static
```

Consequently these accepted forms lose identity:

```text
impl Fn(u8) -> u16 + Send + Sync + 'static
    -> impl Fn(u8) + Send + Sync + 'static

impl for<'a> Fn(&'a u8) + Send + Sync + 'static
    -> impl Fn(&'a u8) + Send + Sync + 'static

impl Fn(u8,) + Send + Sync + 'static
    -> impl Fn(u8) + Send + Sync + 'static
```

I added the first two independently to
`the_projection_preserves_type_identity` at this head; both fail with the
transformations above. The third fails as well because the new punctuation
normalization covers `AngleBracketed` path arguments but not `Fn`'s
`Parenthesized` inputs.

This is not only a future struct-model concern. `extract_fn_trait_args` is the
live acceptance gate for function parameters and is used by core resolution and
both adapters. A non-unit callback is currently accepted while the callback
machinery is planned from its arguments alone; the generated callback path is
void-shaped rather than representing the declared result.

Please make the callback grammar explicit and make lowering total over that
grammar:

- If callbacks are intentionally unit-returning with no HRTB, reject an
  explicit non-unit output and higher-ranked binders at ingest with dedicated
  reasons.
- If those forms are supported, add the output/binders to `SourceType::Callback`
  and to converter planning.
- Canonicalize semantically irrelevant callback punctuation/bound ordering at
  the `TypeKey` boundary, or preserve it until that rule exists.

Add accepted/refused rows for explicit `-> ()`, non-unit output, HRTB, a trailing
input comma, and reordered `Send`/`Sync` bounds. The current source-language
table says the callback form is supported but does not state these limits.

Minor documentation residue: the History heading still says “Six commits”
while the list and top note correctly say eight.

milyin (commented):
Rereview of `ca21e16`: the non-unit callback hole is closed correctly at the
shared gate, the live registry path now receives the same reason as the source
model, callback punctuation/bound order canonicalize before `TypeKey`, the
accepted/refused rows cover the requested cases, the PR record is current, and
all four checks are green.

I found three remaining issues.

### 1. `normalize_type` is no longer idempotent for a parenthesized unit return

`visit_path_arguments_mut` tests whether the return is `()` **before** the
default recursive visitor unwraps `Paren`/`Group`. Therefore:

```text
impl Fn(u8) -> (()) + Send + Sync + 'static

first normalization:
impl Fn(u8) -> () + Send + Sync + 'static

second normalization:
impl Fn(u8) + Send + Sync + 'static
```

I added an idempotence test at this head and reproduced exactly that failure.
This matters because registry items are normalized at ingest and
`TypeKey::from_type` normalizes again, while a directly constructed `TypeKey`
normalizes only once. The resulting key can depend on whether its input was
already normalized.

Please canonicalize the parenthesized output after recursively normalizing it
(or normalize the return before testing it), and add the case above to the
normalization matrix.

### 2. An explicit HRTB is not inherently unable to cross FFI

The definitive `HigherRankedBinder` rejection is based on a false distinction.
For the review's concrete case, the accepted elided spelling is itself
higher-ranked. Rust confirms the two are mutually substitutable:

```rust
fn a(f: impl Fn(&u8)) {
    b(f);
}

fn b(f: impl for<'a> Fn(&'a u8)) {
    a(f);
}
```

This compiles. The binder constrains the Rust closure; it is not a generic
parameter carried on the C/JNI ABI. The destination callback handles one
invocation at a time in either spelling.

Complex HRTBs may require model work, so refusing them for now can be coherent,
but they should be **reserved**, not documented as impossible forever. At
minimum, canonicalize/accept the simple elision-equivalent form above, or track
the unsupported machinery without claiming no adapter can represent it.

### 3. Core callback validation depends on JniGen

`core::registry::extract_fn_trait_sig` currently calls:

```rust
crate::api::lang::jnigen::util::is_unit(ret)
```

That reverses the intended dependency: a language-neutral source rule now
depends on one adapter. Core already has the identical
`core::types_util::is_unit`, currently gated by `unstable-cbindgen`. Move or
ungate the predicate in core and have adapters depend on it, not the reverse.

The existing 428-test suite and CI pass; the idempotence and HRTB-equivalence
cases above are not represented.

milyin (commented):
Rereview of `2b84e02`: the three previous implementation findings are fixed.
Normalization is idempotent for the parenthesized-unit case and now has a
property-style regression matrix; the HRTB case is correctly classified as
reserved rather than impossible; core no longer reaches into JniGen for
`is_unit`. The focused core suite passes locally (119 tests), `git diff
--check` is clean, the PR body is current, and all four GitHub checks pass.

One user-facing correctness issue remains.

### The HRTB diagnostic recommends a rewrite that is not generally equivalent

`CallbackReject::HigherRankedBinder::describe()` is used for every explicit
binder, but says:

```text
write the elided form instead (`Fn(&T)`, which means exactly the same thing)
```

That is true for the simple example `for<'a> Fn(&'a T)`, but not for the whole
rejected class. The enum documentation immediately above already gives the
counterexample:

```text
for<'a> Fn(&'a T, &'a T)
Fn(&T, &T)                    // each elided input gets its own binder
```

Those signatures express different lifetime relationships. The current
diagnostic therefore tells an author with the first API to make a semantic API
change and calls it exact. `docs/source-language.md` ends the HRTB explanation
with the same unconditional “Write the elided form.”

Please make the generic message honest, for example: simple independently-bound
reference inputs may be rewritten with elision; other explicit binders remain
reserved pending #222. If an exact fix-it is desired, the gate must first
classify whether that particular binder is elision-equivalent.

Minor residue from the same correction:
`frontend/model/tests.rs` still labels the HRTB row “UNSUPPORTED” and repeats
the now-retracted claim that no FFI boundary can carry it. That comment should
say “RESERVED” so the executable matrix does not document the opposite policy
from the enum and source-language guide.

milyin (commented):
Rereview of `e76cd88`: no further findings.

The remaining HRTB issue is fixed consistently across the diagnostic, the
source-language guide, and the frontend acceptance matrix. The advice is now
conditional: the simple single-use binder has an exact accepted elided
spelling, while a lifetime that couples multiple inputs is explicitly said to
have no accepted spelling pending #222. That removes the incorrect semantic
fix-it without misclassifying HRTBs as an ABI impossibility.

I also checked the complete delta from `2b84e02`: it is limited to those three
contract sites, `git diff --check` is clean, and the PR body/history describe
the current eleven-commit state.

Verification:

- `cargo test -p prebindgen api::core --all-features`: 119 passed
- all four GitHub checks: green

The intentionally deferred work is now stated accurately and does not leave a
correctness hole in the accepted source language.

PR #221, opened 2026-07-28
jnigen: derive a return expansion from a value form (#213)

milyin:
Closes #213 (Gap A). Gaps B and C were "confirm, or state why it cannot" — both
are answered in [this comment](https://github.com/milyin/prebindgen/issues/213#issuecomment-5108600146),
with follow-ups filed as #217, #218, #219, #220.

## The problem

zenoh-flat publishes a **value form** for each handle type — `SampleStruct`,
`HelloStruct`, `EncodingStruct`, `ReplyStruct`, `ReplyErrorStruct` — reached via
`<type>_to_struct`. Each is literally "the type's own accessors gathered into
one struct". A JniGen consumer that wants a decomposed callback has to restate
that field list by hand, and the two copies drift: of zenoh-flat-jni's five
`expand_return!` blocks, **two had already fallen behind** the struct they
mirror (both missing `timestamp_stack`). Nothing caught it.

## The declaration

```rust
expand_return!(Sample).fields(fields!(sample_to_struct))
```

`.fields()` is the plural of the existing `.field()` — it is exactly `.field()`
applied to each struct field, so it keeps the same rule: **each field crosses by
its own type's default output boundary**. A field type with its own
`expand_return!` splices it (a `KeyExpr` field still crosses as its string, not
as a handle), a declared `data_class!` inlines, a field behind `Option`/`Vec`
stays one leaf.

That is what makes adoption safe: the derived list reproduces the boundary a
hand-written one already had. What changes is that it can no longer drift.

Per-field adjustments live on the `FieldsDecl` and key on the **Rust field
ident**, mirroring `FunctionDecl::expand_param(<param name>, decl)`:

```rust
expand_return!(Sample)
    .fields(fields!(sample_to_struct)
        .field("key_expr", expand_return!(KeyExpr).field_self())  // replace one field's decomposition
        .name("origin.node", "nodeId"))                            // rename; dotted for an inlined field
```

Naming a field the struct does not have is a hard error — that is the drift this
declarator exists to catch. `.fields()` mixes with `.field()` / `.field_self()`;
at most one per decl.

## What it generates

For a value form covering every rule (the `Report` fixture added to
`perftest-flat`):

```kotlin
public fun interface ReportCallback {          // what the user writes against
    public fun run(
        summary__count: Long, summary__total: Double,   // spliced by Summary's own decl
        taken: Stamp?,                                  // Option<data class> — one leaf
        origin__secs: Long, origin__nanos: Long,        // non-optional data class — inlined
        outcome: Lookup,                                // a sum, typed
        label: String,
    )
}

public fun interface ReportCallbackRaw {       // the wire — still ONE crossing
    public fun run(
        summary__count: Long, summary__total: Double,
        taken: Stamp?,
        origin__secs: Long, origin__nanos: Long,
        outcome__tag: Int, outcome__found_v0: Long, outcome__failed_v0: String?,
        label: String,
    )
}
```

```rust
let __vf = perftest_flat::report_to_struct(&__cb_arg0);   // called ONCE
// … each leaf reached off __vf
```

## Implementation

**Core.** `UnfoldLeaf.path` becomes `Vec<PathStep>` (`Call` / `Field`, each
carrying its own optionality) so one path can mix accessor calls and field reads
— `sample_to_struct(v)` → `.key_expr` → `keyexpr_as_str(..)`. `DeconRecord::Fields`
+ `FieldRecord` describe a value form; the adapter walks the struct (only it
knows which are declared classes) and core decides per field whether to splice,
riding the existing `visited` / `UnfoldError::Cycle` guard.
`UnfoldPlan.root_call` hoists the value-form call to one local — without it each
field would rebuild the whole struct, cloning all of it once per leaf.
`Prebindgen::deconstructors` now takes `&Registry`, matching
`value_struct_decons`.

**Sums.** A `sealed_class!` field decomposes in place into its selector and one
group per alternative — a sum has no whole-value converter by construction, so
this is the only shape in which it can cross. The user-facing callback still
receives one typed value: the tag and group slots collapse into a single
parameter rebuilt by an inlined `when`, reusing the `GroupDesc` collapsing a
fixed-builder arg already uses. Two generalizations, both behaviour-preserving
for a sum in whole-return position: the selector leaf carries the sum's own type
as its `out_ty` (so the emitter finds the enum from the leaf, not from
`plan.source`, which names the *containing* value once a sum is a field), and
`encode_sum_leaves` becomes `encode_sum_group` over one leaf segment, with
`encode_plan_leaves` segmenting the list instead of handing the whole plan over.

`Vec<sum>` and `Option<sum>` **fields** are refused by name (#220 tracks the
latter).

## Verification

- 425 lib tests (13 new for `.fields()`, 1 for #213 Gap B), `cargo test` and
  `cargo test --all --all-features`
- `clippy --all-targets --no-default-features --all-features --deny warnings`, `cargo fmt --check`
- `examples/regen-check.sh` clean
- `examples/covertest-kotlin$ ./gradlew run` — **47 sections**, including the new
  `output boundary derived from a value form`, whose callback signature is itself
  the assertion (it would not compile if a field had been derived wrongly) and
  which pins the ownership contract for a handle reached through a value form
  and a sum group

One bug found and fixed mid-way, worth calling out because the Kotlin signature
cannot show it: a terminal `Option` field was being unwrapped, handing `ZStamp`
to a converter typed `Option<ZStamp>` (`ZStamp?` on the Kotlin side either way).
`an_optional_field_reaches_its_converter_whole` asserts the Rust shape.

Commit `a92659d` restores the `example-cbindgen` goldens, which an earlier
`git add -A` here picked up from an `--all-features` regeneration — the generator
output is unchanged; only the committed artifact was wrong. That footgun is #219.

## Downstream

Not in this PR (separate repo): replacing zenoh-flat-jni's `Sample` / `Hello` /
`Encoding` / `ReplyError` blocks with `.fields(fields!(<type>_to_struct))` should
bring back the same callback signatures **with `timestamp_stack` added** — the
two drifted decls closing by themselves — after which zenoh-java's
`FlatCallbacks.kt` needs one more lambda parameter, since it consumes the leaves
positionally.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
@@ -428,34 +428,44 @@ pub(crate) fn encode_plan_leaves(
         }
     }
 
-    // A `.fields()` plan reaches every one of its leaves through ONE call of
-    // the value form: bind it to a local first, so the struct is built once
-    // per delivery rather than once per field.
-    let vf = format_ident!("__vf");
-    let mut stmts = match &plan.root_call {
-        Some(f) => {
-            let m = qualify(f);
-            let base = if by_ref {
-                value.clone()
-            } else {
-                quote!(&#value)
-            };
-            quote! { let #vf = #m::#f(#base); }
+    // Every value form on the plan is evaluated ONCE and bound to a local, so
+    // the struct is built once per delivery rather than once per field. Value
+    // forms compose, so this is a list: each hoist is composed from the longest
+    // hoist that is already a proper prefix of it (they arrive outermost-first),
+    // and from `value` otherwise.
+    let mut stmts = TokenStream::new();
+    let mut bound: Vec<(&[PathStep], syn::Ident)> = Vec::new();
+    // The longest already-bound prefix of `path`, if any.
+    let longest_bound =
+        |bound: &[(&[PathStep], syn::Ident)], path: &[PathStep]| -> Option<(usize, syn::Ident)> {
+            bound
+                .iter()
+                .filter(|(p, _)| p.len() < path.len() && path.starts_with(p))
+                .max_by_key(|(p, _)| p.len())
+                .map(|(p, id)| (p.len(), id.clone()))
+        };
+    for (i, prefix) in plan.hoists.iter().enumerate() {
+        let local = format_ident!("__vf{}", i);
+        let (from, mut expr) = match longest_bound(&bound, prefix) {
+            Some((at, outer)) => (at, quote!(&#outer)),
+            None if by_ref => (0, value.clone()),
+            None => (0, quote!(&#value)),
+        };
+        for step in &prefix[from..] {
+            expr = compose_step(&qualify, step, expr);
         }
-        None => TokenStream::new(),
-    };
-    // Leaves that go through the value form reach off that local with the
-    // shared `Call` step already consumed; a sibling `.field()` / `.field_self()`
-    // leaf still reaches from the value itself.
+        stmts.extend(quote! { let #local = #expr; });
+        bound.push((prefix.as_slice(), local));
+    }
+
+    // Reach a leaf off the innermost value form it sits under, with that
+    // prefix's steps already consumed; a leaf under none of them (a sibling
+    // `.field()` / `.field_self()`) still reaches from the value itself.
     let rebase =
         |leaf: &crate::api::core::unfold::UnfoldLeaf| -> (TokenStream, bool, Vec<PathStep>) {
-            match (&plan.root_call, leaf.path.split_first()) {
-                (Some(rc), Some((PathStep::Call { ident, .. }, rest)))
-                    if ident == rc && !rest.is_empty() =>
-                {
-                    (quote!(#vf), false, rest.to_vec())
-                }
-                _ => (value.clone(), by_ref, leaf.path.clone()),
+            match longest_bound(&bound, &leaf.path) {
+                Some((at, local)) => (quote!(#local), false, leaf.path[at..].to_vec()),
+                None => (value.clone(), by_ref, leaf.path.clone()),
             }
         };
 
--- a/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
@@ -514,6 +514,28 @@ pub(crate) fn encode_plan_leaves(
     for (i, h) in plan.hoists.iter().enumerate() {
         let local = format_ident!("__vf{}", i);
         let (from, mut expr) = match longest_bound(&bound, &h.prefix) {
+            // A NESTED consuming form is handed the parent's field by MOVE: a
+            // hoisted value form is an owned struct and its fields are
+            // disjoint, so moving one out leaves every sibling leaf readable.
+            // `compose_step` borrows (`&(e).f`), so the field run to that field
+            // is projected here instead of going through it.
+            Some((at, outer)) if h.consuming => {
+                let last = h.prefix.len() - 1;
+                let lead = &h.prefix[at..last];
+                if lead.iter().all(PathStep::is_plain_field) {
+                    let segs: Vec<&syn::Ident> = lead.iter().map(PathStep::ident).collect();
+                    (last, quote!(#outer #(.#segs)*))
+                } else {
+                    // Reached through an accessor call, so what is in hand is a
+                    // borrow with nothing to give up — clone once and consume
+                    // the clone, exactly as a borrowed root does below.
+                    let mut e = quote!(&#outer);
+                    for step in lead {
+                        e = compose_step(&qualify, step, e);
+                    }
+                    (last, quote!((#e).clone()))
+                }
+            }
             Some((at, outer)) => (at, quote!(&#outer)),
             // A CONSUMING value form is handed the value itself, so its fields
             // move out instead of being cloned out of a borrow. A borrowed plan
--- a/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
@@ -394,10 +394,11 @@ pub(crate) fn reach_leaf_flat(
     // resolved to the owned type precisely so the owning converter boxes the
     // move instead of the borrowed one cloning it.
     //
-    // A trailing `Option` step is excluded here, not by `steps_are_movable`:
-    // return delivery has no `None` arm to put the absent case in, so the whole
-    // `Option` would have to reach the converter — which is what the callback
-    // path's nullable branch does with a `match`.
+    // A trailing `Option` step cannot arrive here at all: return delivery has
+    // no `None` arm for the absent case, so a nullable leaf is routed to
+    // callback delivery when the plan picks its `Delivery` — see
+    // `single_return` in `core/unfold.rs`. `is_plain_field` is what that rules
+    // out, and it stays as the local statement of the same fact.
     if consuming
         && !matches!(leaf.out_ty, syn::Type::Reference(_))
         && path.iter().all(PathStep::is_plain_field)
--- a/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
@@ -370,8 +370,10 @@ fn project_leading_fields(
 pub(crate) fn reach_leaf_flat(
     qualify: &dyn Fn(&syn::Ident) -> syn::Path,
     leaf: &crate::api::core::unfold::UnfoldLeaf,
+    path: &[PathStep],
     base: TokenStream,
     base_is_ref: bool,
+    consuming: bool,
 ) -> TokenStream {
     use crate::api::core::unfold::LeafSource;
     // An optional step BEFORE the last one needs a `match` whose `None` arm has
@@ -379,14 +381,21 @@ pub(crate) fn reach_leaf_flat(
     // not a `JObject` that could be null — so the shape is refused here rather
     // than composed into code that cannot type-check in the consumer's crate.
     assert!(
-        !leaf.path.iter().rev().skip(1).any(PathStep::is_optional),
+        !path.iter().rev().skip(1).any(PathStep::is_optional),
         "jnigen unfold: leaf `{}` reaches through an optional step but is \
          delivered as a single return value, which has no `None` arm — this \
          shape needs callback delivery",
         leaf.name,
     );
-    let (mut e, lead) = project_leading_fields(&base, base_is_ref, &leaf.path);
-    for step in &leaf.path[lead..] {
+    // Under a CONSUMING value form the leaf OWNS its field, so it is moved out
+    // rather than cloned — the same rule `reach`'s `LeafSource::Field` arm
+    // applies on the multi-leaf path, and the reason both live in this file.
+    if leaf.source == LeafSource::Field && consuming && path.iter().all(PathStep::is_plain_field) {
+        let segs: Vec<&syn::Ident> = path.iter().map(PathStep::ident).collect();
+        return quote!(#base #(.#segs)*);
+    }
+    let (mut e, lead) = project_leading_fields(&base, base_is_ref, path);
+    for step in &path[lead..] {
         e = compose_step(qualify, step, e);
     }
     if leaf.source == LeafSource::Field {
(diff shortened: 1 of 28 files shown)


prebindgen/src/api/lang/jnigen/jni/emit/wrapper.rs
```diff
@@ -210,13 +210,13 @@ pub(crate) fn emit_jni_function_wrapper_with_callee(
...
+        // callers read the same code.
+        let qualify = |id: &syn::Ident| -> syn::Path { ext.fn_module(registry, id) };
         let compose = |base: TokenStream, base_is_ref: bool| -> TokenStream {
-            let mut e = if base_is_ref { base } else { quote!(&#base) };
-            for a in &leaf.path {
-                let m = ext.fn_module(registry, a);
-                e = quote!(#m::#a(#e));
-            }
-            e
+            reach_leaf_flat(&qualify, leaf, base, base_is_ref)
```
milyin: [P1] Honor consuming hoists in the single-leaf return path

This shortcut passes only the leaf path and by_ref flag to reach_leaf_flat; it never consults uplan.hoists or Hoist::consuming. For a one-field value form declared with fields_into, I reproduced this generated expression:

    (&(myflat::z_one_into_struct(&__cvsrc)).label).clone()

z_one_into_struct takes ZOne by value, so the generated Rust is ill-typed before considering that the field is cloned rather than moved. The existing single-leaf regression covers only fields with a borrowing accessor, while the consuming tests all produce multi-leaf callback plans and go through encode_plan_leaves.

Please make Delivery::Return emit/rebase the consuming hoist with the same ownership rules as encode_plan_leaves, or route the one-leaf case through a shared plan-level reach that retains hoist metadata.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -422,25 +495,145 @@ pub(crate) fn encode_plan_leaves(
...
+    let mut order: Vec<usize> = (0..n)
+        .filter(|&i| !plan.leaves[i].identity && !in_sum(i))
+        .collect();
+    order.extend((0..n).filter(|&i| plan.leaves[i].identity && !in_sum(i)));
 
     for idx in order {
         let leaf = &plan.leaves[idx];
         let obj_ident = &obj_idents[idx];
+        let (value, by_ref, path) = rebase(leaf);
+        let consuming = hoist_consumes(&leaf.path);
```
milyin: [P2] Move an identity handle field out of a consuming value form

The consuming state is computed here, but the identity branch below returns before using it. Because a field always has a non-empty rebased path, an identity handle therefore takes the clone-via-converter arm even when the innermost hoist consumed its value.

I reproduced this with a two-field ZEnvelopeStruct containing child: ZChild, where ZChild uses field_self. The generated callback contains:

    ZChild_to_jlong_...(…, &__vf0.child)

rather than boxing/moving __vf0.child. This preserves a handle clone (and requires ZChild: Clone) despite fields_into promising that owned fields move out with no clones. Please propagate consuming into the identity branch and treat a plain field path under a consuming hoist like an owned handle root. Sum payload handles appear to need the same ownership audit because their segment currently matches the field by reference.

prebindgen/src/api/lang/jnigen/jni/emit/wrapper.rs
milyin: Fixed in f3f12d0 — confirmed the repro exactly as described.

The hoist loop is now `bind_hoists`, shared by both delivery paths, and `reach_leaf_flat` takes the rebased path plus its hoist's `consuming` flag. The shortcut binds the same `__vfN` locals as `encode_plan_leaves` and reaches the leaf off the innermost one, so a consuming form is handed the value (`z_one_into_struct(__cvsrc)`) and its field is moved, not cloned.

That is the same shape of fix `6133f91` applied to the reach itself, for the same reason — two derivations of one question drift. Now there is one binder and one reach.

Regression: `a_single_leaf_consuming_value_form_moves_its_field`, which builds a one-field `ZOneStruct` off `z_one_into_struct(o: ZOne)` and asserts both halves (`z_one_into_struct(__cvsrc)` present, `z_one_into_struct(&` absent, `.label).clone()` absent). Verified it fails against the previous commit.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
milyin: Fixed in f3f12d0 — right on both counts, including the `Clone` bound it silently required.

The identity branch now computes the owned **place** rather than testing only for the root: the value itself when the path is empty and the plan owns it, or a plain-field run under a consuming hoist (an `Option` step makes the leaf nullable and is excluded by the same test). Either way it is boxed:

```rust
let __vf0 = myflat::z_envelope_into_struct(__cb_arg0);
let __obj0 = jvalue { j: Box::into_raw(Box::new(__vf0.child)) as jlong };
```

Regression: `a_handle_field_of_a_consuming_value_form_moves`, your `ZEnvelopeStruct { child: ZChild, tag: i64 }` with `ZChild` on `.field_self()`. Asserts `Box::new(__vf0.child)` present and `&__vf0.child` absent; verified it fails against the previous commit.

**Sum payloads: confirmed, not fixed here — filed as #228.** `encode_sum_group` matches by reference (`match &__vf0.outcome`) and clones every payload through one chain (`sum_out.rs:319`), so the handle case keeps its clone. Making it move is not a switch to `match __vf0.outcome`: `encode_group_leaf` builds that chain for every payload kind and an owned handle wants this branch's box rather than the borrowed-opaque converter; the selector leaf reads the same matched value, so tag and payload cannot both take it by value without deriving the tag first; and the inert arms and `Copy` payloads have to stay valid. That is a rework of the emitter's ownership model, and getting it wrong emits non-compiling Rust for a shape that works today — so I would rather land it separately than bundle it here.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -310,21 +311,208 @@ pub(crate) fn cast_wire_to_jobject(
...
+        !path.iter().rev().skip(1).any(PathStep::is_optional),
+        "jnigen unfold: leaf `{}` reaches through an optional step but is \
+         delivered as a single return value, which has no `None` arm — this \
+         shape needs callback delivery",
+        leaf.name,
+    );
+    // Under a CONSUMING value form the leaf OWNS its field, so it is moved out
+    // rather than cloned — the same rule `reach`'s `LeafSource::Field` arm
+    // applies on the multi-leaf path, and the reason both live in this file.
+    if leaf.source == LeafSource::Field && consuming && path.iter().all(PathStep::is_plain_field) {
```
milyin: [P2] Move an identity handle in the single-leaf shortcut too

The consuming special case here only covers LeafSource::Field. A sole handle field decomposed through ZChild.field_self() is an identity/Accessor leaf, so the new hoist binding calls the by-value accessor correctly but the shortcut still returns a borrow:

    let __vf0 = myflat::z_single_envelope_into_struct(__cvsrc);
    &__vf0.child

The normal output phase then invokes the borrowed ZChild-to-jlong converter, cloning the handle and requiring ZChild: Clone. I reproduced this with a one-field ZSingleEnvelopeStruct; it is the cross-product not covered by the new String single-leaf test or the multi-leaf handle test.

Please carry consuming ownership through identity Return delivery as well (including an owned output target/boxing path), or reject this shape. As emitted, fields_into still violates its move/no-Clone contract for a valid one-field handle form.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -510,15 +742,31 @@ pub(crate) fn encode_plan_leaves(
...
+            // owned root, or a field of a CONSUMING value form — that form
+            // handed its value over, so its handle fields move out like every
+            // other field rather than being cloned through the converter (which
+            // would also demand a `Clone` the type need not have). Reached by a
+            // plain-field run, so nothing borrows the local as a whole; an
+            // `Option` step would make the leaf nullable and is excluded by the
+            // same test.
+            let owned_place: Option<TokenStream> = if path.is_empty() && !by_ref {
+                Some(value.clone())
+            } else if consuming && path.iter().all(PathStep::is_plain_field) {
```
milyin: [P2] Option<Handle> fields still clone under fields_into

The new owned-place rule deliberately excludes an optional field because its PathStep is not plain. That leaves the nullable identity branch on the borrowed reach even though the consuming value form owns the Option.

For ZOptionalEnvelopeStruct { child: Option<ZChild>, tag: i64 }, I reproduced:

    match &(&__vf0).child {
        Some(__n0) => ZChild_to_jlong_...(…, __n0),
        ...
    }

so a present child is cloned through the borrowed converter and again requires ZChild: Clone. This is not the sum-payload limitation tracked in #228; it is an ordinary optional handle field. Since fields_into owns the Option, the generated code can match it by value and box/move the Some payload. Otherwise this shape should be rejected or the documented no-clone contract scoped accordingly.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
milyin: Fixed in 421531e — and you were right that the shortcut was the wrong place to patch it.

Both this and the `Option<Handle>` comment below are one question asked twice, so it now gets answered once, in the **plan**: `place_is_owned` decides — where an identity leaf's `out_ty` is chosen — whether the value at that path is the plan's to give away (the root of an owned plan, or a field of a form that consumed its value, reached by a movable run of steps). An owned `out_ty` *is* that statement, and it is already what selects the owning converter, so the emitters follow one decision instead of each re-deriving it. `reach_leaf_flat` then moves whenever the leaf owns its `out_ty`, identity and field leaves alike, rather than special-casing `LeafSource::Field`.

That is also the answer to "an owned output target/boxing path": the target IS the owned type now, so the normal output phase invokes the owning `ZChild`-to-jlong converter — the one that boxes `v` rather than `v.clone()`.

Regression: `a_sole_handle_field_of_a_consuming_value_form_moves`, your one-field `ZSingleEnvelopeStruct`. Asserts `z_single_envelope_into_struct(__cvsrc)` and `__vf0.child` without `&`. Verified it fails against the previous commit.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
milyin: Fixed in 421531e. Agreed on all three points — it is an ordinary optional handle field, not #228, and rejecting it was not an option: `SampleStruct.attachment: Option<ZBytes>` is exactly this shape and exactly the hot path `.fields_into(..)` exists for.

The plain-field test was the wrong rule, so it is gone. `steps_are_movable` (plan.rs) is now the single predicate: field reads only, with an `Option` allowed on the **last** one — a `None` arm still hands the whole `Option` over by value, while an `Option` in the middle has to be unwrapped and so can only be borrowed through. The resolver picks the leaf's ownership with it and both emitters project with it, so the two cannot disagree; a disagreement here is precisely a borrow handed to an owning converter.

The nullable identity branch then matches the option BY VALUE:

```rust
let __obj0: JObject = match __vf0.child {
    Some(__n) => { let __h0 = Box::into_raw(Box::new(__n)) as jlong; box_jlong(&mut env, __h0)? }
    None => JObject::null(),
};
```

Regression: `an_optional_handle_field_of_a_consuming_value_form_moves`, your `ZOptionalEnvelopeStruct { child: Option<ZChild>, tag: i64 }`. Asserts `match __vf0.child` present, `&__vf0.child` absent, `Box::new(__n)` present. Verified it fails against the previous commit.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -310,21 +311,219 @@ pub(crate) fn cast_wire_to_jobject(
...
+        leaf.name,
+    );
+    // Under a CONSUMING value form the leaf OWNS what it reaches, so it is
+    // moved out rather than cloned — a field leaf like `reach`'s
+    // `LeafSource::Field` arm on the multi-leaf path, and an IDENTITY leaf
+    // (a handle field decomposed by `.field_self()`) whose `out_ty` the plan
+    // resolved to the owned type precisely so the owning converter boxes the
+    // move instead of the borrowed one cloning it.
+    //
+    // A trailing `Option` step is excluded here, not by `steps_are_movable`:
```
milyin: [P1] Keep a nullable sole leaf off the flat Return path

The comment correctly says this path has no `None` arm, but the planner still selects it: `single_return` only checks `leaves.len() == 1` and ignores `leaf.nullable`. I reproduced this with a `.fields_into(...)` value form whose only field is `Option<ZChild>` and whose child boundary is `.field_self()`. The generated wrapper is:

```rust
let __vf0 = myflat::z_optional_single_into_struct(__cvsrc);
&(&__vf0).child
```

and then calls the owning `ZChild_to_jlong(..., __out)`, which expects `ZChild`, not `&Option<ZChild>`, so the downstream crate does not compile. The new multi-leaf optional-handle test takes callback delivery and therefore misses this cross-product.

Please either exclude nullable leaves when choosing `Delivery::Return` (so the existing callback branch supplies the `None` arm), or lift the leaf nullability into `Option<leaf_ty>` and make this shortcut match/map the trailing option. Making `out_ty` owned addresses move-vs-clone, but does not represent absence.

milyin: Fixed in aef453d — took the first of your two options.

You are right that ownership and absence are different questions, and that my comment described a rule the planner was not enforcing. `single_return` now also requires the leaf to be non-nullable, so the shape goes to callback delivery, whose `None` arm already exists — the leaf crosses as a boxed `Long` / JVM null. That is one condition on the planner instead of teaching the shortcut to match and map a trailing option it has no way to name in `convert_out_ty`.

Worth recording why this changes no working shape: leaf nullability only ever comes from an `Option` with something **decomposed** below it (a `.field_self()` handle, a nested value form). A plain leaf's own `Option` rides its converter and leaves the leaf non-nullable — so every single-leaf plan that returns today still returns. regen-check is byte-identical and covertest's 47 sections are unchanged.

The stale comment in `reach_leaf_flat` now points at the planner rule rather than claiming to enforce it locally.

Regression: `a_sole_optional_handle_field_takes_callback_delivery`, your one-field `ZOptionalSingleStruct`. Asserts no `&__vf0.child` in any form, and that the callback path's by-value `match __vf0.child` with `Box::new(__n)` is what is emitted. Verified it fails against the previous commit.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -310,21 +311,220 @@ pub(crate) fn cast_wire_to_jobject(
...
+    // (a handle field decomposed by `.field_self()`) whose `out_ty` the plan
+    // resolved to the owned type precisely so the owning converter boxes the
+    // move instead of the borrowed one cloning it.
+    //
+    // A trailing `Option` step cannot arrive here at all: return delivery has
+    // no `None` arm for the absent case, so a nullable leaf is routed to
+    // callback delivery when the plan picks its `Delivery` — see
+    // `single_return` in `core/unfold.rs`. `is_plain_field` is what that rules
+    // out, and it stays as the local statement of the same fact.
+    if consuming
```
milyin: [P1] Move an owned root identity even without a consuming hoist

This still ties the move to `consuming`, but a root identity owns its value without any value-form hoist at all. For a declared `ptr_class!(ZChild)`, a type-level `.expand(expand_return!(ZChild).field_self())`, and `fn z_root_child_make() -> ZChild`, the auto-applied plan gives the sole identity leaf owned `out_ty = ZChild` and selects Return delivery. (`fun!(...).expand_return(...field_self())` is already normalized away; this is the type-default path needed when the same identity boundary is spliced as a value-form field.) `hoisted.rebase` is `None`, so this helper receives `consuming = false` and emits:

```rust
let __cvsrc = myflat::z_root_child_make();
{ &__cvsrc }
```

The wrapper then calls the owning `ZChild_to_jlong(..., __out)`, whose argument is `ZChild`, so the generated consumer crate does not compile. I reproduced this in an isolated test at the current head. An outer `Option<ZChild>` has the same mismatch inside its `map` closure.

Please either suppress the auto-applied root plan for an identity-only type default, matching the existing per-function normalization, or make flat-return ownership include an owned empty-path identity rather than relying only on the rebased hoist’s `consuming` flag. The callback emitter already treats that identity as an owned place; this path needs an equivalent outcome.

milyin: Fixed in b5c17f2 — took the second option.

You are right that the condition was wrong, not just incomplete: "a consuming form gave it to me" is only ONE of the two ways a leaf owns what it reaches. The path now reads ownership where the plan already stated it:

```rust
let reached_is_ours = if leaf.identity {
    // `place_is_owned` resolved this to the owned type exactly when the value
    // is the plan's to give away — and that is what selected the owning converter
    !matches!(leaf.out_ty, syn::Type::Reference(_))
} else {
    // a field leaf's out_ty is the field type as written, owned either way,
    // so ownership is the enclosing form's
    consuming
};
```

I preferred this to suppressing the auto-applied root plan because the identity-only type default is not degenerate the way the per-function one is — it is the declaration that makes the boundary spliceable as a value-form field, so it has to survive; and the flat path had to learn this anyway for the `.fields_into(..)` handle field. Suppression would have left the emitter still asking the hoist.

Worth noting this predates the PR: the previous shape of this path composed `&base` for an empty path regardless of anything, so the owned root has always been borrowed here. It just had no way to be wrong until an owning converter was on the other end.

Regression: `an_owned_root_identity_moves_without_any_value_form`, covering both `-> ZChild` and `-> Option<ZChild>` (the `map` closure), asserting neither `&__cvsrc` nor `&__inner` is emitted. Verified it fails against the previous commit.

PR #225, opened 2026-07-28
F5: cbindgen's struct-field classification reads the model, not the projection (#211)

milyin:
Stage **F5** of #215 (umbrella for #211) — the struct-field half.

## The duplicate

#212 made cbindgen's struct-field path read `SourceType` from `Registry::source_struct`. It then threw the model away:

```rust
for (fname, fty) in &fields {
    let fty = &fty.to_syn();          // model discarded
    if is_string(fty) { .. }          // re-decided from syntax
    else if is_bool(fty) { .. }
```

The frontend had already decided that this field is a `String`, a `bool`, or an array of scalars. Projecting back to `syn` and asking again made the projection a second authority on one fact — the drift #211 exists to stop, and the exact shape #210 took.

`c_field_wire`, `is_scalar_array`, `data_field_wire`, `data_field_owns`, `restricted_validity_field`, and the `in_data_struct` / `out_data_struct` field loops now take a `SourceType` and match it:

```rust
fn c_field_wire(ty: &SourceType) -> Option<syn::Type> {
    match ty {
        SourceType::Str => Some(syn::parse_quote!(*mut ::core::ffi::c_char)),
        SourceType::Scalar(ScalarKind::Bool) => Some(bool_wire()),
        SourceType::Scalar(_) => Some(ty.to_syn()),
        SourceType::Array { .. } if is_scalar_array(ty) => Some(ty.to_syn()),
        _ => None,
    }
}
```

## The rule this establishes

**Shape from the model, identity from the registry.**

*"Is this field a `String`, a `bool`, an array of scalars?"* is a source fact the frontend decided. Asking syntax again is the duplicate.

*"Is this the type the binding declared as a `tagged_union`?"* is a registry lookup that happens to be keyed by `TypeKey`, i.e. by a `syn::Type`. Rekeying the registry is **F4's** job, not this stage's, so those call sites project deliberately and say so in a comment rather than pretending to be migrated.

Later stages should follow the same split; without it, "migrate to the model" degenerates into churning every lookup that touches a type.

## What stays on syntax, and why

`mirror_field_wire` (`is_scalar` / `box_inner` / `is_option`) is **not** migrated. It is one policy serving two callers: the `repr_c_struct` mirror path, which has a model, and the tagged-union payload path, whose **variant** fields the frontend does not model yet (F2 covers structs only). Splitting it into a model version and a syntax version would duplicate the policy — precisely the thing this PR deletes. It waits on F2 modeling enums.

Same for parameters and returns: functions are still `syn` items, so that F5 bullet is blocked on F2 too. Both are recorded as such in `docs/source-frontend.md` rather than left as open boxes with no stated blocker.

## The ledger caught it

First real use of #224's boundary check, and it fired on the *decrease*:

```
BOUNDARY LEDGER DRIFT — source-syntax classification sites changed:
  api/lang/cbindgen/mod.rs: 6 -> 5
```

`is_scalar_array` no longer matches `syn::Type::Array`. Total 196 → 195, and the regenerated ledger is in the diff — which is the point of requiring an edit in both directions.

**Note what that number does not capture.** `is_string`, `is_bool` and `is_scalar` were never *in* the ledger: they classify by ident name (`type_path_tail(ty) == "String"`), one of the blind spots the ledger header lists. The count moved by one; the duplication deleted was larger. That is the ledger working exactly as documented — a tripwire against growth, not a measure of progress — and it is worth seeing on the first stage that uses it rather than discovering later.

## Verification

- `cargo test --all --all-features` — green (428 lib tests + the workspace suite).
- `./examples/regen-check.sh` — **byte-identical**. Per the branch's review protocol this stage is *must not move*: this is a pure re-sourcing of the same decisions, so any generated diff would be a bug.
- clippy `--deny warnings`, fmt — clean.

### DIFF
--- a/docs/source-frontend.md
+++ b/docs/source-frontend.md
@@ -37,8 +37,8 @@ i.e. places that classify source shape today:
 |---|---|---|---|
 | `api/core` | 67 | 67 | 74 |
 | `api/lang/jnigen` | 97 | 97 | 96 |
-| `api/lang/cbindgen` | 25 | 26 | 26 |
-| **total** | 189 | 190 | **196** |
+| `api/lang/cbindgen` | 25 | 26 | 26 → 25 |
+| **total** | 189 | 190 | **196 → 195** |
 
 The first two columns were counted by hand and are wrong in both directions — a
 recount gives 69/96/26, a token-level count a third answer. **The `After F8`
@@ -69,7 +69,7 @@ number.
 | F2 | `SourceModel` — the language-neutral item model | **partial** — types complete, items structs-only |
 | F3 | Move the type-shape classifiers into the frontend | not started |
 | F4 | `Registry` consumes `SourceModel` | not started |
-| F5 | Migrate `Cbindgen` | not started |
+| F5 | Migrate `Cbindgen` | **partial** — struct fields classify off the model; params/returns blocked on F2 |
 | F6 | Migrate `JniGen` *(the long pole — 97 sites)* | not started |
 | F7 | Close the open-syntax boundary in the `Prebindgen` trait | not started |
 | F8 | Mechanical boundary check | **done** — landed early, on purpose (see below) |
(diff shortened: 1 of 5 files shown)


milyin (commented):
Review of `7e8802e`: the implementation is coherent. The migrated data-struct,
ownership, restricted-validity, and converter branches now take their shape
from `SourceType`; the remaining projections are identity lookups against
TypeKey-backed declarations. The nested scalar-array rule is behaviorally
equivalent for canonical frontend types and now correctly avoids treating
foreign paths ending in `String`, `bool`, or a scalar name as built-ins.

I found no code correctness issue, but one contract comment still describes the
state before this PR:

### `struct_fields` rustdoc says field classification still uses syntax

`prebindgen/src/api/lang/cbindgen/emit.rs:318` currently says:

```text
The per-field classification ... still runs on SourceType::to_syn();
migrating that off syn is stage F5 ... not this change.
```

That is now the opposite of the implementation and of
`docs/source-frontend.md`: this commit *is* the F5 struct-field migration, and
the listed shape consumers classify `SourceType` directly. Please update this
paragraph to state the new split—shape comes from the model, while TypeKey-backed
identity queries (and the deliberately deferred shared `mirror_field_wire`
policy) may still project to syntax. In a boundary-migration series, leaving
this comment stale risks directing the next change back toward the duplicate
authority this PR removes.

Verification:

- `cargo test -p prebindgen --all-features`: 429 passed; doctests passed
- `./examples/regen-check.sh`: generated output byte-identical
- `git diff --check`: clean
- all four GitHub checks: green

PR #226, opened 2026-07-28
F2: model enums, and move cbindgen's whole enum path onto the model (#211)

milyin:
Stage **F2** of #215 (umbrella for #211), first of three. Enums.

## Why this one first

Everything still open was blocked on the same fact: **the model covered struct fields and nothing else.** F5's `mirror_field_wire` couldn't move, F5's params/returns couldn't move, F6 couldn't start, and F2's own completion criterion had nothing to be an entry point *to*. Enums unblock the first of those and are a precondition for the rest.

## Not a new model — a move and a completion

`core/types_util.rs` already described an enum language-neutrally: `SumSpec` / `SumVariant` / `SumField`, a tag plus one field group per variant, with jnigen already consuming it throughout. Writing a second enum model beside it would have been exactly the duplicate authority #211 exists to remove.

So this **moves** them into the frontend and completes them:

* **payload types are lowered** through `lower_type` instead of carried as raw `syn::Type` — the change that lets cbindgen ask the model;
* **`EnumShape` / `enum_shape` / `first_payload_variant` fold in** as `SourceEnum::is_unit` / `SourceEnum::first_payload_variant` / `SourceVariant::shape`, so "what shape is this variant" stops being a `match` on `syn::Fields` at each use site;
* **the discriminant is modeled** — see below.

`lower_enum` carries the same contract as `lower_type`: acceptance is a consequence of lowering, and a refused payload names its variant and field (`ScanError::UnsupportedVariantFieldType`) at ingest rather than reaching an adapter.

## The discriminant is an `ArrayExtent` problem again

The two backends need different halves of one fact:

* **cbindgen needs the spelling.** `prereq_enums` re-emits `v.discriminant` verbatim into its `#[repr(C)]` mirror — deliberately, so a `const`- or `cfg`-driven value keeps working and `= 0x07` stays `0x07` rather than being re-rendered as `7`.
* **jnigen needs the number.** A Kotlin `NAME(n)` entry and the generated `jint → variant` decode both match on it.

That is the shape F0 already solved for extents, so it gets the same answer:

```rust
pub struct Discriminant {
    pub value: Option<i64>,          // what jnigen needs
    pub source: DiscriminantSource,  // what cbindgen re-emits
}
pub enum DiscriminantSource { Implicit, Explicit(syn::Expr) }
```

Two things fall out. `value: None` turns a **panic** that was reachable from `enum_class!` (`enum_discriminant_values` panicked on a non-literal) into a refusal naming the variant, raised where the consumer that needs the number actually asks. And `Explicit` carries the expression rather than re-rendering from `value`, which is what preserves `0x07`.

`DiscriminantSource::Explicit(syn::Expr)` is the model's **one carrier of open syntax**. That is deliberate, not an oversight: narrowing it to a closed representation would delete C support that exists today. It is the same class of leak F7 tracks for `Niches { value: syn::Expr }`, and it is now listed there rather than passed over.

## cbindgen's enum path follows #225's rule

With variant payloads modeled, the whole path moves under **shape from the model, identity from the registry**: `mirror_field_wire`, `payload_field_wire`, `payload_wire_owns`, `payload_needs_converter`, `enum_variants`, `variant_pattern`, `variant_ctor`, `payload_wire_of`, `assert_unit_enum`, `assert_payload_enum`.

`mirror_field_wire` is the bullet **#225 had to defer**, and worth being explicit about why it moves now rather than then: it is *one* policy serving both the `repr_c_struct` mirror path (modeled) and the tagged-union payload path (then unmodeled). Splitting it into a model version and a syntax version would have duplicated the policy rather than deleted it — the thing being removed. Modeling enums let it move as one piece.

## The ledger

`api/core/types_util.rs` 41 → 39, total 195 → 193 — `extract_int_literal`'s `syn::Expr::Lit` / `syn::Expr::Unary` matches left for the frontend.

As with #225, note what the number does **not** capture: `enum_shape` and `first_payload_variant` matched `syn::Fields`, which is not in `WATCHED` yet. The deletion is larger than the count. Adding `Fields` to `WATCHED` is PR-C's job, once functions are modeled too — that is the point at which the diff of the regenerated ledger *is* the classification F1 asks for.

## Verification

- `cargo test --all --all-features` — green (426 lib tests + workspace suite).
- `./examples/regen-check.sh` — **byte-identical**, C and Kotlin both. This stage re-sources decisions already made, so a generated diff would be a bug. That the discriminant round-trips byte-identically is the specific thing the `Explicit(Expr)` design buys.
- clippy `--deny warnings`, fmt — clean.
- Enum acceptance matrix ported into `frontend/model/tests.rs` and extended: shape, declaration-order tags vs discriminants as two independent numberings, leaf names and members, a refused payload naming its variant and field, Rust's own discriminant assignment rule, an unevaluable discriminant carried rather than refused, and `0x07` keeping its spelling.

Next in the stack: **PR-B** `SourceFn` (unblocks F5's params/returns and F6), then **PR-C** the public `SourceModel` entry point plus F1.


milyin (commented):
I found two blockers in the new enum model:

- **[P1] Preserve zero-field tuple/struct variant shape** (`frontend/model.rs:424-440`). `fields.is_empty()` does not imply `syn::Fields::Unit`: valid variants such as `B()` and `C {}` have empty payloads but still require `E::B()` / `E::C {}` in constructors and patterns. The model now reports both as `VariantShape::Unit`, so cbindgen/JniGen paths emit `E::B` or `E::C`; rustc rejects those as a constructor function / struct variant rather than an enum value. This also changes the old explicit `B()` classification test. Please retain `VariantShape` from `syn::Fields` during lowering (independently of whether the payload group is empty) and cover the generated Rust for both empty forms.

- **[P2] Do not overflow while computing the next discriminant** (`frontend/model.rs:539`). `next = value.map(|n| n + 1)` panics during ingest for valid Rust such as `#[repr(u64)] enum E { A = 9223372036854775807, B }` (and even needlessly after a final `A = i64::MAX`). This reaches every adapter before C can use the preserved spelling, contradicting the stated contract that an unevaluable number does not block C. `checked_add` should turn the numeric chain into `None` while leaving `DiscriminantSource` intact.

`cargo test --all --all-features` passes locally, so these cases are currently outside the suite.

milyin (commented):
One blocker remains after `8d57a70`:

- **[P1] Keep `VariantShape` on the fieldless-enum paths too.** `SourceEnum::is_unit()` deliberately classifies `enum E { A, B(), C {} }` as fieldless because all groups are empty, so Cbindgen accepts it through `.enum_type()` and JniGen through `enum_class!()`. Those paths still spell the Rust variants bare, though: Cbindgen input constructs `#src::#id` (`trait_impl.rs:293-299`), Cbindgen output patterns on `#src::#id` (`trait_impl.rs:1911-1914`), and JniGen input constructs `#source_module::#ident::#variant` (`emit/convert.rs:418-423`). For `B()` / `C {}`, those spellings are constructor items rather than enum values/patterns, so the generated crate does not compile. There is a second JniGen edge here: `#[repr(i32)] enum E { A, B() = 1 }` is valid Rust and now passes `is_unit()`, but `v as jint` is rejected because an explicitly-discriminated non-unit-syntax variant is not castable.

  The new regression tests only put `Paren()` / `Brace {}` beside a real payload variant, which forces the `tagged_union` / `sealed_class` path where `shape.spell(...)` is already used. Please cover an **all-empty** enum declared via `.enum_type()` and `enum_class!()`. Cbindgen can shape-spell the source-side constructor/pattern while keeping the mirror variant bare; JniGen should shape-spell decode and either encode via a discriminant match or explicitly reject the uncastable case.

Local verification otherwise passes: `cargo test --all --all-features` and `cargo clippy --all --all-features -- -D warnings`.

PR #231, opened 2026-07-29
jnigen: hoist a value form that sits under an `Option`

milyin:
## Why

A value form reached through an optional step was refused outright — *"conditional value-form hoisting is not implemented"* (`unfold.rs`). That made `.fields(..)` / `.fields_self_into(..)` all-or-nothing across a binding: `Sample` could not derive its boundary from `SampleStruct` while *any* other type delivered it as `Option<&Sample>`, so the whole containing decomposition had to stay a hand-written field list.

Downstream this blocked eclipse-zenoh/zenoh-flat-jni from adopting zenoh-flat's consuming value forms, because `Reply` hands its sample over as `Option<&Sample>`.

## What

Such a hoist is now **conditional**: it binds `Option<TStruct>`, built inside the `Some` arm of the optional step, and the leaves under it share ONE `match` on that local — the same treatment a decomposed sum's groups already get, and for the same reason. Their slots exist unconditionally; the absent arm fills each with the wire default an inert sum group uses. `leaf_slot` is factored out of `sum_out.rs` so both shapes answer "what does an unfilled slot carry" the same way.

The arm is taken **by value**, so a consuming form's fields still move out inside it. What the optional step hands over is a borrow, so the accessor gets a clone of it — the trade a borrowed root already makes.

Conditional hoists do **not nest**: an inner form would have to be bound inside the outer one's arm, and the binder has no arm to put a second local in. The old rejection stays for exactly that case, renamed for what it is.

Two supporting changes fall out:

- a nullable slot whose reached place is **not itself** optional (the new shape — the leaf is null because its enclosing form was absent, not because the field was) boxes the handle directly instead of matching a non-`Option`;
- an **empty per-field override** is documented as the way to drop a field a value form carries. It is the complete-set rule at its limit, and without it adopting a value form is all-or-nothing over its whole field set. (A generator-level `expand_return!(T)` with no records is still an error.)

## Emitted shape

```rust
let __vf0 = match zenoh_flat::reply_get_sample(&__cb_arg0) {
    Some(__hb0) => Some(zenoh_flat::sample_into_struct((__hb0).clone())),
    None => None,
};
let (__cb0_obj2, /* … */): (JObject, /* … */) = match __vf0 {
    Some(__u0) => { /* leaves off __u0 */ (__cb0_obj2, /* … */) }
    None => (JObject::null(), /* … */),
};
```

## Tests

- `a_value_form_under_an_optional_accessor_is_hoisted_conditionally` — the new shape: form runs once, built in the `Some` arm, leaves share one by-value match, absent arm wire-defaults.
- `an_empty_per_field_override_drops_the_field` — the newly documented contract.
- `an_optional_nested_value_form_is_rejected_before_emission` — kept, now covering only the nesting case.

374 lib tests pass, clippy clean, `examples/regen-check.sh` reports byte-identical generated output.

Verified downstream: eclipse-zenoh/zenoh-flat-jni builds on this and its `gradle build test` passes, as does zenoh-java against the resulting bindings.


prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -687,9 +808,32 @@ pub(crate) fn encode_plan_leaves(
...
+    // groups get, and for the same reason: their slots exist unconditionally
+    // but only one arm computes them.
+    let mut cond_stmts: std::collections::BTreeMap<usize, TokenStream> = plan
+        .hoists
+        .iter()
+        .enumerate()
+        .filter_map(|(i, _)| {
+            plan.leaves
+                .iter()
+                .any(|l| hoisted.conditional(&l.path).is_some_and(|(j, ..)| j == i))
```
milyin: [P1] Keep sum segments inside the conditional hoist arm

A conditional value form can contain a sealed-class/sum field. In that case `rebase` returns `__uN`, but the sum pre-pass above emits `encode_sum_group` into outer `stmts` before the later `match __vfN` introduces that binding. Since `in_sum` also removes every sum leaf from `order`, nothing routes those statements through `cond_stmts`.

I reproduced this at db45a44 with an optional `ZCarrier` value form whose struct contains one `ZOutcome` sealed-class field. Generated Rust contains:

```rust
match &__u0.outcome { ... }
let (...) = match __vf0 {
    Some(__u0) => ...
    None => ...
};
```

and downstream compilation fails with `cannot find value __u0 in this scope`. Please emit each conditional sum group inside the corresponding `Some` body, including its slot bindings, instead of the unconditional pre-pass.

milyin: Fixed in 724a429 — good catch, and the repro was exact.

The sum pass wrote straight to the outer statements while `in_sum` held those leaves out of the per-leaf ordering, so a conditional form's sum field could never reach the arm by any route. The per-hoist buckets are now built *before* the sum pass, and a segment whose selector sits under a conditional hoist lands in that hoist's bucket — slot declarations included, so the `let __objN: ty;` forward declarations and the `match` stay together inside the `Some` body.

Emitted shape now:

```rust
let (__cb0_obj0, __cb0_obj1, __cb0_obj2): (jvalue, JObject, JObject) = match __vf0 {
    Some(__u0) => {
        let __cb0_obj0: jni::sys::jvalue;
        let __cb0_obj1: jni::objects::JObject;
        match &__u0.outcome { /* variant arms */ }
        let __cb0_obj2: jni::objects::JObject = { /* sibling leaf */ };
        (__cb0_obj0, __cb0_obj1, __cb0_obj2)
    }
    None => (jvalue { i: 0i32 }, JObject::null(), JObject::null()),
};
```

Regression test: `a_sum_field_of_a_conditional_value_form_stays_inside_the_arm` — your fixture (optional `ZCarrier` form with a `ZOutcome` sealed-class field). It fails on db45a44 with exactly the placement you described and passes now. It also asserts nothing before the arm reaches `__u0.`, so the ordering itself is pinned rather than just the presence of the match.

375 lib tests pass, clippy clean, regen-check byte-identical, and zenoh-flat-jni still generates identical output.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -469,12 +553,42 @@ pub(crate) fn bind_hoists(
...
+                // What the `Some` arm binds is a borrow of the value (or an
+                // owned one that was about to be dropped), so a CONSUMING
+                // accessor gets a clone — the same trade a borrowed root makes
+                // below, and the only thing there is to give it here.
+                let call = if consuming {
+                    let m = qualify(last.ident());
+                    let f = last.ident();
+                    quote!(#m::#f((#reached).clone()))
+                } else {
+                    compose_step(qualify, last, reached)
```
milyin: [P1] Borrow an owned optional payload before a borrowing value form

`reach_optional` handles both `Option<T>` and `Option<&T>` steps, but after the match this branch treats `__hbN` as though it were always already borrowed. For an accessor `zh_take_carrier(&ZHolder) -> Option<ZCarrier>` followed by the borrowing value form `zc_to_struct(&ZCarrier)`, head generates:

```rust
match myflat::zh_take_carrier(&__cb_arg0) {
    Some(__hb0) => Some(myflat::zc_to_struct(__hb0)),
    None => None,
}
```

This fails downstream compilation with a mismatched argument (`ZCarrier` supplied where `&ZCarrier` is required). I reproduced it at 724a429 by changing the nesting accessor from `Option<&ZCarrier>` to the equally supported owned `Option<ZCarrier>`; the committed tests only exercise the borrowed form. Please preserve whether the optional call produced an owned or borrowed inner value and borrow the owned payload for this non-consuming call. The consuming branch above has the dual issue: it always clones the payload, unnecessarily requiring `T: Clone` instead of moving an owned `T` when available.

milyin: Fixed in 6a2d6ec, and you were right about the second half too.

Which shape the `Some` arm binds now comes from the accessor's own signature, recorded on the hoist as `Hoist::owned_payload` — read in core where the return type is in hand, rather than guessed at emit time. That resolves the call into the four cases it always had:

| payload | `&Self` form | by-value form |
|---|---|---|
| owned `Option<T>` | `f(&__hb0)` | `f(__hb0)` — moved |
| borrowed `Option<&T>` | `f(__hb0)` | `f((__hb0).clone())` |

Only the step immediately before the value form can be owned — any further step composes as a borrow, and a field read is borrowed by construction — so the question is asked of exactly one step.

Two unit tests cover the owned column (`an_owned_optional_payload_is_borrowed_for_a_borrowing_value_form`, `..._is_moved_into_a_consuming_value_form`); both fail on 724a429 with exactly the two symptoms you named.

**But your point about text-only tests landed.** Both of your findings were compile errors, and assertions over emitted strings cannot catch those — so I added the shape to covertest, which actually compiles the generated bindings and runs them on the JVM: `Ledger` (perftest-flat) holds a `Report` two ways, `ledger_filed -> Option<&Report>` and `ledger_archived -> Option<Report>`, both decomposed by `Report`'s existing `.fields_self_into(..)`. Since `Report` carries a sum, one holder exercises **both** review findings under a real compile, and `ledger_new(n)` drives all four present/absent combinations. That immediately caught a third thing neither of us had listed: `Report` was not `Clone`, so the borrowed-payload case did not compile at all. That one is inherent — a consuming form reached through a borrow has nothing to take — and matches what `bind_hoists` already does for a borrowed root, so `Report` now derives `Clone` with a comment saying why. covertest is at 48 sections, all passing.

**One semantic worth your call.** A sum inside a conditional form has no null on the wire for its tag, so an absent value reads as the tag-0 variant while every sibling leaf is null. The new covertest section pins this explicitly (`"-|0 -|0"` for the both-absent row) rather than leaving it implicit. It matches how the product-under-`Option` shape already discriminates, but for an enum whose first variant is meaningful it is a lie the receiver can only detect via the sibling nulls. If you would rather reject a sum inside a conditional form — consistent with `a_sum_field_behind_option_or_vec_is_rejected_by_name` — or make the tag nullable, say which and I will do it; I did not want to pick that unilaterally after you asked for the segment to be emitted.

377 lib tests, clippy clean, regen-check byte-identical, covertest `gradlew run` green, and zenoh-flat-jni regenerates unchanged.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -442,20 +442,104 @@ pub(crate) struct Hoisted {
...
+            for step in &path[lead..k] {
+                e = compose_step(qualify, step, e);
+            }
+            let opt_e = compose_step(qualify, &path[k], e);
+            let bind = format_ident!("__hb{}", depth);
+            let inner = reach_optional(
+                qualify,
+                &path[k + 1..],
+                quote!(#bind),
+                true,
```
milyin: [P1] Preserve ownership after every optional step

The new `Hoist::owned_payload` handles the optional step immediately before the value form, but this recursion still marks every `Some` binding as already borrowed. An owned optional payload may have another ordinary accessor before the value form. For example:

```rust
z_root_child(&ZRoot) -> Option<ZChild>
z_child_carrier(&ZChild) -> &ZCarrier
zc_to_struct(&ZCarrier) -> ZCarrierStruct
```

The conditional hoist is accepted, but head generates:

```rust
Some(__hb0) => Some(
    myflat::zc_to_struct(myflat::z_child_carrier(__hb0))
)
```

so the generated binding fails to compile because `z_child_carrier` needs `&ZChild`. I reproduced this at 6a2d6ec with that three-level deconstructor chain. Please carry the owned/borrowed state of each optional step through `reach_optional` (or explicitly borrow an owned `__hbN` before composing the remaining calls), rather than hardcoding `base_is_ref = true`; recording ownership only for the final step cannot repair an earlier call in the path.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -1003,6 +1191,39 @@ pub(crate) fn encode_plan_leaves(
...
+                    .is_some_and(|(j, ..)| j == i)
+            })
+            .collect();
+        let ids: Vec<&syn::Ident> = idxs.iter().map(|&k| &obj_idents[k]).collect();
+        let tys = idxs
+            .iter()
+            .map(|&k| leaf_slot(registry, &plan.leaves[k]).ty);
+        let defaults = idxs
+            .iter()
+            .map(|&k| leaf_slot(registry, &plan.leaves[k]).default);
```
milyin, line 1278: [P1] Do not encode an absent conditional sum as variant zero

Reusing the primitive slot default makes the absent form indistinguishable from a legitimate tag-0 value. A minimal conditional value form containing only:

```rust
enum ZOutcome { Empty, Failed(String) }
struct ZCarrierStruct { outcome: ZOutcome }
```

emits `(tag = 0, failed = null)` for both `None` and `Some(ZCarrier { outcome: Empty })`. This is not merely awkward for raw builders: the generated typed callback reconstructs the non-null tag-0 enum for an absent parent (the new `LedgerCallback.asRaw()` turns an absent report into `Lookup.Absent`), so the `Option` semantics are silently lost. Sibling nulls cannot recover presence when the form has only the sum, or when every sibling is itself nullable and absent.

Please make a sum selector nullable when its leaf is under a conditional hoist (box the live tag, use JVM null in the absent arm, and reconstruct `Lookup?`), or reject sums in conditional forms. I recommend the nullable selector because the shape is otherwise valid; pinning zero as the absence representation aliases a real source value.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
milyin: Fixed in e3842e7. You were right that recording it at the value form could not reach an earlier call — so the fact moved to the step.

`PathStep::Call` now carries `owned` (whether its `Option`-peeled return is a value or a borrow), recorded in core where the signature is in hand. `reach_optional` consults it at **every** optional step: an owned binding is borrowed before composing what follows, a borrowed one passes through. The terminal case — no steps left, the value form takes the binding directly — deliberately stays untouched, which is what still lets a consuming form MOVE it.

`Hoist::owned_payload` is gone; `bind_hoists` reads the same per-step flag. One source, so the two readings cannot drift.

Your chain is the regression test, `an_owned_optional_payload_is_borrowed_for_the_steps_after_it`:

```rust
zh_child(&ZHolder) -> Option<ZChild>
zchild_carrier(&ZChild) -> &ZCarrier
zc_to_struct(&ZCarrier) -> ZCarrierStruct
```

which now emits `zchild_carrier(&__hb0)` and fails on 6a2d6ec exactly as you described.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
milyin: Fixed in e3842e7 — nullable selector, as you recommended.

Your argument settled it: I had pinned tag-0 as the absence representation in covertest and asked which way to go; "pinning zero aliases a real source value" is the right reading, and the sibling-nulls fallback genuinely does not exist when the form carries only the sum.

A selector under a conditional hoist is now nullable end to end:

- `leaf_is_prim` stops force-priming a `SumTag` when it is nullable, so it takes an object slot and `leaf_slot` gives it `JObject::null()` as its absent default;
- `encode_sum_group` boxes the live tag (`box_jint`) into that slot;
- the iface parameter becomes `Int?` and the reconstructed group `T?`;
- `sum_reconstruct` gains a leading `null -> null` arm, so an absent parent reconstructs as null instead of falling through to the invalid-tag throw.

Generated Kotlin for the covertest case:

```kotlin
ledgerFiled__outcome: Lookup?,                    // typed callback
ledgerFiled__outcome__tag: Int?,                  // raw builder
when (ledgerFiled__outcome__tag) { null -> null; 0 -> Lookup.Absent; ... }
```

An unconditional sum is untouched — still a raw non-null `Int`.

The covertest section now asserts the property at runtime rather than the old behavior: `check((fOutcome == null) == (fLabel == null))` in the typed callback, plus `"-|null"` rows at the raw builder. That is the check that would have caught this, since it only shows up once the value crosses.

378 lib tests, clippy clean, regen-check byte-identical, covertest `gradlew run` green at 48 sections, zenoh-flat-jni regenerates unchanged.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -442,20 +442,115 @@ pub(crate) struct Hoisted {
...
+    base: TokenStream,
+    base_is_ref: bool,
+    depth: usize,
+    body: &dyn Fn(TokenStream) -> TokenStream,
+) -> TokenStream {
+    let (mut e, lead) = project_leading_fields(&base, base_is_ref, path);
+    match (lead..path.len()).find(|&i| path[i].is_optional()) {
+        None => {
+            for step in &path[lead..] {
+                e = compose_step(qualify, step, e);
```
milyin: [P1] Borrow owned results between ordinary accessor steps

`PathStep::owned` is now recorded for every call, but this loop composes ordinary steps without consulting it. The fix handles the optional binding and the final value-form argument; it still loses ownership when a non-optional call in between returns an owned value. For example:

```rust
zh_child(&ZHolder) -> Option<ZChild>
zchild_middle(&ZChild) -> ZMiddle
zmiddle_carrier(&ZMiddle) -> &ZCarrier
zc_to_struct(&ZCarrier) -> ZCarrierStruct
```

Head generates:

```rust
zc_to_struct(zmiddle_carrier(zchild_middle(&__hb0)))
```

and downstream compilation fails because `zmiddle_carrier` receives `ZMiddle` instead of `&ZMiddle`. I reproduced this at e3842e7 with that accepted four-level deconstructor chain. Please make the path fold ownership-aware after every call—when a step yields an owned value and another call follows, borrow that expression before composing the next receiver. Restricting `yields_owned()` to the optional step and the terminal call leaves intermediate owned results broken.

milyin: Fixed in 48e6c3b. Right diagnosis again, and the framing — "restricting it to the optional step and the terminal call leaves intermediate owned results broken" — is what the fix is built on.

The rule now lives in one place, `fold_steps`: compose a step onto what is in hand, borrowing first when the previous step returned a value and the next step is a call. Every fold routes through it — a leaf's reach (`reach_leaf`, `reach_leaf_flat`), a conditional prefix (`reach_optional`), the hoist prefix in `bind_hoists`, and a sum's matched value — so a chain composes identically wherever it appears. `Field` steps need no borrow either way, since `&(e).f` reads through a value and a reference alike.

The one remaining by-value receiver is a consuming value form's own call, which is composed explicitly in the branch that already special-cases it.

That was the actual defect behind all three of these reports: the rule was being applied at whichever sites came to mind rather than at the single place steps are composed. Consolidating it is what makes your four-level chain work without a fourth patch:

```rust
zh_child(&ZHolder) -> Option<ZChild>
zchild_middle(&ZChild) -> ZMiddle      // owned, mid-chain
zmiddle_carrier(&ZMiddle) -> &ZCarrier
zc_to_struct(&ZCarrier) -> ZCarrierStruct
```

now emits `zmiddle_carrier(&myflat::zchild_middle(&__hb0))`, pinned by `an_owned_intermediate_result_is_borrowed_for_the_next_step`, which fails on e3842e7 exactly as you reported.

379 lib tests, clippy clean, regen-check byte-identical, covertest `gradlew run` green at 48 sections, zenoh-flat-jni regenerates unchanged.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -509,12 +674,20 @@ pub(crate) fn bind_hoists(
...
-        }
+        // Every arm above that starts from an OWNED expression does so to hand
+        // it to the trailing by-value value-form call — the move the whole
+        // branch exists for — and positions itself right before it. The rest
+        // start from a reference and fold by the ordinary rule.
+        let tail = &h.prefix[from..];
+        expr = if h.consuming && from + 1 == h.prefix.len() {
+            compose_step(qualify, &tail[0], expr)
+        } else {
+            fold_steps(qualify, tail, expr, false)
```
milyin: [P1] Preserve the by-value boundary of a consuming hoist

This sends the whole tail through `fold_steps(..., false)` unless the consuming call is the only remaining step. That breaks a valid ordinary-accessor → consuming-value-form chain:

```rust
zh_carrier(&ZHolder) -> ZCarrier
zc_into_struct(ZCarrier) -> ZCarrierStruct

expand_return!(ZHolder).field(zh_carrier)
expand_return!(ZCarrier).fields_self_into(fields!(zc_into_struct))
```

I reproduced this at 48e6c3b with an owned `ZHolder` callback argument. Generated Rust is:

```rust
zc_into_struct(&zh_carrier(__cb_arg0))
```

so downstream compilation fails at both boundaries: `zh_carrier` needs `&ZHolder`, while its owned `ZCarrier` result must move into `zc_into_struct`, not be borrowed. The required expression is `zc_into_struct(zh_carrier(&__cb_arg0))`.

Please keep the terminal consuming call out of the ordinary borrowing fold (and preserve the starting/terminal ownership of the folded prefix). The current one-step special case handles a direct consuming form, but not the same form reached through one or more ordinary accessors.

milyin: Fixed in 820ab4f. This was a regression I introduced in 48e6c3b — the one-step special case was me guessing at the shape instead of separating the two boundaries, and you are right that they are independent.

They are now decided separately:

- everything **before** the value form is an ordinary accessor chain and folds by the borrowing rule, from wherever it starts;
- the value form's **own** receiver is decided from what that fold ENDS UP holding — moved in when it is ours, cloned when it is not, borrowed when the form takes `&Self`.

Deciding it from the fold's result rather than from its start is what makes your chain work: the accessor in front borrows, the form itself still moves.

```rust
zc_into_struct(myflat::zh_carrier(&__cb_arg0))
```

pinned by `a_consuming_value_form_keeps_its_by_value_boundary_behind_accessors`, which fails on 48e6c3b with exactly the doubly-broken expression you quoted.

It also collapses a special case rather than adding one: the borrowed-root clone is no longer a distinct starting expression, just the not-ours arm of the terminal rule. One visible consequence in an existing test — the clone is now parenthesized, `zc_into_struct((__inner).clone())`, because it applies to whatever the fold holds and that is not always a bare name (`&__out.clone()` would parse as `&(__out.clone())`). Same semantics, and I updated that assertion with the reason.

380 lib tests, clippy clean, regen-check byte-identical, covertest `gradlew run` green at 48 sections, zenoh-flat-jni regenerates unchanged.

prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
```diff
@@ -469,52 +595,106 @@ pub(crate) fn bind_hoists(
...
-            // A CONSUMING value form is handed the value itself, so its fields
-            // move out instead of being cloned out of a borrow. A borrowed plan
-            // has no value to give up, so it clones first — the same cost the
-            // borrowing form of the accessor would have paid, which keeps one
-            // declaration usable by both owned and `&T` returns of the type.
-            None if h.consuming && by_ref => (0, quote!(#value.clone())),
-            None if h.consuming => (0, value.clone()),
-            None if by_ref => (0, value.clone()),
-            None => (0, quote!(&#value)),
+            Some((outer, rest, _)) => (h.prefix.len() - rest.len(), quote!(&#outer), false),
```
milyin: [P1] Project leading sibling fields before borrowing a rebased hoist

This generic rebased branch starts every mixed path from `&outer`. If an earlier sibling hoist consumed one parent field, and this hoist starts with another field before an accessor, `fold_steps` emits a borrow of the parent as a whole and the generated Rust fails with E0382.

I reproduced this at 820ab4f with an accepted shape: a consuming outer value form returns `{ direct: ZCarrier, wrapper: ZWrapper }`; `ZCarrier` has a consuming value form; and `ZWrapper` reaches another owned carrier through `zw_carrier(&ZWrapper) -> ZCarrier`. The hoists are emitted in field order:

```rust
let __vf1 = zc_into_struct(__vf0.direct);
let __vf2 = zc_into_struct(zw_carrier(&(&__vf0).wrapper));
```

The first line moves `__vf0.direct`; the `&__vf0` in the second line then produces `borrow of partially moved value: __vf0`. This is the same distinction documented by `project_leading_fields`: `&__vf0.wrapper` is a disjoint field borrow and remains valid, while `&(&__vf0).wrapper` first borrows the partially moved parent.

Please project the leading plain-field run of a rebased mixed path directly before folding the remaining accessor steps, instead of seeding this arm with `&outer`. The all-fields consuming special case above handles the pure move path, but the field-then-accessor path needs the same projection invariant.

milyin: Fixed in 75efef8, and you pointed at the fix as well as the bug — `project_leading_fields` already states this invariant, the generic rebased arm just was not using it.

That arm now projects its leading plain-field run directly off the parent local and folds the remaining accessor steps from there:

```rust
let __vf1 = myflat::zc_into_struct(__vf0.direct);
let __vf2 = myflat::zc_into_struct(myflat::zw_carrier(&__vf0.wrapper));
```

`&__vf0.wrapper` is disjoint from the moved `__vf0.direct`, so it survives it; the all-fields consuming arm above already had the projection and this path simply did not.

Your shape is the regression test, `a_rebased_hoist_projects_its_leading_fields_past_a_sibling_move` — consuming outer form returning `{ direct: ZCarrier, wrapper: ZWrapper }`, `ZCarrier` consuming, `ZWrapper` reaching another owned carrier. It fails on 820ab4f with exactly the `&(&__vf0).wrapper` you quoted, and it also asserts the absence of `&(&__vf0)` outright so the whole-parent borrow cannot come back by another route.

381 lib tests, clippy clean, regen-check byte-identical, covertest `gradlew run` green at 48 sections, zenoh-flat-jni regenerates unchanged.

PR #260, opened 2026-08-01
JniGen keeps declarations, not the builder that made them

milyin:
Before #253 a build script held the phases apart itself: it built a `Flat`, made a `RegistryBuilder`, resolved it into a `Registry`, and handed that to the generator. **That type split was the enforcement**, and it lived in the caller's hands.

#253 moved it inside the generator — a better API, and the guarantee went with it. `build(self)` consumed the builder and then stored it back inside the built object:

```rust
pub struct JniGen {
    pub(crate) gen: JniGenBuilder,   // 54 mutators ride along
    pub(crate) registry: Registry<KotlinMeta>,
}
```

The type meaning *"still being described"* was retained by the type meaning *"finished"*.

**Nothing was wrong, and that is the point.** `.gen` / `.registry` were touched only in `mod.rs`, every other file already went through the `&`-accessors, and no `&mut JniGen` existed anywhere. Convention held it — and convention is what this replaces, for the reason #252 gave for `Registry`, whose `supply` survived two commits claiming it was gone.

## The split

| | |
|---|---|
| **`Declarations`** | what was declared. **No `&mut self` method at all.** What a `JniGen` keeps and every emitter reads. |
| **`JniGenBuilder { decls, sources }`** | the describing surface, and the only type with mutators. `build_with` takes `decls` out and never puts them back. |

It fell almost exactly on file boundaries, which is why it is mostly a rename: the **17 mutators** were already concentrated in `builder.rs` and `config.rs`, and the **~117 read methods** were already in files that only read.

`sources` stays on the builder — it is *input to* building, not a declaration, and keeping it out is what makes `Declarations` mean one thing.

`JniGen`'s fields are now module-private rather than `pub(crate)`. **`Flat` comes along for free**: `Registry::flat` is private and `Registry::flat()` returns `&Flat`, so a `Registry` that cannot be reached mutably seals the model too.

## Two guards, both proven to fail before being trusted

**`a_built_jnigen_exposes_no_mutation`** — reads the source with **all** whitespace stripped, so a multi-line signature is indistinguishable from a one-line one, and complains about any `fn(&mut self)` inside an `impl Declarations`, `impl JniGen`, or the `Prebindgen` impl. Sabotaged with a `&mut self` on `Declarations`: it fails and names the file. `builder.rs` and `config.rs` are skipped, exactly as the registry's own test skips `declare.rs` — the builder is *meant* to be mutable.

**`declarations_interior_mutability_is_only_the_two_memos`** — `iface_specs` and `fn_plans` stay `RefCell` deliberately: they cache a pure function of `(declarations, registry)`, both frozen by the time anything reads them, and caching that is not re-declaring anything. Naming them means a third has to be argued for rather than merely added.

Worth noting: its first version scanned line by line and could not see `iface_specs`, whose declaration wraps — the very hazard the sibling test exists for. It strips whitespace too now.

## Verification

- **Reported**: regen-check byte-identical on every committed artifact. The two untracked `example_flat_aarch64_unstable.{rs,h}` are the pre-existing #219 disagreement.
- **Explained**: a pure representation change — nothing about what is generated moves. **Ledger unchanged at 154**, which is the check that no classification site was rewritten while things were being moved.
- **Asserted**: the type a built binding keeps has no mutators.
- 527 lib tests (525 + the two guards), `--all-features` clean, clippy clean on both toolchains with `-D warnings`, and **no new rustdoc warnings** — 31 before, 31 after (the mutator doc-links moved to `JniGenBuilder`).

`Cbindgen` is untouched by decision, and follows by analogy — which is why the guard is written to be copied rather than generalised over both generators now.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/tests/phases.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/tests/phases.rs
@@ -19,15 +19,22 @@
 /// styling: `Registry::supply` survived two commits claiming it was gone because
 /// the check for it was a single-line grep and its signature spanned lines.
 ///
-/// `builder.rs` and `config.rs` are skipped, exactly as the registry test skips
-/// `declare.rs` — `JniGenBuilder` is *meant* to be mutable, and that is the whole
-/// point of it being a different type.
+/// **No file is skipped**, and the exemption is per `impl` block rather than per
+/// file. `builder.rs` and `config.rs` are where the mutators live, but they also
+/// hold six `impl Declarations` blocks between them — most of the read API — so
+/// skipping those files wholesale, as the first version of this test did, left
+/// the larger half of what it claims to guard unguarded.
+///
+/// The self type is read off the `impl` header and matched by **suffix**, so a
+/// qualified path is not a way around it: `impl super::JniGen` occurs in
+/// `kotlin_emit.rs` and `report.rs`, and an unqualified-prefix match missed both.
 #[test]
 fn a_built_jnigen_exposes_no_mutation() {
     let mut offenders: Vec<String> = Vec::new();
 
     let root = concat!(env!("CARGO_MANIFEST_DIR"), "/src/api/lang/jnigen/jni");
     let mut dirs = vec![std::path::PathBuf::from(root)];
+    let mut sealed_blocks = 0usize;
     while let Some(dir) = dirs.pop() {
         for entry in std::fs::read_dir(&dir).expect("jnigen module dir") {
             let path = entry.expect("dir entry").path();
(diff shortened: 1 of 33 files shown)


prebindgen/src/api/lang/jnigen/jni/tests/phases.rs
```diff
@@ -0,0 +1,139 @@
...
+            }
+            if path.extension().is_none_or(|e| e != "rs") {
+                continue;
+            }
+            let name = path
+                .file_name()
+                .expect("file name")
+                .to_string_lossy()
+                .to_string();
+            if name == "builder.rs" || name == "config.rs" {
```
milyin: **[P1] Scan `Declarations` impls in these files too**

Both skipped files also define `impl Declarations` (`builder.rs` contains several of them and `config.rs` contains one), so adding an `fn ...(&mut self)` to any of those impls still leaves this test green. That includes a large share of the read API this guard is intended to protect. The scanner already distinguishes `impl JniGenBuilder` from `impl Declarations`, so please scan these files and exempt only the builder impl blocks rather than skipping the files wholesale; otherwise the stated “no `&mut self` method at all” invariant is not actually enforced.

PR #261, opened 2026-08-01
L2b: one layer read, and api/core stops classifying types

milyin:
**Ledger 154 → 135.** `expand.rs` 4 → 0, `unfold.rs` 16 → 1.

Every one of those sites asked the same question by taking a spelling apart: `option_inner_type`, then `vec_inner_type`, then a `syn::Type::Reference` match, each rebuilding the peeled type as it went. That is `Optional(Sequence(Ref))` — which the model already names.

## `TypeRef::layers()`

Answers it once. **The order is fixed and each layer peels at most once** — `Optional`, then `Sequence`, then `Ref` — because that is the shape of a boundary: *an optional list of borrowed things*. Reading layers in whatever order they appear would call `Vec<Option<T>>` an optional list, which is a different type. A layer out of position stays on the core, where a caller that cares matches `kind` directly: `&[T]` reports `by_ref` with a `Sequence` core, because the borrow is outside the run.

`wrappers` carries the three spellings a fold **registers** — the whole type, the collection, the element — which is a different question from what finally crosses, and the one site that un-requires a leaf-delivered layer needs all three. `TypeRef::fallible_parts()` covers the `Result` sites.

## Not every site wanted the full peel, and a test proved it

`peel_borrow` is the model-driven `peel_ref`, kept separate **deliberately**: a site that peels only the borrow means it, because the layer underneath is what it is about to classify. `Vec<T>` answers `(false, Vec<T>)` there and `(iterable, T)` in `layers()`.

Using `layers()` at the callback-arg site turned *"this arg is a collection"* into *"this arg is a T"* — caught by `callback_arg_nonbare_skipped`, which is exactly the fixture for it. Worth flagging as the one real semantic hazard in this change.

## The blind spot the ledger cannot see

Also migrated the ident-name classifiers — the `seg.ident == "Option"` family. Nine call sites of `option_inner_type` / `vec_inner_type` / `result_err_type` in `unfold`, plus `result_ok_type` in `expand`, which had the model's reading in hand already (`f.ret` is a `TypeRef`).

Those are a **listed blind spot** of the boundary check, so leaving them would have let `api/core` keep classifying while the count claimed otherwise.

## What stays, and why

`peel_ref` stays, and so does its one ledger site: `jnigen` calls it three times, so it cannot be deleted until L4. **The plan said it had no adapter callers; that was wrong**, and the honest count for this stage is 1 rather than 0.

`api/core` is now **13**: `types_util` 10, `registry/scan` 2, `unfold` 1.

- The ten in `types_util` have between 13 and 40 adapter callers each and take no model, so only L3/L4 can free them.
- The two in `scan` inspect a key a build-script author wrote, to diagnose that spelling, and stay for good.

## Verification

- **Reported**: regen-check byte-identical on every committed artifact.
- **Explained**: nothing about what is generated moves — the classification is the same, taken from the model instead of re-derived.
- **Asserted**: no site in `unfold` or `expand` decides what a source type means.
- 527 tests, `--all-features` clean, clippy clean on both toolchains with `-D warnings`, rustdoc still at the 31 baseline.

### DIFF
--- a/prebindgen/src/api/core/expand.rs
+++ b/prebindgen/src/api/core/expand.rs
@@ -31,7 +31,7 @@ use quote::quote;
 
 use crate::api::core::{
     registry::{Registry, TypeKey},
-    types_util::{ident, option_inner_type, result_ok_type},
+    types_util::ident,
 };
 
 mod error;
@@ -197,11 +197,7 @@ pub fn apply<M>(
                 .ok_or_else(|| ExpandError::UnknownFunction(ed.func.clone()))?;
             let param_ty = find_param_type(&item_fn, &ed.param)
                 .ok_or_else(|| ExpandError::UnknownParam(ed.func.clone(), ed.param.clone()))?;
-            let inner = option_inner_type(&param_ty).unwrap_or(param_ty);
-            let bare = match &inner {
-                syn::Type::Reference(r) => (*r.elem).clone(),
-                other => other.clone(),
-            };
+            let bare = crossing_core(registry.flat(), &param_ty);
             if TypeKey::from_type(&bare) != TypeKey::from_type(declared) {
                 return Err(ExpandError::ParamTypeMismatch {
                     func: ed.func.clone(),
@@ -249,11 +245,7 @@ pub fn apply<M>(
             let receiver_key = method_receivers.get(func);
             let mut receiver_skipped = false;
             for (pname, pty) in fn_params(&item_fn) {
-                let core = option_inner_type(&pty).unwrap_or(pty);
-                let bare = match &core {
-                    syn::Type::Reference(r) => (*r.elem).clone(),
-                    other => other.clone(),
-                };
+                let bare = crossing_core(registry.flat(), &pty);
                 let bare_key = TypeKey::from_type(&bare);
                 if !receiver_skipped && receiver_key == Some(&bare_key) {
                     receiver_skipped = true;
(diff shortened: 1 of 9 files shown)


prebindgen/src/api/core/expand.rs
```diff
@@ -312,16 +304,16 @@ fn process_expand<M>(
...
-        None => (false, param_ty.clone()),
-    };
-    let (by_ref, target) = match &inner {
-        syn::Type::Reference(r) => (true, (*r.elem).clone()),
-        other => (false, other.clone()),
-    };
+    // The boundary layers: `Option<&T>` → optional + by_ref, `Option<T>` →
+    // optional, `&T` → by_ref, and `target` is what is left under them.
+    let reading = registry.flat().classify(&param_ty).ok();
+    let layers = reading.as_ref().map(|r| r.layers());
```
milyin: **[P1] Preserve `Sequence` until expansion can build an iterable plan**

`layers().core` also peels a `Sequence`, but this path historically peeled only `Option` and `&`, and `build_plan` is still given no `iterable` bit: it can emit only `Base` or `Optional(Base)`. Consequently, if `T` has a default constructor and a declared function takes `xs: Vec<T>`, the default-application loop now matches `xs` as `T`, this code builds a `FoldPlan { target: T, shape: Base }`, and the generated wrapper reconstructs one `T` and passes it to a Rust parameter expecting `Vec<T>`. The same over-peel occurs in `build_arg` for constructor parameters of type `Vec<T>`. The existing test commentary explicitly says `Iterable(Construct)` is not produced by `apply`, so these expansion sites need an Optional/Ref-only model peel (or full iterable-plan support) rather than `layers().core`.

prebindgen/src/api/core/flat/ty.rs
```diff
@@ -42,6 +42,93 @@ pub struct TypeRef {
...
+    /// make it fit a shape a caller hoped for.
+    ///
+    /// Returning the stack rather than a set of flags is what lets a caller
+    /// **decline**: a consumer that can only build `Base` and `Optional(Base)`
+    /// matches those and falls through on anything else, instead of silently
+    /// consuming a layer it cannot honour.
+    pub fn layer_stack(&self) -> (crate::api::core::shape::Shape, &TypeRef) {
+        match &self.kind {
+            TypeKind::Optional(inner) => {
+                let (rest, core) = inner.layer_stack();
```
milyin: **[P1] Stop the stack when the next layer is out of order**

This recursion contradicts the contract immediately above it. For `Vec<Option<T>>`, the `Sequence` arm recurses into the inner `Optional` and returns `Iterable(Optional(Base))` over `T`; the docs say it must return `Iterable(Base)` over `Option<T>`. This is behaviorally significant because `returns_type` compares `peel(...).core`: a `Vec<Option<T>>` return now matches a deconstructor/value-decomposition target `T`, so `wire_fixed_returns` installs a nested optional fold that the old fixed-order peel deliberately did not install (and the explicit path still rejects as unsupported). Please encode the accepted order—at most outer `Optional`, then `Sequence`, stopping on an out-of-position/repeated layer—and add the documented `Vec<Option<T>>` regression test. `layer_types()` should stop at the same core so its registration view stays aligned with the returned shape.

PR #264, opened 2026-08-01
L4a: the crossing hands over the reading

milyin:
The entry point of jnigen's converter dispatch discarded the model at the door:

```rust
let (dir, key) = crossing;
let ty = key.to_type();                 // rebuild a spelling
self.select_input_type(&ty, built)      // and classify it again
```

…while the registry holds the reading in the cell the scan filled. jnigen's currency is `TypeKey`, so the **whole dispatch chain ran on spellings that were then re-classified** — the same defect #263 removed from `api/core`, one layer out, and the reason a file-by-file L4 would keep producing new instances.

## `Conversions::reading`

That trait is the right seam, and its own doc already said so: *a helper takes `&impl Conversions<M>` and works either side of the boundary*. `Building` answers during the fill, `Registry` afterwards, both from the same cell — so the answer does not change across that line. This is also what L3 will use, which is why doing it here makes `Cbindgen` cheap.

## The selector reads layers, one at a time

`select_input_type` takes a `&TypeRef`. **Deliberately not `layer_stack`**: each arm emits exactly one shape and hands the rest back through `subs` to be selected on its own, so peeling further would claim a shape this selector does not emit. That is the rule L2 paid three defects to learn — the peel is chosen by the consumer's capability.

The callback fallback stops calling `extract_fn_trait_args`: `TypeKind::Callback` carries the argument types as `TypeRef`s, so there is nothing to re-extract.

## What stays

The wildcard patterns the selector builds — `Option<_>`, `&_`, `Vec<_>`. Those are the adapter's own shapes composed from tokens, and composing is spelling rather than reasoning. Six of the file's seven sites are that, which is why **the ledger moves by one**.

## One behaviour question the model settled

The slice arm used to match `syn::Type::Slice`, so `&[T]` took the `Vec<_>` shape and `&Vec<T>` did not. The model classifies both as `Ref(Sequence)` — deliberately, since ownership is the `Ref` layer's fact — so they now take the same arm.

regen-check is byte-identical, so nothing in tree relied on the distinction, and the model's reading is the more correct one. Flagging it because it is the one place this PR could have changed behaviour.

## Verification

- **Reported**: regen-check byte-identical on every committed artifact, **covertest included** — which is the check that counts here, since this touches emission rather than planning.
- **Explained**: same dispatch, chosen from the reading instead of a re-parse.
- **Asserted**: the converter dispatch starts from the model, not from a spelling rebuilt out of a key.
- Ledger 135 → 134, and the small number is the point: it measures who matches syn variants, and this moved who reasons from `origin`.
- 531 tests, `--all-features` clean, clippy clean on both toolchains.

Next in L4: the leaves by what they ask — layer questions (`emit/flat_input`, `emit/wrapper`), names (`emit/names`, `render`), the enum shape.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/selector.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/selector.rs
@@ -33,6 +33,25 @@ fn ref_wildcard(r: &syn::TypeReference) -> syn::Type {
     syn::Type::Reference(pr)
 }
 
+/// Whether a decoded `Vec<T>` local can be borrowed where `referent` is expected.
+///
+/// A **spelling** question, deliberately: it decides what the generated Rust must
+/// be able to say, and Rust distinguishes forms the boundary classification does
+/// not. `[T]` is reached by deref coercion from `&Vec<T>` and `Vec<T>` is the
+/// thing itself; a transparent wrapper such as `Box<Vec<T>>` or `Cow<'_, [T]>`
+/// classifies identically and cannot be reconstructed from the decoded local.
+fn decoded_vec_satisfies(referent: &syn::Type) -> bool {
+    match referent {
+        syn::Type::Slice(_) => true,
+        syn::Type::Path(tp) => tp
+            .path
+            .segments
+            .last()
+            .is_some_and(|s| s.ident == "Vec" && tp.path.segments.len() == 1),
+        _ => false,
+    }
+}
+
 impl Declarations {
     /// Select the input converter for `ty`: terminals, user wrappers, then
     /// built-in structural wrappers.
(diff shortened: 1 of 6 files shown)


prebindgen/src/api/lang/jnigen/jni/selector.rs
```diff
@@ -38,64 +38,81 @@ impl Declarations {
...
             // `Vec<T>`; the call site borrows it (`&Vec<T>` deref-coerces to
             // `&[T]`). Wire/Kotlin type are `List<T>`, identical to a by-value
             // `Vec<T>` input (the writer dedupes the shared converter fn by ident,
             // so the two can coexist). `&mut [T]` is intentionally not supported
             // (no write-back of the decoded Vec).
-            if r.mutability.is_none() {
-                if let syn::Type::Slice(s) = &*r.elem {
-                    let elem = (*s.elem).clone();
+            if matches!(mode, RefMode::Shared) {
+                if let Some(elem) = inner.sequence_elem() {
```
milyin, line 131: **[P1] Do not decode every borrowed `Sequence` as a bare `Vec`**

`TypeKind::Sequence` also covers transparent Rust wrappers around a sequence. For an accepted parameter such as `v: &Box<Vec<i32>>`, `inner` has `Sequence` kind but its spelling is `Box<Vec<i32>>`; this branch now selects `input_vec`, whose converter returns a plain `Vec<i32>`. `emit_plain_decode` then borrows that decoded local and emits a call equivalent to `f(&v)`, so the argument is `&Vec<i32>` where Rust requires `&Box<Vec<i32>>` and the generated binding does not compile. A borrowed `Cow` sequence has the same shape problem. The old syntactic slice shortcut did not claim these forms. Please restrict this shortcut to referent spellings it can actually reconstruct or coerce (`[T]` and direct `Vec<T>`), or add a reconstruction stage for the exact transparent wrapper, with a compile-level regression for `&Box<Vec<T>>`.

PR #267, opened 2026-08-01
Take the reading from the declaration; `reading` becomes a lookup

milyin:
Closes #266.

## What was wrong

`Registry::reading` fell back to `Flat::classify` on a table miss. Instrumenting it showed the fallback firing constantly — `i64` ×48, `String` ×25, `bool` ×16, `Option<ZStamp>` ×16 across the test suite.

Scalars reaching a lookup miss is the tell. `i64` is certainly registered by the time a binding is built, so these were never unknown types — they were an **inverted order**. `flatten` asked `registry.reading(&fr.ty)` for exactly the leaf types its caller registers one loop later:

```rust
let plan = build_plan(..)?;              // <- the walk asks here
for leaf in &plan.leaves {
    registry.require_output(&leaf.out_ty);   // <- registration happens HERE
}
```

Because `classify` returned the *right* answer, the output was correct and nothing showed it. That is the actual defect: an ordering bug papered over by a second source of readings, with the registry no longer the single authority on what a type means.

## What this does instead

Not "register earlier". A `FieldRecord` is adapter-declared, and it is built from an element whose every field **already has a reading** — there was never anything to look up. `FieldRecord::ty` becomes a `TypeRef`, the walk reads it, and the reason the fallback existed goes with it.

That means the value-form walk needs per-field readings, so `TypeKind::DataStruct` carries `&flat::Struct` rather than its `syn::ItemStruct` — the one leak #229's L4 list names by hand:

> `classify.rs` — … Its one leak is `DataStruct { st: &syn::ItemStruct }`, where `flat::Struct` exists

The probes follow: `optional_inner` / `sequence_elem` / `borrow_target` off `kind` instead of re-matching syntax. That retires `unfold::peel_ref` and takes `api/core/unfold.rs` off the boundary ledger entirely (136 → 135). Four sites that only ever wanted the syntax now say `st.origin.syntax`.

`api/core/unfold.rs` now calls `reading` **zero** times. The sole consumer left is `convert_crossing`, whose key comes from `Registry::crossings()` and so always hits a cell.

## Acceptance

- [x] The fallback is deleted; `reading` is a lookup.
- [x] `Flat::classify` keeps exactly one caller, `ensure_entry`.
- [x] Regen byte-identical.

## Verification

- 605 lib tests; `cargo test` and `cargo test --all --all-features` both green (14/14 binaries).
- `examples/regen-check.sh` byte-identical **after a forced rebuild** (`cargo clean -p` the four example crates first — a cached build passes this vacuously).
- `covertest-kotlin ./gradlew run` — 48/48 sections on the JVM.
- The issue's own probe is silent: a clean `cargo build --workspace` prints **0** `PROBE_READING_MISS`, and the test suite prints exactly 1 — the new test deliberately asking for an unregistered `i64`.
- clippy `--all-targets --no-default-features --all-features -- --deny warnings`, and fmt with the CLI-only config.

## Notes for review

- **The new test earns its place.** `classify_has_no_caller_outside_the_registry` cannot catch this: it excludes `core/registry/` by design, because that is where `classify` legitimately lives. The second door was inside the room, so `reading_is_a_lookup_not_a_classification` pins the behaviour directly.
- **Doc comments were the other half.** `Registry::reading`, `Flat::classify` and `flatten` all carried prose *arguing for* the fallback; leaving that would have been worse than leaving the code. `Flat::classify` also named a stale test (`classify_has_exactly_one_caller`), now corrected.
- **Two deliberate widenings**, both toward the model and both regen-neutral: `optional_inner()` sees through `Box<Option<T>>` where the syn matcher did not, and a value form written `-> ()` now gets the clearer "returns nothing" diagnosis instead of "not a struct".
- **The ledger delta is not the measure here** (136 → 135). Most of this change lives in the ledger's own declared blind spot — helper delegation through `classify.rs`, which has zero watched sites. #229 makes exactly this point about #264.
- **`zenoh-flat-jni` could not be regen-checked**: that checkout's `build.rs` predates the current API (`JniGen::new`, `Registry::builder()` arity) and fails to compile identically on unmodified `language-integration`. Pre-existing drift, unrelated to this change.

### DIFF
--- a/prebindgen/src/api/core/unfold.rs
+++ b/prebindgen/src/api/core/unfold.rs
@@ -105,8 +105,17 @@ pub struct FieldRecord {
     pub members: Vec<syn::Ident>,
     /// The leaf name (already `__`-joined across inlined nesting).
     pub name: String,
-    /// The field's type as written, `Option` / `Vec` layers included.
-    pub ty: syn::Type,
+    /// The field's **reading**, `Option` / `Vec` layers included — and its
+    /// syntax with it, which is what a leaf's `out_ty` spells.
+    ///
+    /// The declaration carries this rather than naming a `syn::Type` for the
+    /// walk to look up, because there was nothing to look up: a field record is
+    /// built from an element whose every field already has a reading, and the
+    /// types it names are the ones the caller registers *after* the walk
+    /// returns. Asking the registry here was asking before registration
+    /// (#266) — a lookup that could only miss, answered by a second source of
+    /// readings that hid the ordering.
+    pub ty: crate::api::core::flat::TypeRef,
     /// How this field decomposes.
     pub decon: FieldDecon,
 }
@@ -935,14 +944,6 @@ fn peel_borrow(ty: &crate::api::core::flat::TypeRef) -> (bool, &crate::api::core
     }
 }
 
-/// Strip a single leading `&` (one level) from a type.
-pub(crate) fn peel_ref(ty: &syn::Type) -> syn::Type {
-    match ty {
-        syn::Type::Reference(r) => (*r.elem).clone(),
-        other => other.clone(),
-    }
-}
-
 /// True when `ret` is `T` / `&T` / `Option<T|&T>` / `Vec<T|&T>` with
 /// `T == key` — the default-output match. `Result<_, _>` is NOT peeled, so a
 /// fallible factory (`-> Result<T, E>`) keeps its handle return; the error
@@ -1460,18 +1461,17 @@ fn flatten<M>(
                 });
 
                 for fr in fields {
-                    // A field record is adapter-declared, so it has no element —
-                    // but the scan took its reading when the cell was born, and
-                    // that is what this reads. Nothing is re-classified.
-                    let fr_reading = registry.reading(&fr.ty);
+                    // The declaration carries the field's reading, so nothing is
+                    // looked up and nothing is re-classified.
+                    //
                     // A field's own `Option` makes everything under it nullable,
                     // exactly as an `Option`-returning accessor step does.
-                    let (opt, core) = match fr_reading.as_ref().and_then(|r| r.optional_inner()) {
-                        Some(inner) => (true, inner.origin.syntax.clone()),
-                        None => (false, fr.ty.clone()),
+                    let (opt, core) = match fr.ty.optional_inner() {
+                        Some(inner) => (true, inner),
+                        None => (false, &fr.ty),
                     };
-                    let child_ty = peel_ref(&core);
-                    let child_key = TypeKey::from_type(&child_ty);
+                    let child_ty = core.borrow_target().unwrap_or(core);
+                    let child_key = child_ty.key();
 
                     // Same three-way choice a `.field()` record makes: declared
                     // override, else the field type's own deconstructor, else
@@ -1535,13 +1535,8 @@ fn flatten<M>(
                             acc,
                             registry,
                             &child_records,
-                            // Again the registry's answer, not a new one.
-                            &registry.reading(&child_ty).ok_or_else(|| {
-                                UnfoldError::Unsupported {
-                                    func: func.clone(),
-                                    reason: "a field type the language cannot express",
-                                }
-                            })?,
+                            // The declaration's reading, peeled — not a new one.
+                            child_ty,
                             &field_path,
                             &seg_name(&fr.name),
                             by_ref,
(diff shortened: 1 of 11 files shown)


prebindgen/src/api/core/unfold.rs
```diff
@@ -1460,18 +1461,17 @@ fn flatten<M>(
...
-                    let fr_reading = registry.reading(&fr.ty);
+                    // The declaration carries the field's reading, so nothing is
+                    // looked up and nothing is re-classified.
+                    //
                     // A field's own `Option` makes everything under it nullable,
                     // exactly as an `Option`-returning accessor step does.
-                    let (opt, core) = match fr_reading.as_ref().and_then(|r| r.optional_inner()) {
-                        Some(inner) => (true, inner.origin.syntax.clone()),
-                        None => (false, fr.ty.clone()),
+                    let (opt, core) = match fr.ty.optional_inner() {
```
milyin, line 1469: **[P1] Preserve transparent wrappers when peeling the optional field**

`optional_inner()` can see through a transparent outer `Box`, so a field spelled `Box<Option<Child>>` reaches this branch as `(true, Child)`. But the resulting `PathStep::Field` records only the field name and `optional: true`; the outer `Box` is discarded. The delivery emitter therefore generates the same optional access chain it uses for `Option<Child>` directly, against a value whose Rust type is `Box<Option<Child>>`. It never dereferences the box before applying the optional `map`/`and_then` path, so a decomposed non-`Copy` child produces invalid generated Rust (moving through a shared `Box`, or the wrong receiver type). The old syntax peel did not claim this shape. Please either keep `Box<Option<T>>` whole until the path algebra can represent the transparent wrapper, or add an explicit dereference/reconstruction step and a compile-level regression for a value-form field of that type.

milyin: Confirmed the defect, and it is **real** — but it is pre-existing, not introduced here. Addressed in cc88b0a by rejecting the shape and filing #268 for the fix.

**The bad codegen reproduces.** With a value-form field `Box<Option<ZKeyExpr>>` and a decomposed child:

```rust
let __cb0_obj6: jni::objects::JObject = match &(&__vf0).boxed_kex {
    ::core::option::Option::Some(__n0) => { .. z_keyexpr_as_str(__n0) .. }
    ::core::option::Option::None => jni::objects::JObject::null(),
};
```

Scrutinee `&Box<Option<ZKeyExpr>>`, patterns `Option`'s. Reduced and compiled: `E0308, expected Box<Option<K>>, found Option<_>`. Exactly as you describe.

**It is not new.** I ran that shape on this branch and on the base commit (`0380d59`) in a worktree. The generated `gen.rs` is **byte-identical**, and the whole `value_form` module gives the same tallies on both (33 pass / 8 fail with a non-decomposed child; 41 pass / 0 fail with a decomposed one).

**Why the premise doesn't hold at this line:** `flatten` never used a syntax peel. The old code here was

```rust
let fr_reading = registry.reading(&fr.ty);
match fr_reading.as_ref().and_then(|r| r.optional_inner()) { .. }
```

— already `optional_inner()` on a **reading**, and `Box` is transparent in readings by design (`flat/ty.rs`: `"Box" => return Ok(args.remove(0).kind)`, *"`Box<T>` **is** `T`"*). So `Box<Option<Child>>` arrived as `(true, Child)` before this PR too. The syn peel (`option_inner_type`) lived in `builder.rs::walk_value_form`, where it only fed the `type_kind` dispatch — it never reached `FieldRecord::ty`.

**What this PR does change in that area,** which I should have called out: `Box<Option<Sum>>` now reaches the `sealed_class` arm and errors with the "optional sum needs a present flag" diagnosis, where the base fell through to `TypeKind::Other` and silently emitted one leaf. A better refusal, but a behaviour change.

**So "keep `Box<Option<T>>` whole" would be changing behaviour this PR didn't touch,** and the deref step is the path-algebra work #268 owns. I took the third option the repo's own diagnostic rule points at — reject now, name the issue:

```
field `ZSampleStruct.kex` is `Box<Option<ZKeyExpr>>` — an `Option` behind a
transparent wrapper. The wrapper is invisible to the model and load-bearing in
the generated Rust, and the leaf access path cannot yet say `deref, then match`
(reserved — see .../issues/268). Spell the field `Option<ZKeyExpr>`, or override
it with .field("kex", ...)
```

The guard is precise — an `Option` the *reading* sees but the *syntax* does not spell — so plain `Option<T>` is untouched, and `&Option<T>` reads as `Ref` rather than `Optional` so it never trips. The test asserts the supported spelling still builds alongside the rejection.

One thing worth recording from this: the shape passed the existing suite because **nothing in it compiles the generated Rust** — `contains` assertions went straight over invalid output. #268's acceptance asks for a compile-level regression for that reason.

Regen still byte-identical; no binding in this repo uses the shape.

PR #271, opened 2026-08-01
Don't assume how Rust spells an optional

milyin:
Closes #268.

## The bug

A value-form field spelled `Box<Option<Child>>` emitted a match against a place still typed `&Box<Option<Child>>` — `E0308`. Confirmed against `rustc` on a reduced case.

## The model is not at fault, and the obvious fix would have made it so

`kind` states the **destination-language invariant**. It does not care whether a reference is `&T` or `Cow<'_, T>`, a sequence `[T]` or `Vec<T>`, an optional `Option<T>` or `Box<Option<T>>`. `flat/ty.rs` says so for both wrappers already. Recording a "transparent-wrapper depth" on `TypeRef` — which is where I started — would push a Rust representation detail back into the one thing built to abstract it away.

The defect is on the **interpreting** side. This is `classify off kind, spell off syntax` broken in the direction nothing was watching: the emitter classified off `kind` correctly (*optional ⇒ needs a `Some`/`None` split*) and then **spelled** off `kind` too, assuming the place literally *is* an `Option`.

## The fix costs nothing

A type-ascribed `let` is a coercion site, and deref coercion is transitive **and** a no-op when the types already match:

```rust
let __o0: &::core::option::Option<_> = &(&__vf0).kex;
match __o0 { Some(__n0) => .., None => .. }
```

One shape serves `Option<T>`, `Box<Option<T>>`, `Rc<Option<T>>` and nested wrappers; the plain spelling behaves identically. The payload stays `_` — what it is, is the source's business.

## Borrowed vs owned, and why the tests earned their keep

Applied only where the destructured expression is a **borrow**. `fold_steps` composes a field read to a reference and an accessor call to an owned value, so the two are told apart by `PathStep::is_field`.

My first attempt coerced unconditionally, and two existing tests failed immediately — borrowing an accessor's returned `Option` to coerce it changes ownership for everything downstream that moves the payload. Owned positions therefore keep their direct match, and that is documented rather than silently left: deref coercion applies to references, and moving a payload out of a wrapper is something only some of them permit (`Box` can, `Rc` cannot).

## The audit

Swept the adapter for the same assumption. Only **two** sites destructured a place read from the source: `reach_leaf` (the reported bug) and the `Option<sum>` present-flag split in `struct_out.rs`. Everything else emitting `Option::Some` — 7 sites in `flat_input.rs`, 2 in `wrapper.rs` — is **construction**, a value whose type the emitter owns, which says nothing about the source's representation and is safe.

`reach_optional`'s `.map`/`.and_then` was left alone deliberately: a method call is not a coercion site, and its optional steps are `Call`s returning owned `Option`s (nested conditional hoists are rejected in `flatten`), so a wrapped field cannot reach it. Not restructured speculatively.

## The guard, and what it is not

`destructure_ledger` (`jni/emit/mod.rs`) is a committed census of destructuring sites. It counts **patterns**, not constructions — which is why `flat_input.rs` and `wrapper.rs` sit at zero.

Stated plainly in its own docs: it **cannot tell a coerced destructure from a raw one**, so a count that does not move is not proof of correctness. Its only job is to make a new site impossible to add silently, forcing review to say which of three categories it is. I verified it works by adding a destructure and watching it fail, then reverting.

It is an explicit half-measure standing in for a compile check — see below.

## Verification

- 534 lib tests; `cargo test` and `--all --all-features` (14/14 binaries); clippy `--deny warnings`; fmt with the CLI-only config.
- **Goldens move**, unlike #267 — this change gives up byte-identity by design. The diff is exactly **two coercion sites and nothing else**, verified by reading it. Forced rebuild first.
- covertest-kotlin **48/48** on the JVM. This carries more weight than usual here, since byte-identity is not available as a net.
- The `Box<Option<T>>` output checked against `rustc` by hand.

## Known gap

**Nothing here compiles the generated Rust** — the reason this bug survived years of green tests. #269 owns that and has been updated to name this PR's test as its first customer, with the ledger flagged for reconsideration once a real compile check exists.

Split out rather than smuggled in: **#270** — a *non*-decomposed `Box<Option<T>>` field still fails to resolve, because converter identity keys on the spelling. Same root cause, different layer (it reaches `TypeKey` and resolution), and it is what the two hardcoded `TypeKey == "Box < String >"` matches in `trait_impl.rs` are already working around by hand.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/mod.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/mod.rs
@@ -87,33 +87,54 @@ mod destructure_ledger {
     //! of them permit (`Box` can, `Rc` cannot), so a site that must MOVE keeps
     //! its direct match and is correct only for representations that allow it.
     //!
+    //! ## Coverage
+    //!
+    //! The table is checked against the **directory listing**, not kept by
+    //! hand. A hand-listed census had already let `convert.rs` sit outside the
+    //! guard entirely with two uncounted destructures, and would have let any
+    //! new emitter module do the same — "a new destructure cannot be added
+    //! silently" is only true if a new *file* cannot be either.
+    //!
     //! To change it deliberately: update the table below in the same commit,
     //! and say in review which category the new site is.
 
     use proc_macro2::{Delimiter, TokenStream, TokenTree};
 
-    /// `(file, destructuring-pattern count)`.
+    /// `(file, destructuring-pattern count)` — **every** `.rs` in this
+    /// directory, checked against the directory listing so a new emitter module
+    /// cannot sit outside the census.
     const LEDGER: &[(&str, usize)] = &[
+        ("callback.rs", 0),
+        // The `Option<T>` output converters' niche and boxed-primitive arms.
+        // They destructure `v`, the converter's own parameter — whose Rust type
+        // is the crossing's SPELLING. Correct today only because a wrapped
+        // spelling gets no converter at all (#270); if that is fixed and
+        // `Box<Option<T>>` becomes a crossing, these two need coercing.
+        ("convert.rs", 2),
         // 2 fn-return matches (owned), 2 leaf reaches (one coerced, one owned
         // accessor return), 1 owned identity move, 1 emitter-bound local.
         ("delivery.rs", 6),
+        ("flat_input.rs", 0),
+        // This file. The spellings in `the_census_is_spelling_independent` are
+        // string literals, which tokenize as one `Literal` and are never walked.
+        ("mod.rs", 0),
+        ("names.rs", 0),
         // Both coerced: the `Option<sum>` present-flag split, and the nested
         // plan's present-flag split.
         ("struct_out.rs", 2),
-        // Constructions only.
-        ("flat_input.rs", 0),
+        ("sum_out.rs", 0),
+        ("vec_build.rs", 0),
         ("wrapper.rs", 0),
     ];
 
-    /// Count `Some ( … ) =>` patterns inside `quote!` bodies.
+    /// Count `Some ( … ) =>` patterns inside `quote!` / `parse_quote!` bodies.
     ///
     /// `in_quote` is what keeps the emitter's own `if let Some(..)` out of the
-    /// count: only tokens below a `quote!` are emitted Rust.
+    /// count: only tokens below one of those macros are emitted Rust.
     fn count(ts: TokenStream, in_quote: bool, n: &mut usize) {
         let toks: Vec<TokenTree> = ts.into_iter().collect();
         let mut i = 0;
         while i < toks.len() {
-            // `quote! { … }` / `quote!( … )` — descend, now emitting.
             if let TokenTree::Ident(id) = &toks[i] {
                 let bang =
                     matches!(toks.get(i + 1), Some(TokenTree::Punct(p)) if p.as_char() == '!');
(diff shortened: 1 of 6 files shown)


prebindgen/src/api/lang/jnigen/jni/emit/mod.rs
```diff
@@ -28,3 +28,97 @@ pub(crate) use struct_out::*;
...
+        let mut drift: Vec<String> = Vec::new();
+        for (file, expected) in LEDGER {
+            let src = std::fs::read_to_string(std::path::Path::new(dir).join(file))
+                .unwrap_or_else(|e| panic!("read {file}: {e}"));
+            // A pattern arm is `Option::Some(<binding>) =>`; a construction has
+            // no `=>` after its closing paren. Whitespace is stripped so line
+            // wrapping cannot hide a site.
+            let bare: String = src.chars().filter(|c| !c.is_whitespace()).collect();
+            let found = bare
+                .match_indices("option::Option::Some(")
```
milyin: **[P2] Count unqualified `Some` patterns in the destructure ledger**

This search only recognizes fully qualified `option::Option::Some(...)` arms, but `struct_out.rs` already has another emitted destructure written as `match &#value { Some(#cbind) => ... }` in the optional `PlanFieldKind::Nested` branch. Consequently the claimed census reports one `struct_out.rs` site while the file actually emits two, and adding another unqualified source-place match would leave `found` unchanged and pass this guard silently. That defeats the ledger’s stated purpose of making every new destructure move the count. Please make the census recognize both qualified and unqualified patterns (ideally by parsing the quoted syntax rather than matching one spelling), and include the existing nested branch in the expected count/regression.

milyin: Correct on both counts, and the second one was a real bug. Fixed in 51b1d60.

**The site was live.** `struct_out.rs`'s nested-plan branch emitted `match &#value { Some(#cbind) => .. }` against a source place — an actual unfixed instance of this PR's bug, not just a miscount. It shows up in `Annotated_to_JObject` in covertest's committed bindings, so it was exercised code. My audit missed it because I grepped for `option::Option::Some(`, and the ledger inherited the same blind spot.

That is the worst place for this particular guard to be blind: the thing it exists to catch was hiding in the one spelling it did not match, and any new unqualified destructure would have passed it silently. A guard that can be evaded by choosing a different spelling of the same pattern is not a guard.

**The census now walks tokens**, as you suggested. It descends only into `quote!` bodies — which is what "emitted Rust" means, and which keeps the emitter's own `if let Some(x)` out of the count — and recognizes `Some ( … ) =>` whatever path prefix it carries:

```rust
if in_quote && id == "Some" {
    let parens = matches!(toks.get(i + 1), Some(TokenTree::Group(g))
        if g.delimiter() == Delimiter::Parenthesis);
    let fat_arrow = /* '=' then '>' */;
    if parens && fat_arrow { *n += 1; }
}
```

`the_census_is_spelling_independent` pins it on all three spellings (`Some`, `Option::Some`, `::core::option::Option::Some`) **and** on the two things that must not count — a construction inside `quote!`, and a host-code `if let Some(x)`. I'd rather prove that than assert it in the docs, given the first version's docs also claimed to be complete.

`struct_out.rs` goes **1 → 2** in the table: the count was wrong, not only the instrument.

Goldens move by exactly the third coercion site; covertest 48/48 on the JVM. Full suite, clippy and fmt green.

One thing I'd note for the record: this is the second time in two PRs that a text scan has been the wrong instrument for a guard in this repo. The boundary ledger avoided it by walking tokens from the start, and I should have followed that precedent rather than reinventing a weaker one.

prebindgen/src/api/lang/jnigen/jni/emit/mod.rs
```diff
@@ -28,3 +28,184 @@ pub(crate) use struct_out::*;
...
+    //! of them permit (`Box` can, `Rc` cannot), so a site that must MOVE keeps
+    //! its direct match and is correct only for representations that allow it.
+    //!
+    //! To change it deliberately: update the table below in the same commit,
+    //! and say in review which category the new site is.
+
+    use proc_macro2::{Delimiter, TokenStream, TokenTree};
+
+    /// `(file, destructuring-pattern count)`.
+    const LEDGER: &[(&str, usize)] = &[
```
milyin, line 106: **[P2] Cover every emitter file in the census**

The token walker fixes the spelling blind spot, but the census still runs only over these four hand-listed files. `convert.rs` already emits two genuine destructures via `syn::parse_quote!` (`Some(value) =>` at lines 323 and 335), and neither is counted; likewise a destructure added to `callback.rs`, `sum_out.rs`, or a new emitter module would pass without moving this ledger. Thus the adapter-wide “new destructure impossible to add silently” invariant is still not enforced. Please include `convert.rs` with count 2 and make the test verify that every emitter `.rs` file is represented (or discover the files and compare their per-file counts), so adding a site or module cannot fall outside the guard.

milyin: Right again, and the same shape of miss as the first: the guard's scope was narrower than its claim. Fixed in f50de7d.

Verified your numbers — running the walker over every file in the directory gives `convert.rs: 0 -> 2`, `delivery.rs: 0 -> 6`, `struct_out.rs: 0 -> 2`, rest zero. So `convert.rs` was outside the guard entirely, and a destructure added to `callback.rs`, `sum_out.rs` or a new module would have moved nothing.

**The table is now checked against the directory listing**, which is the property that actually matters: "a new destructure cannot be added silently" is only true if a new *file* cannot be either.

```rust
assert_eq!(listed, on_disk,
    "the destructure census must cover every emitter file — add the new \
     module to LEDGER with its count");
```

Both properties proved by doing them rather than claiming them:

- dropped a `probe_new.rs` emitting a destructure into the directory → fails with the listing in the message;
- appended an unqualified destructure to `sum_out.rs` → `sum_out.rs: 0 -> 1`;
- reverted both → green.

**`convert.rs` is recorded at 2 with what it actually is**, rather than waved through as safe. Those two arms destructure `v`, the converter's own parameter, whose Rust type is the crossing's *spelling*. They are correct today **only** because a wrapped spelling gets no converter at all — which is #270, not a property of this code. The ledger comment says exactly that, so if #270 lands, the next person knows these two need coercing rather than rediscovering it:

```rust
// The `Option<T>` output converters' niche and boxed-primitive arms.
// They destructure `v`, the converter's own parameter — whose Rust type
// is the crossing's SPELLING. Correct today only because a wrapped
// spelling gets no converter at all (#270); if that is fixed and
// `Box<Option<T>>` becomes a crossing, these two need coercing.
("convert.rs", 2),
```

No emitted output changes here, so goldens are untouched. Suite, clippy and fmt green; CI re-running.

Both catches were the same failure on my side: I wrote a guard and then described it as stronger than it was — first on spelling, now on scope. The docs now state the coverage property explicitly rather than leaving it implied by a list.

PR #272, opened 2026-08-01
The selector reads the model instead of guessing types by name

milyin:
Closes #270.

## The issue was wrong about its own cause

I filed #270 guessing this was `TypeKey` identity — two spellings of one meaning not sharing a converter. It isn't. The selector decided **what a type is** by manufacturing a string from its spelling and comparing it, and then rebuilt the type it generates **by name**:

| Site | What it did | What the model already said |
|---|---|---|
| `with_first_arg` | rebuilt a wildcard off the spelling's last segment — `Box<Option<T>>` → `Box<_>` | `kind` is `Optional` |
| `pat_match(pat, "Option < _ >")` ×10 | dispatch by **string comparison** | `TypeKind::{Optional,Sequence,Ref,Unit}` |
| `parse_quote!(Option<#t1>)` ×7 | the generated signature **rebuilt by name** | `origin.syntax` |
| `option_inner_type` / `vec_inner_type` | peel by segment ident | `optional_inner()` / `sequence_elem()` |
| `TypeKey == "Box < String >"` ×2 | one hardcoded arm per direction | `kind` is `Str` |

`select_output_type` was the clearest: `convert_crossing` **fetched the reading and threw it away**, passing a bare spelling.

## The rule

`classify off kind, spell off syntax`, applied to converter construction. #268 fixed an emitter that classified correctly and then *spelled* off `kind`; this is the same rule broken the other way — classifying off the spelling.

Dispatch now comes from `TypeKind` (a `WrapperShape`, not a rendered string) and the signature from `origin.syntax`, so a `Box<Option<T>>` converter takes a `Box<Option<T>>` instead of declaring `Option<T>` and mismatching its own call site.

## Representations: state the requirement, let rustc check

Rather than listing wrappers by name — which is the guessing being removed:

- **Input** builds the canonical value and `.into()`s the spelling. Reflexive when they match.
- **Output** binds the canonical *out of* the spelling. This direction is genuinely asymmetric and I did not pretend otherwise: the bodies **move** the payload, and std has no trait for `Box<T> → T`. So `read_as_canonical` asks the one honest spelling question — is this already canonical? — and derefs once if not. Anything not deref-movable is a rustc error naming the type.

Both `Box<String>` arms are **deleted** — the evidence the mechanism generalizes. `kind == Str` covers `Box<String>` and `Cow<'_, str>` alike; the output body already read through either (`v.as_str()` derefs), only the dispatch was spelling-keyed.

## Two corrections the tests forced

- **`Rc` is not transparent.** Only `Box` and `Cow` are erased by the model. My first doc comments claimed `Rc<String>` would work; the model rejects it outright. Comments and tests corrected.
- **A bare `[T]` is `Sequence` but unsized.** Dispatching on `kind` alone tried to build `fn f(..) -> [T]`, which broke the example crates. `is_unsized_spelling` is the guard — a spelling question in the same family as the existing `decoded_vec_satisfies`.

## Result

`pat_match` is gone, and `with_first_arg` / `ref_wildcard` with it. Boundary ledger **135 → 130**, all of it `selector.rs` (8 → 3) — the #229 L4 "layer questions" item, whose stated payoff was retiring `option_inner_type` / `vec_inner_type` from the selector.

## Verification

- 609 lib tests; `cargo test --all --all-features` 14/14; clippy `--deny warnings`; fmt.
- **The generated Rust compiles.** `regen-check.sh` builds the four example crates, which `include!` their generated bindings — so this change's output type-checks for every shape those examples use. That is a real compile check, and worth noting given #269.
- covertest-kotlin **48/48** on the JVM.
- `Box<Option<T>>` and `Box<String>` output hand-checked against `rustc` on reduced cases.

**Goldens move**, and the diff is large but mechanical: `.into()` wrappers, canonical bindings, and two new `#[allow]`s (`unused_parens`, `clippy::useless_conversion`) — the degenerate form a representation-agnostic converter takes for the plain spelling. Suppressing those per-shape would mean asking which spelling this is. The `Box<String>` converter is renamed because its signature now comes from the source's spelling rather than a hardcoded `::std::boxed::Box<::std::string::String>`.

The `convert.rs` note in #268's destructure census predicted this change and is now resolved: those two destructures are correct *by construction*, because the caller binds the canonical `Option<T>` before handing the body `v`.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/selector.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/selector.rs
@@ -83,6 +83,21 @@ impl Declarations {
                     return Some(c);
                 }
             }
+            // An optional BORROW is the deep handler's alone. It declined —
+            // either the inner is not a handle (then the shallow handler below
+            // is right, and only for the canonical spelling) or the spelling
+            // carries a wrapper it cannot bridge. The shallow handler cannot
+            // tell those apart and would decode the jlong as a `*mut &T`, so a
+            // wrapped optional borrow stops here rather than resolving wrong.
+            if inner.borrow_target().is_some() {
+                let canonical: syn::Type = {
+                    let b = &inner.origin.syntax;
+                    syn::parse_quote!(Option<#b>)
+                };
+                if syntax.to_token_stream().to_string() != canonical.to_token_stream().to_string() {
+                    return None;
+                }
+            }
             let inner_ty = inner.origin.syntax.clone();
             if let Some(mut c) =
                 self.input_wrapper_shape(WrapperShape::Optional, syntax, &inner_ty, registry)
--- a/prebindgen/src/api/lang/jnigen/jni/selector.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/selector.rs
@@ -1,37 +1,7 @@
 //! Structural converter-selection policy for [`Declarations`].
 
-use super::*;
+use super::{trait_impl::WrapperShape, *};
 use crate::api::core::registry::Conversions;
-
-/// Clone a single-type-arg generic (`Option<X>` / `Vec<X>` / any `Path<X, …>`)
-/// replacing its last segment's first type argument with `repl` — yielding the
-/// canonical shape (`Option<_>`) the built-in wrapper handlers key on, with the
-/// type's own path/qualification preserved exactly.
-fn with_first_arg(ty: &syn::Type, repl: syn::Type) -> syn::Type {
-    let mut out = ty.clone();
-    if let syn::Type::Path(tp) = &mut out {
-        if let Some(seg) = tp.path.segments.last_mut() {
-            if let syn::PathArguments::AngleBracketed(ab) = &mut seg.arguments {
-                for a in ab.args.iter_mut() {
-                    if let syn::GenericArgument::Type(t) = a {
-                        *t = repl;
-                        break;
-                    }
-                }
-            }
-        }
-    }
-    out
-}
-
-/// Clone a reference type replacing its referent with the `_` wildcard,
-/// preserving the lifetime and mutability (`&'a T` → `&'a _`, `&mut T` →
-/// `&mut _`) so the reconstructed pattern matches what the enumerator emitted.
-fn ref_wildcard(r: &syn::TypeReference) -> syn::Type {
-    let mut pr = r.clone();
-    *pr.elem = syn::parse_quote!(_);
-    syn::Type::Reference(pr)
-}
 
 /// Whether a decoded `Vec<T>` local can be borrowed where `referent` is expected.
 ///
@@ -52,6 +22,20 @@ fn decoded_vec_satisfies(referent: &syn::Type) -> bool {
     }
 }
 
+/// Whether a spelling has no size, so no by-value converter can name it.
+///
+/// A **spelling** question, like [`decoded_vec_satisfies`]: `[T]` and `Vec<T>`
+/// are one concept to the model — both `Sequence` — and Rust can return only
+/// one of them. A bare slice is reached exclusively through a borrow, whose own
+/// arm handles it; claiming it here would generate `fn f(..) -> [T]`.
+///
+/// `str` is the same shape of fact and is handled the same way, one layer up:
+/// its terminal arm resolves it to the borrowed `&str` converter rather than
+/// pretending an owned `str` exists.
+fn is_unsized_spelling(ty: &syn::Type) -> bool {
+    matches!(ty, syn::Type::Slice(_))
+}
+
 impl Declarations {
     /// Select the input converter for `ty`: terminals, user wrappers, then
     /// built-in structural wrappers.
@@ -62,13 +46,14 @@ impl Declarations {
     ) -> Option<ConverterImpl<KotlinMeta>> {
         use crate::api::core::flat::RefMode;
 
-        // The spelling, for the wildcard patterns this selector builds. Those are
-        // the adapter's own — `Option<_>`, `&_` — so composing them from tokens is
-        // spelling, not reasoning. What the type *is* comes from `ty` below.
+        // What the type IS comes from `kind`; what generated Rust must SPELL it
+        // comes from here. The converter yields this spelling, so a
+        // `Box<Option<T>>` crossing produces a `Box<Option<T>>` — the shape it
+        // is dispatched as no longer decides what it is called.
         let syntax = &ty.origin.syntax;
 
         // 1. Terminal categories (incl. the terminal user-wrapper lookup).
-        if let Some(c) = self.input_terminal(syntax, registry) {
+        if let Some(c) = self.input_terminal(ty, registry) {
             return Some(c);
         }
         // 3. Built-in wrapper shapes, read one layer at a time rather than as a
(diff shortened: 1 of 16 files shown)


prebindgen/src/api/lang/jnigen/jni/trait_impl.rs
```diff
@@ -612,33 +619,90 @@ pub(crate) fn build_handle_destructor_items(
...
+///
+/// A representation that is not deref-movable fails to compile, naming the
+/// type. That is the same contract as the input side: the adapter states what
+/// it needs of a representation and rustc enforces it, rather than the adapter
+/// guessing which wrappers exist.
+fn read_as_canonical(produced: &syn::Type, canonical: &syn::Type) -> TokenStream {
+    if produced.to_token_stream().to_string() == canonical.to_token_stream().to_string() {
+        quote!(v)
+    } else {
+        quote!((*v))
```
milyin: **[P1] Do not resolve transparent spellings that this read cannot canonicalize**

The model erases `Cow` as well as `Box`, but every non-canonical output spelling is lowered to `*v`. That works for one `Box<T>` because Rust permits moving out of a box; it does not work for `Cow`. For example, a valid `Cow<…, Option<String>>` now classifies as `Optional`, resolves a converter, and emits `let v: Option<String> = *v`, which fails with E0507 because a payload cannot be moved through `Cow::deref` (`into_owned()` is required). `Cow<…, [String]>` similarly selects the sequence converter but cannot become the expected `Vec<String>` via `*v`. The new Cow string test does not cover this because that terminal converter only borrows with `as_str()`.

Recursive transparent wrappers have the same problem: `Box<Box<Option<T>>>` is also classified as `Optional`, while one dereference leaves `Box<Option<T>>`; on input, one `.into()` likewise cannot chain through both boxes. These are accepted source spellings, so deferring the mismatch to generated-code compilation turns successful resolution into invalid output. Please either canonicalize each supported spelling correctly/recursively (`Cow::into_owned`, repeated box moves), or reject unsupported representations during selection, and add compile-level cases for at least `Cow<…, Option<String>>` and a nested `Box`.

milyin: Both halves confirmed and fixed in 91493c7 — and both were live, not theoretical.

**Reproduced first.** Each shape resolved and emitted `let v: Option<String> = (*v);`:

```
PROBE Cow<Option<String>>:      RESOLVED -> ["let v: Option<String> = (*v);"]
PROBE Box<Box<Option<String>>>: RESOLVED -> ["let v: Option<String> = (*v);"]
```

and against `rustc`, exactly the two errors you named — `E0507: cannot move out of dereference of Cow<'_, Option<String>>` and `E0308` for the nested box.

**Your framing of the severity is the part I had wrong.** My plan said "state the requirement and let rustc enforce it", and I carried that over from the input side without noticing it does not hold here: successful resolution is what tells a binding its type is supported. Deferring to the consumer's build turns a diagnosis into a broken build with nothing named.

So `box_layers_to` counts the layers it can actually undo and returns `None` otherwise, and the handlers propagate that with `?`. An unsupported representation is now an unresolved crossing naming the type:

```
PROBE Cow<Option<String>>:      rejected (2 required type(s) could not be resolved)
PROBE Box<Box<Option<String>>>: RESOLVED -> ["let v: Option<String> = (*(*v));"]
```

**Compile-level coverage.** `boxed_note_echo` returns `Box<Box<Option<String>>>` from covertest, whose binding is `include!`d and compiled — nested deliberately, because one dereference leaves a type that still type-checks and would only fail at the build. `plain_note_echo` sits beside it as the control, and the two externs come out identical, which is the actual invariant:

```kotlin
external fun boxedNoteEcho(note: String?, errorSink: Any): String?
external fun plainNoteEcho(note: String?, errorSink: Any): String?
```

Generated Rust:
```rust
pub(crate) unsafe fn Box_Box_Option_String_to_JString_299999e0<'a>(
    env: &mut jni::JNIEnv<'a>, v: Box<Box<Option<String>>>,
) -> ::core::result::Result<jni::objects::JString<'a>, __JniErr> {
    Ok({ let v: Option<String> = (*(*v)); ... })
}
```

A `Cow` payload **cannot** be a compiled fixture — it has to be refused, and only a unit test can assert a refusal. `a_transparent_wrapper_is_bridged_only_where_it_can_be` covers all three: one box, two boxes, and the `Cow` rejection.

**The fixture found a second bug, which I have not fixed here.** My first attempt put the shape on a data-class field, and covertest's Kotlin failed to compile: `is_option_type` — `path_tail_is(ty, "Option")` — decides Kotlin nullability, so a wrapped `Option` **parameter** renders non-null while the identical-meaning plain one renders `String?`. Same root cause, different layer, ten call sites that mostly hold a bare `&syn::Type` with no registry in reach. That is #229 L4 "layer questions" work rather than a local patch, so it is **#273** and the commit says so rather than leaving it as a silent limitation.

Ledger 130 → 131 — `box_layers_to` matches `syn::Type::Path`, an honest new spelling site. Still 135 → 131 net. Suite, clippy, fmt green; covertest 48/48 on the JVM.

prebindgen/src/api/lang/jnigen/jni/selector.rs
```diff
@@ -62,45 +46,57 @@ impl Declarations {
...
         //    whole stack: each arm handles exactly one, and hands the rest back
         //    through `subs` to be selected on its own. Peeling further here would
         //    claim a shape this selector does not emit.
         if let Some(inner) = ty.optional_inner() {
-            // `Option<&T>` tries the DEEP `Option<&_>` (borrowed-handle →
-            // `Option<OwnedObject<T>>`) before the shallow `Option<_>`; the shape
+            // `Option<&T>` tries the DEEP `OptionRef` (borrowed-handle →
+            // `Option<OwnedObject<T>>`) before the shallow `Optional`; the shape
             // that resolves correctly wins.
             if let Some(target) = inner.borrow_target() {
```
milyin, line 67: **[P1] Refuse wrapped borrow shapes that bypass the canonical bridge**

Kind-based dispatch now also makes `Box<&Handle>` and `Box<Option<&Handle>>` reach these borrow handlers, but neither handler produces the spelled type and neither uses `box_layers_to`. `input_borrow` reuses the inner converter, yielding an owned `T`; `input_option_ref` yields `Option<OwnedObject<T>>`. The call-site adaptation in `emit/wrapper.rs` only adds `&decoded` for a syntactic `Type::Reference` and `.as_deref()` for a syntactic `Option<&T>`. With either outer `Box`, both checks miss, so resolution succeeds and the generated call passes `T` or `Option<OwnedObject<T>>` to a parameter expecting `Box<&T>` or `Box<Option<&T>>`.

This is separate from the Kotlin-nullability issue filed as #273: the Rust call itself is ill-typed. Please either bridge these specialized borrow representations at the call site, or require the direct `&T` / `Option<&T>` spellings before selecting `Borrow` / `OptionRef` (letting wrapped forms remain unresolved), with compile-level regressions for both shapes.

milyin: Confirmed and fixed in 2f20588. Reproduced first — both resolved and passed an owned `T` straight through:

```
PROBE &ZThing:               RESOLVED -> ["myflat::z_take(&t);"]
PROBE Box<&ZThing>:          RESOLVED -> ["myflat::z_take(t);"]     ← ill-typed
PROBE Box<Option<&ZThing>>:  RESOLVED -> ["myflat::z_take(t);"]     ← ill-typed
```

I took the structural route rather than adding two guards, since the standing question is what adding `Rc` later would cost.

**The model owns the set; the adapter owns the operations.** `core::flat` now exposes `TRANSPARENT_WRAPPERS`, and `lower_path` consults it instead of naming `Box` and `Cow` in its own match — so the model cannot drift from its own list. The adapter has one table:

```rust
const WRAPPER_OPS: &[WrapperOps] = &[
    WrapperOps { name: "Box", read: Some(|e| quote!((*#e))), build: Some(|e| quote!(Box::new(#e))) },
    WrapperOps { name: "Cow", read: None, build: None },
];
```

`box_layers_to` — which hardcoded `"Box"` and could only *count* — is gone. `bridge_layers` returns the chain of ops, and read/build ask each layer whether it supports that direction instead of assuming every layer behaves like a box.

**Adding `Rc` is one entry and one row**: `read: None` (its payload cannot be moved out), `build: Some(Rc::new)`. I verified that rather than claiming it — adding `"Rc"` to the model makes `every_erased_wrapper_has_ops` fail with exactly the instruction needed:

```
the model erases ["Rc"], and this adapter has no `WrapperOps` row for them — add one
(`read`/`build` may be `None` when the representation does not allow it, which refuses
the shape instead of mis-generating it)
```

So the second step can't be discovered later by a mis-generated binding.

**The borrow shapes** ask `is_canonical_spelling` and refuse anything wrapped. Framed as "these converters do not produce the spelled type, so there is nothing to bridge" rather than "reject Box" — the reason is the shape, not the wrapper.

**That refusal opened a fall-through my probe caught**, which is worth flagging because it was not in your list: with `OptionRef` declining, `Box<Option<&T>>` reached the *shallow* optional handler, which decoded the handle as `*mut &T` — well-typed and semantically wrong, so it would have passed a compile check. An optional borrow is the deep handler's alone; it now stops there unless the spelling is canonical.

```
PROBE &ZThing:               RESOLVED -> ["myflat::z_take(&t);"]
PROBE Option<&ZThing>:       RESOLVED -> ["myflat::z_take(t.as_deref());"]
PROBE Box<&ZThing>:          rejected
PROBE Box<Option<&ZThing>>:  rejected
```

`a_wrapped_borrow_has_nothing_to_bridge_and_refuses` pins all four, canonical forms included — a guard that also refused those would be worse than none.

Ledger 131 → 130: the `syn::Type::Path` match moved into `core/flat`, where deciding what syntax means belongs. **Goldens byte-identical** this round — no existing shape was reachable through these paths. Suite, clippy, fmt green; covertest 48/48.

PR #274, opened 2026-08-01
Ask the model whether a type is optional, not the spelling

milyin:
Closes #273.

## The bug, proven

Kotlin nullability was decided by `is_option_type` — literally `path_tail_is(ty, "Option")`. The model **erases** transparent wrappers, so `Box<Option<String>>` classifies `Optional` and that check answers `false`.

Verified against the base commit by putting the wrapped parameter through it:

```kotlin
external fun boxedNoteEcho(note: String,  errorSink: Any): String?   // 237ee8c
external fun plainNoteEcho(note: String?, errorSink: Any): String?
```

**Not cosmetic.** A non-null Kotlin parameter for an optional value is a wrong *contract*: Kotlin rejects `null` at the call site, so the absent case becomes unexpressible.

## The fix has one home

`Conversions` gains `is_optional` / `optional_inner` / `sequence_elem` / `is_optional_borrow` as **default methods over `reading`** — the way `input_entry` and `output_entry` already sit over `conversion`.

That is deliberate rather than eight copies of `reading(ty)?.optional_inner()`: adding a transparent wrapper (an `Rc`) becomes a change to `TRANSPARENT_WRAPPERS` and **nothing else**, because every site asking the question follows automatically. Same shape as #272's `WRAPPER_OPS`, one layer out.

All eight sites already had a registry in scope, so nothing needed plumbing.

**Two are not `?` decisions**, and are worth flagging for review: `render.rs` picks a handle's **lock mode** and `emit/callback.rs` picks a **jvalue union member** — a deadlock and an ABI mismatch respectively, not a missing question mark. covertest's handle-locking and callback sections cover both.

## The cluster and the loop

`struct_plan::classify_field` peeled the same field **seven times** by name, right beside the nullability decision it would then disagree with. One reading at the top serves all seven.

`builder.rs`'s niche-depth loop peeled a spelling and re-keyed each result; it walks the reading instead, which drops the re-lookup.

`is_option_ref` is **deleted** — both halves read the spelling, and nothing called it once `is_optional_borrow` existed.

## The census

A token-walking count of the remaining spelling helpers under `api/lang/jnigen`, counted **by callee name** so a qualified or aliased call cannot hide. That is the #271 lesson: its first guard matched text, missed a bare spelling, and that spelling turned out to be a live bug.

Proven both ways — a call added to a zero-count file fails, and so does a brand-new module:

```
  jni/render.rs: 0 -> 1
  jni/probe_new.rs: 0 -> 1
```

`flat_input.rs`'s 20 calls stay **listed** as the L4 remainder rather than bundled here; the census is what keeps them visible instead of forgotten.

## Verification

- 543 lib tests; `cargo test --all --all-features` 14/14; clippy `--deny warnings`; fmt.
- **The regen diff is the interesting part.** The fixture's parameter is now `Box<Option<String>>`, and `covertest.kt` **is not in the diff at all** — wrapping a parameter changes nothing the JVM can see, which is the invariant. The Rust converter correctly takes `Box<Option<String>>`; the only Kotlin change is my own doc comment.
- covertest-kotlin **48/48** on the JVM. This carries real weight here: a wrong `?` is a Kotlin *compile* error in the harness, which is how the bug was found in the first place.
- Boundary ledger 130 → 129.

## Scope left open

`emit/flat_input.rs` (20 calls) is the L4 "layer questions" remainder — a separate consumer, and bundling it would have made one review of both impossible. It is in the census with its count, so it cannot drift quietly.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/mod.rs
+++ b/prebindgen/src/api/lang/jnigen/mod.rs
@@ -52,3 +52,151 @@ pub use jni::{
 // unchanged (`KotlinFile` aliases the model's `KtFile`).
 pub use crate::api::gen::kotlin::KtFile as KotlinFile;
 pub use crate::api::gen::kotlin::WriteKotlinError;
+
+#[cfg(test)]
+mod spelling_census {
+    //! A committed census of every place this adapter asks a **spelling** what
+    //! a type's layers are, instead of asking the model.
+    //!
+    //! `is_option_type` is `path_tail_is(ty, "Option")`, and
+    //! `option_inner_type`/`vec_inner_type` peel by last path segment. The model
+    //! **erases** transparent wrappers — `Box<T>` *is* `T`, and so is
+    //! `Cow<'_, T>` — so every one of these answers "no layer" for a type the
+    //! model says is `Optional` or `Sequence`.
+    //!
+    //! That is how #273 happened: Kotlin nullability was decided this way, so a
+    //! `Box<Option<String>>` **parameter** rendered non-null while the
+    //! identical-meaning `Option<String>` rendered `String?` — and a non-null
+    //! parameter for an optional value makes the absent case unexpressible.
+    //! `Conversions::{is_optional, optional_inner, sequence_elem,
+    //! is_optional_borrow}` ask the model, and every site with a registry in
+    //! scope should use those.
+    //!
+    //! ## What this is for
+    //!
+    //! The counts go **down**. A file at zero has been migrated and must not
+    //! regress; a file above zero is remaining work, and #229's L4 "layer
+    //! questions" is where it is tracked. Either way a NEW call cannot appear
+    //! without moving a number, which is what stops the next site from reaching
+    //! for the spelling because it was the easiest thing in scope.
+    //!
+    //! It does not say the remaining calls are wrong *today* — some may be
+    //! legitimate spelling questions, the way `decoded_vec_satisfies` and
+    //! `is_unsized_spelling` are. It says each one is a decision someone made,
+    //! and moving the number is what puts it in front of review.
+    //!
+    //! ## Why it walks tokens
+    //!
+    //! A text scan is the wrong instrument for anything with more than one
+    //! spelling: `option_inner_type(..)`,
+    //! `types_util::option_inner_type(..)` and a `use`-aliased call are the same
+    //! call. #271's first guard matched text and missed a bare `Some(..)` that
+    //! turned out to be a live bug, so this counts **call expressions by callee
+    //! name**, whatever path qualifies them.
+
+    use proc_macro2::{Delimiter, TokenTree};
+
+    /// The helpers that read a spelling where the model has the answer.
+    const SPELLING_HELPERS: &[&str] = &[
+        "is_option_type",
+        "is_option_ref",
+        "option_inner_type",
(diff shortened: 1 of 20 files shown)


prebindgen/src/api/lang/jnigen/mod.rs
```diff
@@ -52,3 +52,151 @@ pub use jni::{
...
+    //! It does not say the remaining calls are wrong *today* — some may be
+    //! legitimate spelling questions, the way `decoded_vec_satisfies` and
+    //! `is_unsized_spelling` are. It says each one is a decision someone made,
+    //! and moving the number is what puts it in front of review.
+    //!
+    //! ## Why it walks tokens
+    //!
+    //! A text scan is the wrong instrument for anything with more than one
+    //! spelling: `option_inner_type(..)`,
+    //! `types_util::option_inner_type(..)` and a `use`-aliased call are the same
```
milyin, line 92: **[P2] Make the census actually account for aliased helper calls**

Qualified calls are covered because their final identifier remains `option_inner_type`, but a `use` alias is not: `use ...::option_inner_type as peel_optional; peel_optional(ty)` presents `peel_optional` to `count`, which is absent from `SPELLING_HELPERS`. In a new or zero-count file this therefore passes without moving any number, contradicting the explicit claim here and the guard’s “a NEW call cannot appear” purpose.

Please either resolve tracked names through `use` trees, reject aliases of these helpers explicitly, or narrow the documented guarantee. Given this census exists to avoid the spelling blind spot from #271, I would add a regression containing an aliased call and require it to count/fail.

milyin: Correct, and it is worse than hypothetical — the sibling adapter does exactly this today:

```rust
// api/lang/cbindgen/mod.rs:106
use ...::{first_type_arg, is_option_type as is_option, is_result_type as is_result, is_unit, ...};
```

and calls `is_option(ty)` at a dozen sites. So aliasing these exact helpers is an **established idiom in this repo**, and anyone writing jnigen code in that style would have walked straight through a guard whose entire claim is that they cannot. Fixed in 72cb58e.

**The overclaim was the real catch.** My docs said the census counts calls "whatever path qualifies them" — true of qualification, false of aliasing. That is the same failure as #271's first guard: a guard that describes itself as stronger than it is.

`tracked_names` now collects the local names a file can call a helper by — its own, plus every `<tracked> as <alias>` — and the count runs against that set. Found in **tokens**, not by parsing `use` items, for the reason the module docs already give: a token walk sees macro bodies and nested `use` groups alike, and cbindgen's real form is a rename *inside a group*.

`the_census_sees_through_a_use_alias` pins all four positive forms and both negatives:

| form | counts |
|---|---|
| `option_inner_type(t)` | 1 |
| `types_util::option_inner_type(t)` | 1 |
| `use … as peel_optional;` → `peel_optional(t)` | 1 |
| `use {a, is_option_type as is_option};` → `is_option(t)` | 1 |
| `use x::option_inner_type;` (import, no call) | 0 |
| `use x::other as option_probe;` → `option_probe(t)` | 0 |

And end to end, since a table in a unit test is not the same as the real scan: dropping an aliased call into a real zero-count file moves it.

```
  jni/render.rs: 0 -> 1
```

Test-only change; no generated output moves. Suite, clippy and fmt green.

PR #278, opened 2026-08-02
Core's plan leaves carry the reading

milyin:
Second of three for #275 (#276 was the first). The last part deletes `is_optional` and closes the issue.

## Why

`UnfoldLeaf::out_ty` and `FoldLeaf::ty` were `syn::Type`, so an adapter consuming a decomposition had to hand the spelling back to the registry to learn what it meant — the round trip #263 removed from `api/core`, surviving in the plans.

And the reading was **discarded at construction**: `flatten` peels `TypeRef`s and stored `origin.syntax`.

## `flat` composes its own types

Most producers just stop discarding. Five genuinely compose a type no source wrote — the borrow of a value, two `Option` layers, a presence flag, a selector — and **`Flat::classify` is not available** for them: it lowers *source syntax*, and `classify_has_no_caller_outside_the_registry` keeps it that way. Correctly, in my view: composing is a different act from classifying.

So the model composes:

```rust
TypeRef::{borrowed, optional, scalar, named}
```

Each builds `kind` **and** the matching `origin.syntax` in one place, so the classification and the spelling cannot disagree — the invariant every `TypeRef` consumer relies on.

**`scalar`/`named` are placeless by construction.** No file wrote a presence flag, and claiming a location would make a fabricated one indistinguishable from a real one. That is the call `Flat::classify` already makes for a composed spelling and `ensure_entry` for adapter-authored cells (pinned by `an_adapter_authored_type_cell_is_classified_but_placeless`). The layered ones keep the inner's location — a borrow exists *because of* that value.

## The risk I pinned rather than assumed

A composed `&T` must key **exactly** as `parse_quote!(&#t)` did, or `require_output` registers a different cell and resolution silently changes. `a_composed_type_keys_as_its_spelling` checks all five forms plus the placeless/inherited location split. Verified before writing the migration, not after.

## Two things that fell out

- **`synth_sum_leaves` takes `flat::Variant`** and zips its alternatives, so payload types are readings rather than syn fields — and the sum tag's readback now reads `TypeId` (*`TypeId` **is** the name*) instead of taking a path apart with `bare_path_ident`.
- **`opt()` is deleted** — the `parse_quote!(Option<#ty>)` helper whose whole problem was building a spelling with no classification beside it.

**The lookup #276 added** in `fn_plan` is deleted with its comment: the fold leaf carries its own reading, so there is nothing to fetch and nothing that can miss.

## Verification

- **Goldens byte-identical.** The bar for a refactor this size, and the number to judge it on.
- 546 lib tests; `cargo test --all --all-features` 14/14; clippy `--deny warnings`; fmt.
- covertest-kotlin **48/48** on the JVM.
- Both ledgers (boundary, spelling census) **unmoved**.

## What's left

`is_optional` goes 4 → 3. The remaining callers need the adapter-side reading *sources* — `SumSpec` still building from `syn::ItemEnum`, and callback args reconstructed from `TypeKey`s rather than read off `TypeKind::Callback`. That is #275's last part, and deleting the accessor is its acceptance test.

One composition site is left alone deliberately: `unfold.rs`'s `convert_out_ty` still composes `Option<#leaf_ty>` as a `syn::Type` field. Noting it so it is not mistaken for an oversight.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/sum_out.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/sum_out.rs
@@ -193,7 +193,14 @@ pub(crate) fn encode_sum_group(
             tag_leaf.out_ty.key()
         )
     };
-    let ident = syn::Ident::new(&id.name, proc_macro2::Span::call_site());
+    // Raw-aware: a sum may legitimately be named `r#type`, and `Ident::new`
+    // rejects that spelling.
+    let ident = id.ident().unwrap_or_else(|| {
+        panic!(
+            "jnigen sum unfold: selector type `{}` is not a single identifier",
+            id.name
+        )
+    });
     let module = ext.fn_module(registry, &ident);
     let source: syn::Path = syn::parse_quote!(#module::#ident);
 
--- a/prebindgen/src/api/lang/jnigen/jni/emit/sum_out.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/sum_out.rs
@@ -43,38 +43,45 @@ pub(crate) const SUM_TAG_LEAF: &str = "tag";
 pub(crate) fn synth_sum_leaves(
     ext: &Declarations,
     sum_cfg: &SumConfig,
-    item_enum: &syn::ItemEnum,
+    sum: &crate::api::core::flat::Variant,
 ) -> Vec<crate::api::core::unfold::UnfoldLeaf> {
     use crate::api::core::{
         types_util::SumSpec,
         unfold::{LeafSource, UnfoldLeaf},
     };
 
-    let spec = SumSpec::from_item_enum(item_enum);
+    // `SumSpec` still reads the item — it owns the leaf-NAMING convention, which
+    // is jnigen's own. The payload TYPES come from the element beside it, whose
+    // fields are already readings, so nothing here has to compose or look one up.
+    let spec = SumSpec::from_item_enum(&sum.origin.syntax);
     // The selector rides ahead of the groups it chooses between, and carries
     // **which sum** it selects over as its `out_ty` — that is how the emitter
     // finds the enum to `match` when the sum is a field rather than the whole
     // returned value. Nothing looks up a converter for it (`has_converter()` is
     // false for a `SumTag`): there is no value to convert, the emitter assigns
     // the tag literal per arm. Its wire is a `jint` by definition.
-    let enum_ident = &item_enum.ident;
+    let enum_ident = &sum.name;
     let mut leaves = vec![UnfoldLeaf {
         name: SUM_TAG_LEAF.to_string(),
         path: Vec::new(),
-        out_ty: syn::parse_quote!(#enum_ident),
+        // Composed: the tag names WHICH sum it selects over, and no source
+        // wrote that as a standalone type. Nothing resolves a converter for it
+        // (`has_converter()` is false), but the emitter reads it back to find
+        // the enum to `match`.
+        out_ty: crate::api::core::flat::TypeRef::named(enum_ident),
         identity: false,
         nullable: false,
         source: LeafSource::SumTag,
         group: None,
     }];
-    for variant in &spec.variants {
+    for (variant, alt) in spec.variants.iter().zip(&sum.alternatives) {
         let kotlin_name = ext.sum_variant_class_name(sum_cfg, &variant.ident);
-        for field in &variant.fields {
+        for (field, alt_field) in variant.fields.iter().zip(&alt.fields) {
             let prop = sum_field_prop_name(field);
             leaves.push(UnfoldLeaf {
                 name: sum_slot_fragment(&kotlin_name, &prop),
                 path: Vec::new(),
-                out_ty: field.ty.clone(),
+                out_ty: alt_field.ty.clone(),
                 identity: false,
                 nullable: false,
                 source: LeafSource::VariantField {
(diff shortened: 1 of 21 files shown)


prebindgen/src/api/lang/jnigen/jni/emit/sum_out.rs
```diff
@@ -178,12 +185,15 @@ pub(crate) fn encode_sum_group(
...
+    let crate::api::core::flat::TypeKind::Named { id } = &tag_leaf.out_ty.kind else {
         panic!(
-            "jnigen sum unfold: selector type `{}` is not a path type",
-            TypeKey::from_type(&tag_leaf.out_ty)
+            "jnigen sum unfold: selector type `{}` is not a named type",
+            tag_leaf.out_ty.key()
         )
-    });
+    };
+    let ident = syn::Ident::new(&id.name, proc_macro2::Span::call_site());
```
milyin: **[P2] Preserve raw identifiers when rebuilding the sum name**

`TypeId::name` comes from `Ident::to_string()`, so a valid declaration such as `enum r#type { Value(u8) }` stores `"r#type"` here. `proc_macro2::Ident::new("r#type", ..)` panics (`"r#type" is not a valid Ident`), which makes JNI generation abort when this sum is encoded. The previous `bare_path_ident(&tag_leaf.out_ty)` path preserved the raw `Ident`. Please reconstruct with a raw-aware parser (for example `syn::parse_str::<Ident>(&id.name)`) or retain the original identifier syntax, and add a raw-name regression test.

milyin: Confirmed and fixed in 2dd360d. Reproduced first:

```
to_string()             = "r#type"
Ident::new("r#type")    = PANICKED   ("r#type" is not a valid Ident)
parse_str::<Ident>      = ok, round-trips to r#type
```

So it panics rather than errors — generation aborts outright, and only when a source happens to use a keyword.

**The recovery went on `TypeId`, not the call site.** `TypeId::ident()` parses instead of constructing, and returns `None` for a name that is not a single identifier — the same answer `bare_path_ident` gave for a path-qualified one. One place to be right, because a caller getting this wrong gets no signal until that keyword shows up, and #275's last part will want the same recovery.

**Writing the end-to-end test rather than only the unit one paid for itself.** `a_raw_named_sum_generates` drives a raw-named sum through `resolve` + `write_rust` — and it immediately found a **second** site:

```
panicked at jni/trait_impl.rs:351: "r#type" is not a valid Ident
```

That is `emitted_source_type_names`, which I had assessed as *"generated names, never raw"* when scanning for other instances. I was wrong — `rust_short_name_opt` returns whatever the source wrote.

It is **pre-existing**, verified against the base commit. I fixed it here rather than filing it, because a test asserting "a raw-named sum generates" while a known panic sat on the same path would be worse than no test at all.

Two tests: `a_raw_identifier_survives_typeid` for the round trip and the not-an-identifier case, and the end-to-end one above.

Goldens still byte-identical; 548 lib tests, `--all --all-features` 14/14, clippy, fmt, covertest 48/48.

Worth noting for the record: my "other `Ident::new` sites are all synthetic" check was a grep plus a judgement call, and the judgement was wrong on one of six. The end-to-end test is what caught it — the unit test would have passed happily.

prebindgen/src/api/lang/jnigen/jni/tests/sealed.rs
```diff
@@ -1623,3 +1623,66 @@ fn slice_of_sum_callback_arg_is_rejected_with_its_reason() {
...
+    ];
+    let registry =
+        crate::api::test_util::reg_from_items(declare_referenced(items)).expect("index items");
+    let jni = JniGenBuilder::new()
+        .set_package_prefix("io.test.jni")
+        .package(
+            crate::package!()
+                .class(crate::ptr_class!(ZThing))
+                .class(crate::sealed_class!(r#type))
+                .fun(crate::fun!(z_emit)),
```
milyin, line 1687: **[P2] Exercise the output sum encoder in this regression**

This currently declares only `z_emit`; the test run reports `skipping undeclared #[prebindgen] fn z_read`. Consequently `g.rs` contains the raw name through the sealed-class **input** converter (`JObject_to_r_type`), but contains no `match` over `myflat::r#type`, so `encode_sum_group`—the site fixed from the review finding—is never called. The assertion would still pass if that call site regressed to `Ident::new`. Please also bind `z_read` (or another function returning `r#type`) and assert the generated output-side match/path, so the test actually fails at the original regression site.

milyin: Right, and the test proved nothing. Fixed in 5a6af81.

**Verified the accusation rather than assuming it.** With the `sum_out.rs` fix reverted, the old test still passed — so the raw name it was matching came from the sealed-class *input* converter, exactly as you said, and `encode_sum_group` never ran.

A regression test that survives its own regression is worse than none: it converts "unguarded" into "guarded" in the reader's head.

**The sum is now a value-form field**, delivered through `expand_return!(ZThing).fields(...)` plus a callback — the shape that puts it through the output encoder — and the assertion is on the output-side match path rather than on the raw name appearing anywhere in the file:

```rust
assert!(rust.contains("myflat::r#type::Missing") && rust.contains("myflat::r#type::Exact"));
```

**Proved both directions this time**, which is what I should have done originally:

```
fix reverted → panicked at emit/sum_out.rs:198  → test result: FAILED
fix restored →                                    test result: ok
```

Goldens byte-identical; 548 lib tests, `--all --all-features` 14/14, clippy, fmt.

Three review rounds on this branch have each caught something real, and two of the three were me asserting a guarantee I had not measured — the "other `Ident::new` sites are synthetic" judgement, and now this test's coverage. The habit worth keeping is the one this comment forced: revert the fix, watch the test fail, then restore.

PR #279, opened 2026-08-02
The adapter's reading sources, and the last accessor goes

milyin:
Closes #275 — the last of four PRs (#274, #276, #278).

## Why

`Conversions::is_optional` took a spelling and asked the registry for its reading. A type with no cell answered `false` — *"not optional"* — rather than saying it never entered the pipeline. That is #266's shape, and its symptom is #273's missing `?`.

Three callers remained, all at places where the adapter **obtains** a type rather than consumes one. **Deleting the accessor is this issue's acceptance test.**

## Sum payload

`write_sealed_classes` holds the `flat::Variant` beside the `SumSpec` — the split #278 drew in `synth_sum_leaves` — so `build_sealed_class` zips the alternatives in and `sum_payload_kt_type` takes the alternative's own `flat::Field`.

**`SumSpec` is untouched**, deliberately: it owns the leaf-*naming* convention, which is jnigen's own, and rebuilding it on the model would touch 8 production sites and 5 tests for no gain here.

**The existing tests caught a regression I introduced**: a fieldless enum declared `sealed_class!` used to panic with a diagnosis, and my first `else { continue }` made it a silent skip. It keeps the diagnosis — now sourced from the model's own `Type::Variant` / `Type::Enum` distinction rather than re-derived with `enum_shape`, which is unused here as a result.

## Callback args

`convert_crossing` destructured `TypeKind::Callback { args }` — **already `Vec<TypeRef>`** — and immediately discarded it with `.map(|a| a.origin.syntax.clone())`. That line is gone; `dispatch_fn_input` and `callback_input` take `&[TypeRef]`, and the borrow peel reads `borrow_target()` instead of matching `syn::Type::Reference`.

### The one judgement call

`callback_iface_spec` is reached **only** through the `SpecKey`-keyed memo, and `SpecKey` needs `Ord` — which a `TypeRef` cannot give, since its `Origin` carries a `SourceLocation` and two identical readings from different files would compare unequal. `derive_iface_spec` is also contractually a pure function of the key (`iface_spec` debug-asserts it).

So `derive_iface_spec` resolves the key's `TypeKey`s back to readings. That is a **lookup, not a classification** — `reading` has been a pure lookup since #267, and `convert_crossing` is the existing precedent.

**It changes behaviour, deliberately:** a missing reading now *defers* the spec (the "not yet derivable, retried" state `iface_spec` already documents) instead of answering "not optional" and rendering a non-null Kotlin param. The old answer was the #273 bug shape.

## The risk with no other signal

A `SpecKey` built from `TypeRef::origin.syntax` must equal one built from `extract_fn_trait_args`, or the memo silently emits **two** interfaces for one callback — not a panic, not an unresolved type, just a duplicate.

`a_callback_identity_is_the_same_from_the_reading_or_the_syntax` pins it. Written **before** the migration, and verified to fail when the two disagree.

## Stale docs, mine

Three references named `Conversions::{optional_inner, sequence_elem, is_optional_borrow}` — all deleted in #276. The worst was the spelling-census **panic message**, which handed a developer that list at the exact moment they tripped the guard. They now point at the `TypeRef` methods that exist.

## Verification

- **Goldens byte-identical** — which is also what says the `SpecKey` identity did not split.
- 549 lib tests; `cargo test --all --all-features` 14/14; clippy `--deny warnings`; fmt.
- covertest-kotlin **48/48** on the JVM — callbacks and sealed classes are both heavily exercised there, which is exactly this PR's blast radius.
- Boundary ledger **129 → 127**, `emit/callback.rs` off it entirely.
- Acceptance, as a command: `grep -rn "is_optional" prebindgen/src` returns only `PathStep::is_optional`, an unrelated inherent method.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/callback.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/callback.rs
@@ -21,14 +21,14 @@ use crate::api::core::registry::Conversions;
 /// returned), so they are converted to `__JniErr` and logged via `tracing`.
 pub(crate) fn callback_input(
     ext: &Declarations,
-    args: &[syn::Type],
+    args: &[crate::api::core::flat::TypeRef],
     registry: &impl Conversions<KotlinMeta>,
 ) -> Option<(syn::Type, syn::Expr)> {
     // Human-readable tag for attach/log messages.
     let name = format!(
         "Fn({})",
         args.iter()
-            .map(|t| TypeKey::from_type(t).to_string())
+            .map(|t| t.key().to_string())
             .collect::<Vec<_>>()
             .join(", ")
     );
@@ -46,7 +46,13 @@ pub(crate) fn callback_input(
     let arg_names: Vec<syn::Ident> = (0..args.len())
         .map(|i| format_ident!("__cb_arg{}", i))
         .collect();
-    let arg_pat_ty: Vec<TokenStream> = args.iter().map(|t| quote!(#t)).collect();
+    let arg_pat_ty: Vec<TokenStream> = args
+        .iter()
+        .map(|t| {
+            let t = &t.origin.syntax;
+            quote!(#t)
+        })
+        .collect();
 
     // Per-arg encode preludes binding the typed `run`'s args in declared
     // order (a decomposed arg contributes one arg per leaf). Each entry of
@@ -71,7 +77,7 @@ pub(crate) fn callback_input(
         // user callback's `run(List<T>)`. Reuses the OUTPUT fold's folder
         // interface + appender singleton, driven from the trampoline.
         if let Some(plan) = registry
-            .callback_arg_plan(&TypeKey::from_type(arg_ty))
+            .callback_arg_plan(&arg_ty.key())
             .filter(|p| super::render::is_iterable_fold(&p.shape))
         {
             // Every leaf converter must already be resolved (deferral safety).
@@ -162,7 +168,7 @@ pub(crate) fn callback_input(
 
         // Decomposed arg: deliver the leaves of its type-level canonical
         // output, exactly like a return delivery.
-        if let Some(plan) = registry.callback_arg_plan(&TypeKey::from_type(arg_ty)) {
+        if let Some(plan) = registry.callback_arg_plan(&arg_ty.key()) {
             // Deferral safety: every leaf converter (and identity-leaf
             // projection) must already be resolved — return None so the rank
             // resolver retries this converter later otherwise. A synthesized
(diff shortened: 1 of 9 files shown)


prebindgen/src/api/lang/jnigen/jni/emit/callback.rs
```diff
@@ -193,15 +199,18 @@ pub(crate) fn callback_input(
...
-                    let core = (*r.elem).clone();
-                    (quote!((#cb_arg).clone()), registry.output_entry(&core)?)
-                }
-                _ => return None,
-            },
+            // A borrow: the callback hands out a reference, and the value is
+            // cloned for the JVM. Off the reading — `borrow_target` is the
+            // model's answer, not a syn match.
+            None => {
+                let core = arg_ty.borrow_target()?;
```
milyin, line 227: **[P2] Keep the clone fallback limited to a directly spelled borrow**

`borrow_target()` also returns `Some(T)` for transparent-wrapper spellings whose model kind is `Ref`, e.g. a callback argument `Box<&ZThing>`. That spelling intentionally has no output converter (`a_wrapped_borrow_has_nothing_to_bridge_and_refuses` pins the same rule for ordinary crossings), so this new fallback is taken. But `(#cb_arg).clone()` then has type `Box<&ZThing>`, while the selected core converter accepts `ZThing`; the generated trampoline resolves but fails when the consumer compiles it. Previously the `syn::Type::Reference` match declined this case. Please retain a direct-reference/bridgeability guard for the value transformation (while still using `borrow_target()` for classification), and add a callback regression for `Box<&T>`.

milyin: Done, but the measurement changed the answer — reporting it rather than just complying.

**The guard you asked for is unreachable.** I added the direct-reference check, then tried to make it fail: with it removed, `Box<&String>` and `Box<&ZThing>` produce **byte-identical** refusals. The fallback does fire, but the function is rejected anyway.

**What actually holds the invariant** is one level out. A callback arg's whole-value output converter is a *required* type, and `output_wrapper_shape`'s borrowed-opaque arm matches `syn::Type::Reference` on `produced` structurally — `Box<&T>` is a `Type::Path`, so it never gets one. (Not #272's `is_canonical_spelling`, which I checked first; both of its call sites are input-side.)

**The bug you described is real**, and I reproduced it by neutralising that arm — passing the canonical `&t1` as `produced` instead of the crossing's `syntax`:

```rust
Box::new(move |__cb_arg0: Box<&myflat::ZThing>| {
    let __cb0_enc = ZThing_to_jlong_11822692(&mut env, __cb_arg0)?;   // fn takes &myflat::ZThing
```

So the regression test is in, and it is a real one: `a_wrapped_borrow_callback_arg_declines` asserts plain `&T` resolves and `Box<&T>` refuses, and it **fails on exactly that emitted clone** when the arm is made spelling-blind.

**I dropped the local guard.** `boundary_ledger` failed on it — `emit/callback.rs: - -> 1`, a new source-syntax classification site outside `core::flat`, which is what #229 exists to remove. Paying a ledger entry for a classification that provably never fires seemed the wrong trade against a test that pins the invariant where it actually lives. The comment at the site now records the measurement and names the arm that carries it. Happy to put it back if you read the defence-in-depth differently — it is three lines and one ledger line.

`borrow_target()` still does the classification, as you asked.

Verified: 550 lib tests · `--all --all-features` 14/14 · clippy `--deny warnings` clean · fmt (CLI config) · regen-check byte-identical after `cargo clean -p` · covertest-kotlin 48/48.

PR #285, opened 2026-08-02
The wrapper-shape handlers take the inner type as a reading

milyin:
First of #284's three steps.

## This is a prerequisite, not a win on its own

Stating that up front because the numbers say so:

| | before | after |
|---|---|---|
| `.syntax()` calls in the two files | 19 | **24** (up) |
| `boundary_ledger` | 127 | 127 (unchanged) |

**My own plan predicted the ledger would fall here. It does not**, and reading the sites says why: `selector.rs`'s 3 and `trait_impl.rs`'s 11 counted matches are `is_unsized_spelling`, `decoded_vec_satisfies`, and the `Type::Reference` bridgeability guards from #272/#279 — all genuine **spelling** questions, and the documented exemption. Nothing was owed there, so nothing moved.

## What it does change

`input_wrapper_shape` / `output_wrapper_shape` and their four sub-handlers took `t1: &syn::Type`, and `selector.rs` produced it by destructuring a reading it already held:

```rust
let t1 = inner.syntax().clone();
self.output_wrapper_shape(WrapperShape::Borrow { mutable }, syntax, &t1, registry)
```

…while `input_terminal` / `output_terminal` **right beside them** already took `reading: &TypeRef`. Now `t1` is a `&TypeRef` too, so:

- a handler cannot be reached with tokens that have no reading — enforced at every call site by the type system;
- #284's step 2 has something to pass when `input_entry` / `output_entry` start taking a reading. That is where the ~90 downstream lookups actually get fixed.

## Why `produced` stays a spelling

Deliberate, and the `&[T]` arm is what settles it. There the adapter **composes** `Vec<#elem>` — and #280 sealed minting to the model, so `api::lang` has no `Vec<T>` reading to make.

That turns out to be consistent rather than awkward: `produced` is *defined* as the tokens the converter yields, and every question asked of it (`is_canonical_spelling`, the `Type::Reference` guards) is a spelling question. So the split carries meaning — **`produced` = what is emitted, `t1` = what is wrapped.**

## Verification

- regen-check **byte-identical** after `git clean -fd examples/` + `cargo clean -p`, rebuild confirmed by the `Compiling` lines
- covertest-kotlin 48/48

## Next

#284 step 2: `conversion` / `input_entry` / `output_entry` take `&TypeRef`, `reading` becomes `reading(&TypeKey)`. That is the step that deletes the 42 `key.to_type()` round trips and turns this one's `+5 .syntax()` into a much larger net removal.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/selector.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/selector.rs
@@ -72,14 +72,13 @@ impl Declarations {
                         ..
                     }
                 );
-                let t1 = target.syntax().clone();
                 if let Some(mut c) = self.input_wrapper_shape(
                     WrapperShape::OptionRef { mutable },
                     syntax,
-                    &t1,
+                    target,
                     registry,
                 ) {
-                    c.subs = vec![t1];
+                    c.subs = vec![target.syntax().clone()];
                     return Some(c);
                 }
             }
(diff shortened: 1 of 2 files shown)


milyin (commented):
Reviewed at `31e774a`; no findings.

The selector now passes the exact inner `TypeRef` produced by the model into every wrapper-shape handler, while dependency `subs` and generated Rust continue to use that reading’s syntax. Keeping `produced` as a spelling is consistent with its bridgeability role and is necessary for the borrowed-slice arm’s adapter-composed `Vec<T>`. I found no mismatch across the borrow, optional-borrow, optional, sequence, or output-only handle/string paths.

The generated artifacts are reported byte-identical, and all four PR checks are green.

PR #286, opened 2026-08-02
Registry lookups take the reading, not a spelling

milyin:
Second of #284's three steps. **Stacked on #285** — targets `wrapper-shape-readings`, to be merged together.

| | |
|---|---|
| `reading(&TypeKey)` | the one keyed door |
| `conversion` / `input_entry` / `output_entry` | take `&TypeRef` |
| `reading_of(&syn::Type)` | the visible *"I only had tokens"* step |

## The guarantee

An entry lookup **cannot be called about a type the registry does not know**. #280 sealed minting, so a `TypeRef` can only come from the model or from `reading`/`reading_of` — and those answer `None` for an unregistered type, which the caller must handle. Before, any tokens could be passed and got a silent `None` back.

## The important find was not in the trait

`Registry` carried **inherent** `input_entry` / `output_entry` taking a `&syn::Type` (`scan.rs:579/585`), and an inherent method wins over a trait method on a concrete receiver. So every caller holding a `Registry` used the spelling door, and the trait's signature could not close it.

**Changing the trait alone left 104 sites silently compiling against the old path.** Closing the inherent pair is what surfaced them — and it is why this PR is larger than #284 predicted.

Same *"second door inside the room"* that hid `classify` behind `Registry::reading` until #267. Worth noting that a green build was not evidence here: the code compiled precisely *because* the old door was still open.

## #285's payoff lands here

Every `t1_ty` in the wrapper-shape handlers became `t1` — the handler already holds the reading. That is exactly what #285 was for, and it is why the two want to merge together.

## Honest numbers

| | before | after |
|---|---|---|
| `to_type()` (prod) | 45 | **37** |
| `boundary_ledger` | 127 | 127 |

The ledger is unchanged and **structurally so** — it counts syn *variant mentions*, and a signature change names none. The 37 remaining `to_type()` are spelling needs (C type names, `quote!` targets, diagnostics).

**Acceptance test, as a command** — none of them feeds a lookup:

```
grep -rn "to_type()" prebindgen/src | grep -v tests \
  | grep -E "input_entry|output_entry|conversion|reading"     # empty
```

## One deliberate non-shortcut

`reading_of` returns a **reading**, not an entry. A spelling-taking `entry_of` would have been far less churn at ~80 call sites — and would have restored exactly the door being removed. Keeping the two steps separate is what makes the `None` visible where a caller has only tokens.

## Verification

- regen-check **byte-identical** after `git clean -fd examples/` + `cargo clean -p`
- covertest-kotlin 48/48

## Next

#284 step 3 — the remaining callers that hold only a key go through `reading(&key)` first. Most of the `reading_of` sites introduced here are the honest inventory of where an adapter still peels or composes tokens; each is a candidate for carrying a reading instead.

### DIFF
--- a/prebindgen/src/api/core/registry/declare.rs
+++ b/prebindgen/src/api/core/registry/declare.rs
@@ -427,14 +427,18 @@ impl<M> RegistryBuilder<M> {
 ///
 /// [`conversion`]: Conversions::conversion
 impl<M> Conversions<M> for RegistryBuilder<M> {
-    fn reading(&self, ty: &syn::Type) -> Option<crate::api::core::flat::TypeRef> {
-        self.registry.reading(ty)
+    fn reading(&self, key: &TypeKey) -> Option<crate::api::core::flat::TypeRef> {
+        self.registry.reading(key)
     }
     fn flat(&self) -> &crate::api::core::flat::Flat {
         &self.registry.flat
     }
-    fn conversion(&self, dir: Direction, ty: &syn::Type) -> Option<&TypeEntry<M>> {
-        self.built.get(&(dir, TypeKey::from_type(ty)))
+    fn conversion(
+        &self,
+        dir: Direction,
+        reading: &crate::api::core::flat::TypeRef,
+    ) -> Option<&TypeEntry<M>> {
+        self.built.get(&(dir, reading.key()))
     }
     fn crossing_keys(&self, dir: Direction) -> Vec<TypeKey> {
         self.order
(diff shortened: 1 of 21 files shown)


milyin (commented):
Reviewed at `0fa7385`; no findings.

The PR closes both lookup paths: the `Conversions` trait and the concrete `Registry` inherent methods now require a registered `TypeRef`, while `reading(&TypeKey)` / `reading_of(&syn::Type)` make key- or spelling-only transitions explicit and fallible. The migrated call sites preserve input/output direction and previous miss behavior, and #285’s wrapper handlers now use their carried inner readings directly as intended.

The PR is correctly stacked on `wrapper-shape-readings`, generated artifacts are reported byte-identical, and all four checks are green.

PR #287, opened 2026-08-02
struct_plan peels off the model instead of the tokens

milyin:
First piece of #284 step 3. **Stacked on #286** (which is stacked on #285) — all three merge together.

Small on purpose. See the finding at the bottom for why the rest of step 3 is not what the issue assumed.

## Three round trips

`struct_plan.rs` already held `reading`, `optional_inner` and `bare_ref` as `TypeRef`s, then downgraded all three and looked them back up:

```rust
let effective_ty = reading.syntax().clone();
...
registry.reading_of(&effective_ty).and_then(|tr| registry.output_entry(&tr))?
```

Gone. **The file now has zero `reading_of` calls.**

## Two latent defects, not round trips

Two of the five sites were asking a *spelling* a question the model answers — the #270/#272 family:

| | |
|---|---|
| `pat_match_top(&slot_ty, "Vec")` | compares the **last path segment**, so `Box<Vec<T>>` answered **false**. Now `slot.sequence_elem().is_some()` |
| `bare_path_ident(&slot_ty)` | takes the spelling apart for a name, which answers about the **wrapper** for `Box<T>`. Now the name comes off `TypeKind::Named { id }` |

Neither is reachable from the in-tree examples — goldens byte-identical — so these are the "correct output, no signal" shape of #266/#273, not observed breakage. Reporting them as latent rather than claiming a fixed bug.

## Finding: the rest of step 3 is bigger than the issue assumed

`emit/flat_input.rs` holds **20 of the 34 `option_inner_type` callers and 12 `reading_of` sites**. I expected a peel substitution there. It is not:

**the file walks `syn::Fields::Named` directly**, while `flat::Struct::fields` already carries a `TypeRef` per field.

So it needs the **element-walking** change — take `&flat::Struct`, walk `struct.fields` — which is exactly the follow-up the umbrella already records from #283 for `scan_struct`/`scan_enum`. That is its own PR, not a rushed extension of this one, and #284's step-3 description should be updated to say so.

## Ledger

Unchanged at 127, and honestly so: `types_util` only falls when its callers stop needing it, and `option_inner_type` still has 34.

## Verification

- regen-check **byte-identical** after `git clean -fd examples/` + `cargo clean -p`
- covertest-kotlin 48/48

### DIFF
--- a/prebindgen/src/api/core/registry/declare.rs
+++ b/prebindgen/src/api/core/registry/declare.rs
@@ -427,14 +427,18 @@ impl<M> RegistryBuilder<M> {
 ///
 /// [`conversion`]: Conversions::conversion
 impl<M> Conversions<M> for RegistryBuilder<M> {
-    fn reading(&self, ty: &syn::Type) -> Option<crate::api::core::flat::TypeRef> {
-        self.registry.reading(ty)
+    fn reading(&self, key: &TypeKey) -> Option<crate::api::core::flat::TypeRef> {
+        self.registry.reading(key)
     }
     fn flat(&self) -> &crate::api::core::flat::Flat {
         &self.registry.flat
     }
-    fn conversion(&self, dir: Direction, ty: &syn::Type) -> Option<&TypeEntry<M>> {
-        self.built.get(&(dir, TypeKey::from_type(ty)))
+    fn conversion(
+        &self,
+        dir: Direction,
+        reading: &crate::api::core::flat::TypeRef,
+    ) -> Option<&TypeEntry<M>> {
+        self.built.get(&(dir, reading.key()))
     }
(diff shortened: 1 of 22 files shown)


milyin (commented):
Reviewed at `ef17240`; no findings.

`classify_field` now uses the readings it already owns for the field, optional inner, and descriptor slot, eliminating all three local `reading_of` round trips without changing lookup direction or deferral. The two classification fixes are also sound: `slot.sequence_elem()` recognizes transparent `Box<Vec<T>>`-style spellings, and `TypeKind::Named` supplies the modeled nominal name instead of the wrapper’s last path segment. Converter-wire descriptor precedence and optional-enum handling remain intact.

The PR is correctly stacked on `lookups-take-readings`, generated artifacts are reported byte-identical, and all four checks are green.

PR #293, opened 2026-08-02
core: the model says what it erased, and one rebuild stops guessing

milyin:
Item 1 of #292, plus the audit its comment asked for — which turned up one live miscompilation, fixed here.

## The model

`TypeKind` erases `Box` and `Cow`, and that erasure is right: `Box<Option<T>>` is one optional string to every destination language. But **conversion follows the syntax**, and the two facts a rebuild needs were not on the model. #290 shipped `erased_wrapper()` — the outermost wrapper of the top layer — which is enough to decide *whether* to refuse and not enough to rebuild.

* `TypeRef::erased_wrappers() -> Vec<&'static str>` — every wrapper, outermost first. `Box<Cow<'_, T>>` is two *different* operations, so a name and a count would not do.
* `TypeRef::stripped_syntax() -> syn::Type` — the spelling under them.

**Derived, never stored.** `lower_type` keeps discarding the wrapper into `kind`; these recompute from `origin.syntax` at the point of use, exactly as the shipped `erased_wrapper()` already does. A stored spelling is a second thing that can disagree with `syntax`, and a representation detail recorded back into the model is what the flat model exists to avoid.

`stripped_syntax` is defined by its **invariant**, not by its loop: *the spelling whose own lowering yields exactly this `kind`*. So the peel runs to a **fixed point** — `Box<Box<T>>` classifies as `T`, and one strip leaves a `Box<T>` that does not match. The test asserts the property (strip, lower again, compare classifications) over five wrapped rows plus an unwrapped control, and was seen to fail against a one-layer implementation before being trusted.

Both carry the #290 ordering finding on their docs, because **an erasure sits outside the layer it wraps**:

| Spelling | at the top | on `borrow_target()` |
|---|---|---|
| `Box<&Vec<T>>` | `["Box"]` | `[]` — `kind` is `Ref`; peeling first drops the wrapper |
| `&Box<Vec<T>>` | `[]` — a `Type::Reference` cannot be peeled | `["Box"]` |

Neither check subsumes the other and the two classify identically, so a walk must ask at every layer on the way down. That pair is a test.

## The audit

The #290 comment left this explicitly unchecked: the guards were in jnigen's **input** lowerings only; the output side and cbindgen were not looked at.

The population is two, and only one has to ask. A site that **classifies** must never consult the wrapper — that is the erasure working. A site that **binds a source value and destructures or rebuilds it** must.

| Site | Kind | Verdict |
|---|---|---|
| cbindgen — all 27 non-test peels (`opaque_ptr_payload_inner`, `box_inner`, `is_option`, `produces_array`) | classify → a C wire | correct as is; the `Box`-derived wire is #230's question, not this one |
| jnigen `selector` / `struct_plan` / `fn_plan` / `iface` / `builder` / `render` / `fold` / `kotlin_emit` | classify → a Kotlin type, a JNI wire, a plan shape | correct as is |
| jnigen `emit/flat_input` ×3, `emit/vec_build` ×3 | rebuild | guarded by #290 |
| `delivery.rs` field reaches | destructure | coerced via `bind_as_option` (#268) |
| wrapper.rs `is_convert` → `raw.map(..)` | destructure | **compiles** — method resolution auto-derefs, verified against `rustc` |
| **wrapper.rs → `emit_unfold_delivery`** | **destructure** | **unguarded — `E0308`** |

Builder delivery binds the returned value and matches it against `Option`'s patterns, and match ergonomics does not see through a `Box`. `read_through_erased_wrappers` undoes the wrappers at the **single point the value enters the delivery**, not at each of the four matches downstream; `Cow` panics with a message that names #292 item 3 as where it becomes possible, per the reserved-vs-refused convention.

## Evidence

#269 means 737 `contains(..)` assertions pass on Rust that does not compile, so the fixture is where an `E0308` is actually a failure: `perftest_flat::boxed_latest`, whose generated binding covertest `include!`s and CI builds. `Summary` carries an output expansion, so its return takes no converter to name the spelling for it — the shape `boxed_note_echo` (a converted return) does not reach.

- **Verified by disabling the fix**: the build breaks with `expected Box<Option<Summary>>, found Option<_>` at `generated_bindings.rs:12768`.
- **Round-tripped on the JVM**: new `transparent wrapper crossings` section in `Test.kt`, asserting the wrapped and unwrapped twins have the same surface and the same values. `PASS - 49 sections`.
- `cargo test` and `cargo test --all --all-features` green; clippy `--deny warnings` clean; fmt with the CI config clean.

## One golden line moved

`Box`'s `read` op drops its parens: `let v: Option<String> = (*(*v));` → `**v`. Every consumer splices into a `let` initializer. Converters happen to `#[allow(unused_parens)]` and wrapper externs do not, which is why the lint surfaced only at the new site — fixing it at the op's one definition rather than special-casing the caller is the right altitude. `value_form.rs` pinned the old spelling and is updated with the reason.

Regen drift is confined to covertest's three artifacts (the new fixture plus that line); `example_flat_aarch64.{rs,h}` did **not** move.

## Ledger

Unchanged — no watched `syn` variant site is added or removed. `erased_wrappers`/`stripped_syntax` live in `core::flat`, where classification belongs, and `read_through_erased_wrappers` consults the model rather than matching syntax.

Registered on #229: a row in the *Why the syntax rides along* table, a `What L4 taught` subsection, and an L5 bullet.

Items 2 (#230, cbindgen's wire from `kind`) and 3 (jnigen rebuilds instead of refusing) stay on their issues and consume this.

### DIFF
--- a/docs/language-integration.md
+++ b/docs/language-integration.md
@@ -65,6 +65,7 @@ classification stays small and genuinely neutral:
 | `Foo<'a, T>` | `TypeRef::origin.syntax` | generated Rust only |
 | "it is a `Foo`" | `TypeKind::Named` | every adapter |
 | `[u8; TAG_LEN]` — spelling / number / const identity | `TypeRef::origin.syntax` / `ArrayExtent::value` / `ExtentSource::Const` | C header / Kotlin / both |
+| the `Box` in `Box<Option<T>>`, and the `Option<T>` under it | `TypeRef::erased_wrappers()` / `stripped_syntax()` — derived from the syntax, not stored | an emitter that **rebuilds or destructures** a Rust value |
 | where an item came from | `Origin::location` — **absent for a synthesized one** | diagnostics |
 
 ### The rule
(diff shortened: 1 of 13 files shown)


milyin (commented):
Reviewed the full diff and found no blocking issues.

The model/emitter split is coherent here: `erased_wrappers()` and `stripped_syntax()` derive from the original spelling without changing `TypeKind`, and builder delivery removes wrappers from the value actually being delivered (the `Ok` side after domain-error routing, when applicable). Applying the wrapper reads outermost-first also handles nested `Box` layers correctly, while the explicit `Cow` refusal avoids emitting an invalid move through `Deref`. The generated fixture exercises the previously uncompiled `Box<Option<Summary>>` path at the Rust/JNI boundary and on the JVM.

Verification: focused `erased_wrapper` and `value_form` tests pass locally (48 tests total), and all PR checks are green (`build` on stable/1.85, `covertest`, and `smoke-asan`).

PR #294, opened 2026-08-02
jnigen: rebuild the spelling instead of refusing it (#292 item 3, closes #289)

milyin:
Item 3 of #292, and #289 with it. The two travel together because **#289 alone breaks the build**: reading a field's layer off the model is what makes the emitter rebuild an `Option` for a slot ascribed `Box<Option<_>>`. Either both, or neither.

## The rebuild

`build_through_erased_wrappers` is the input dual of #293's `read_through_erased_wrappers` — same `WRAPPER_OPS` rows, applied innermost-out. The three declared sites descend instead of refusing, collecting each layer's wrappers on the way down, because an erasure sits outside the layer it wraps.

**A layer's wrappers are applied only where that layer exists.** Applying them unconditionally double-wraps when two layers are the same reading (`arg == under_borrow` with no borrow; `under_borrow == core` with no `Option`). I got that wrong twice; both times a fixture caught it, and neither was reachable by any text assertion.

**`Cow` keeps `build: None`, and that is policy rather than impossibility.** `Cow::Owned(v)` is well-typed — an input rebuild owns its value. It is refused because always-`Owned` pays a copy per call and removes the borrow path the source asked for, and the callee can see the difference (`matches!(c, Cow::Borrowed(_))`). The row says so, and says it is one line to change on purpose. The read/build asymmetry is now two different reasons, not one flag.

## Two findings the fixtures forced out

Neither was in the plan, and neither is visible to the test suite.

**A wrapper silently cost a parameter its lowering.** The data-class declaration was looked up by `inner.key()` — the *wrapped* spelling — so `Box<Payload>` found no `Payload` declaration and fell to the general converter. No error, no diff, just a lost lowering. A declaration says what a type **is**, so it is keyed by the new `TypeRef::stripped_key()`. `key` stays for conversions, which genuinely differ per spelling.

**A wrapper over a terminal had no converter at all.** The layer arms bridge a wrapper as part of handling their layer; nothing covered `Box<Payload>`, which is `Named`. `input_transparent_bridge` delegates to the stripped spelling's converter and re-wraps — tried **last**, so no existing route changes.

## Refusals, each with a stated reason

Per the reserved-vs-definitive rule these say *why*, not just *no*:

| Shape | Why |
|---|---|
| `Box<&T>` | a converter produces an owned value; there is nothing for the borrow to point at that outlives the call (`E0106`) |
| `&Box<Vec<T>>` | interposing a `Box` between the caller's Vec and the callee needs a per-call `clone()` and an unguaranteed `T: Clone` |
| `Vec<Box<T>>` element | the helper trio is named for the element's Kotlin class, so two spellings of one class collide on `payloadVecNew` |

The first two were checked against `rustc` rather than reasoned about — which is how I learned the borrowed run only compiles *with* a clone.

## #289

`build_flat_struct_node` takes `&flat::Struct` and peels its fields off the model. `flat_probe_inner` is gone — its last caller asks `is_kotlin_enum_reading`, whose `enum_probe` peels the same layers, so `Box<Priority>` probes as `Priority` where it answered about the wrapper.

**Both censuses move down**, which no PR in the #284 chain managed, because this retires callers rather than re-typing signatures — exactly #289's prediction:

- spelling helpers in `flat_input.rs`: **18 → 9**
- boundary ledger: `flat_input.rs` **7 → 6** (127 → **126**)

Still on `syn::Fields`: `struct_input_body` / `sum_input_body`, the `.jobject_input()` decoders — separate walk, same rule, and the remaining 9.

## Evidence

Seven fixtures in `perftest-flat`, whose binding covertest `include!`s and CI builds — the only place an `E0308` is a test failure (#269).

**Every wrap was verified by disabling it** and reading the error naming its shape:

| Disabled | Result |
|---|---|
| `wrap_core` | `expected Box<Payload>, found Payload` |
| `wrap_optional` | `expected Box<Option<Payload>>, found Option<Payload>` |
| option-scalar wrap | `expected Box<Option<Priority>>, found Option<Priority>` |
| vec run wrap | `expected Box<Vec<Payload>>, found Vec<Payload>` |
| field wrap | `expected Box<Option<i64>>, found Option<i64>` |
| `input_transparent_bridge` | resolution fails, naming the three types |
| `stripped_key` | **no error at all** — the lowering just vanishes |
| optional-node gate (review) | `holderTagOr(null, ..)` fails at runtime: "Operation on a closed native handle." |

That last row is the one worth keeping: it is the failure mode a compile check cannot see either, and the reason the fix is a keying change rather than a guard.

`Holder { tag, summary: Summary }` is the first declared data class with a **required handle** field, added for the review fix: taken as `Option<Holder>`, the absent case carries a placeholder that cannot be decoded, so it is the only shape that can tell the two decode orders apart. Every pre-existing fixture decodes its placeholders successfully, which is why CI stayed green over the bug.

JVM round trip in `Test.kt` (`PASS - 49 sections`), asserting the wrapped and unwrapped twins have identical surfaces and identical values. `WrappedFields` pins the #289 change directly: `boxed: Box<Option<i64>>` and `plain: Option<i64>` both cross as `Long?` on the decoupled pair, and only the first gets a `Box::new`.

`cargo test` and `--all --all-features` green; clippy `--deny warnings` clean (`large_enum_variant` fired mid-way — the fix was to carry wrapper *lists* rather than whole readings, which is smaller and says more precisely what a rebuild uses); fmt with the CI config clean.

**Regen after a forced clean**: against the merge base the generated bindings have **zero genuinely-removed lines** — the diff is insertions plus moves from sort-order shifts. `example_flat_aarch64.{rs,h}` did not drift. Per #289, an existing golden that moved would have been a found defect.

> **Correction.** This originally read "303 insertions, 0 deletions". That figure was the diff against the *previous commit on this branch*, not the merge base, so it reported only the second commit's delta. Measured properly the first push had **55 changed lines of existing output**, every one an optional-node gate — the regression the review caught. Fixed in `eb9df58`; the statement above is now true and was not before.

## Remaining

#230 (item 2, cbindgen's wire from `kind`) is now the only open item on #292.

### DIFF
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -522,6 +522,19 @@ fn main() {
                 // with the wrong number of dereferences fails the build (#270).
                 .fun(fun!(boxed_note_echo))
                 .fun(fun!(plain_note_echo))
+                // Transparent wrappers on the INPUT side (#292 item 3), one per
+                // specialized lowering. These rebuild their parameter rather
+                // than decoding it, so the erased wrapper has to go back on
+                // before the value reaches the signature — and each layer is
+                // applied at a different point in the construction, which is why
+                // one shape cannot cover them all. Compiling this crate is the
+                // check: a missing `Box::new` is an `E0308`, invisible to any
+                // text assertion.
+                .fun(fun!(boxed_payload_id))
+                .fun(fun!(boxed_opt_payload_id))
+                .fun(fun!(boxed_opt_priority_weight))
+                .fun(fun!(boxed_elem_id_sum))
+                .fun(fun!(boxed_run_id_sum))
                 // The same wrapper over a DECOMPOSED return (#292). `Summary`
                 // has an output expansion, so this return takes no converter to
                 // name the spelling for it — the extern binds the value and
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -206,6 +206,11 @@ fn main() {
                     .method(fun!(payload_label_len)),
             ),
         )
+        // `Option<Holder>` where `Holder` has a REQUIRED handle field: the
+        // absent case passes pointer 0 for it, so the field decodes must stay
+        // inside the presence gate or `null` becomes a binding error instead of
+        // `None` (PR#294 review).
+        .package(package!().class(data_class!(Holder)))
         // A data class whose FIELDS carry transparent wrappers (#289 + #292):
         // `boxed: Box<Option<i64>>` must cross exactly as `plain: Option<i64>`
         // does — the decoupled `(present, value)` pair — with the `Box` put back
@@ -537,6 +542,7 @@ fn main() {
                 // check: a missing `Box::new` is an `E0308`, invisible to any
                 // text assertion.
                 .fun(fun!(wrapped_fields_sum))
+                .fun(fun!(holder_tag_or))
                 .fun(fun!(boxed_payload_id))
                 .fun(fun!(boxed_opt_payload_id))
                 .fun(fun!(boxed_opt_priority_weight))
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -206,6 +206,12 @@ fn main() {
                     .method(fun!(payload_label_len)),
             ),
         )
+        // A data class whose FIELDS carry transparent wrappers (#289 + #292):
+        // `boxed: Box<Option<i64>>` must cross exactly as `plain: Option<i64>`
+        // does — the decoupled `(present, value)` pair — with the `Box` put back
+        // on the Rust side. Peeling the field by path segment answered "not
+        // optional" and boxed it instead.
+        .package(package!().class(data_class!(WrappedFields)))
         // ── Subpackage `model`: enum + value class + nested data class ──────
         .package(
             package!("model")
@@ -530,6 +536,7 @@ fn main() {
                 // one shape cannot cover them all. Compiling this crate is the
                 // check: a missing `Box::new` is an `E0308`, invisible to any
                 // text assertion.
+                .fun(fun!(wrapped_fields_sum))
                 .fun(fun!(boxed_payload_id))
                 .fun(fun!(boxed_opt_payload_id))
                 .fun(fun!(boxed_opt_priority_weight))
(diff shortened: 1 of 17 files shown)


milyin (commented):
Found one blocking issue.

[P1] Keep optional-node field decodes inside the parent presence gate — `prebindgen/src/api/lang/jnigen/jni/emit/flat_input.rs:1971`.

In the new optional branch, `#decodes` runs before `#gate`, so every field is decoded even when the parent `*_present` flag is false. Those leaf values are deliberately inert placeholders when the Kotlin object is null and must not be interpreted. For example, an `Option<DataClass>` whose data class has a required handle field passes pointer `0` for the absent object; the unconditional direct-handle decode then calls `signal_binding_error` and returns instead of delivering `None`. An enum without discriminant 0 or a fallible custom converter can fail on its placeholder similarly. The base implementation kept `#decodes` inside the true arm.

Please build the conditional as `if present { #decodes; Some(#built) } else { None }` first, then apply `outer(...)` around that entire expression. The new wrapped fixtures use fields whose defaults all decode successfully, so their null round trips do not expose this regression.

All four CI checks are green, but they do not cover this invalid-placeholder case.

milyin (commented):
Re-reviewed the updated head (`eb9df58`). The previous blocking finding is resolved.

`render_flat_struct_node` now keeps `#decodes` inside the `present != 0` arm and applies the transparent wrapper around the complete optional expression, preserving both the original lazy decode behavior and the new spelling rebuild. The added `Option<Holder>` fixture is a strong regression test: the absent case supplies pointer `0` for a required handle leaf and verifies that it reaches the fallback rather than the binding-error channel; the present case verifies normal ownership/decode behavior. The regenerated Rust confirms the handle check and `Box::from_raw` are confined to the present arm.

No new findings. All checks pass on the updated head: Rust 1.85, stable, Kotlin/JNI covertest (including regen), and ASan smoke.

milyin (commented):
Reviewed the current head (`eb9df58`) again. There are no changes since my previous rereview and no outstanding findings.

The optional-node fix remains correct: inactive leaf placeholders are not decoded, transparent wrappers are rebuilt around the full optional expression, and the handle-bearing `Option<Holder>` covertest covers both absent and present paths. All four checks are green (Rust 1.85, stable, Kotlin/JNI covertest with regen verification, and ASan smoke).

Issue #291, opened 2026-08-02
Make `TypeKey` blind: a key should not be a back-door to `syn::Type`


milyin:
`TypeKey` serves two purposes, and only one of them is its job.

It is the identity a table is indexed by — and it is also a back-door to
`syn::Type`, via `to_type()` over the `ty: Rc<syn::Type>` it keeps beside the
canonical string. That second channel is wrong: **the only way to reach a
type's syntax should be `TypeRef::syntax()`**, because a reading is the thing
that pairs a spelling with the classification that vouches for it. A key that
hands out tokens lets a caller spell a type the model never classified, which
is the #211 boundary from the other direction — not "who decides what syntax
means", but "who is allowed to produce syntax at all".

`TypeKey::key()`'s own doc already concedes the narrower point: "The canonical
spelling is what a key *is* (#113), and reading it is legitimate — but it should
be the model's answer rather than every caller reaching into `syntax` for it."
`to_type()` is the reach it warns about, granted by the key itself.

## What it costs

40 `to_type()` call sites, but they are not 40 problems. Classified by what the
site actually wants:

| Category | Sites | What blinding costs |
|---|---|---|
| **A** — `key.to_type()` fed straight to `reading_of` / `immediate_edges` / `intern` | 12 | **Negative.** These are `key → syntax → key → reading` round trips; they get shorter. |
| **B** — declaration-owned types (`ptr_class!(Foo)`, `convert!(D)`) | 11 | Mechanical: the decl should keep the `syn::Type` the build script wrote. |
| **C1** — wants a *name*, not syntax (`bare_path_ident(&key.to_type())`) | 9 | Small: add `TypeKey::ident()`. |
| **C2** — genuinely needs syntax from a key alone | 8 (3 are test fixtures) | Real work — **5 production sites.** |

So: ~5 sites that need thought, ~20 mechanical edits, minus 12 that get simpler.

### Category A is the argument for doing it

`immediate_edges(dir, ty: &syn::Type)` (`registry/scan.rs`) opens by computing
`TypeKey::from_type(ty)` to look the reading up. So `resolve.rs`, `order.rs` and
five `cbindgen/trait_impl.rs` sites spell a key into syntax *purely so the
callee can re-key it*. The most common use of `to_type()` today is undoing
itself, at the cost of a normalize pass and a token render per call.

### Category B is the design point

Those keys came from a build script writing `ptr_class!(Foo)`: a real
[...]

PR #298, opened 2026-08-02
core: the edge walk takes the key it was re-deriving anyway (#291 stage A)

milyin:
First of four steps on #291, following that issue's own sequencing: category A alone, which shrinks code and proves the `reading(key)` route **before** anything is removed.

## What

`Registry::immediate_edges` asked for a `&syn::Type` and opened by re-keying it — twice, once for the structural children and once for the declared fields. Its `&syn::Type` parameter was used for nothing else. So `resolve.rs`, `order.rs` and `register_type_inner` each spelled a key into tokens purely so the callee could undo that.

```rust
pub(crate) fn immediate_edges(
    &self,
    dir: Direction,
    key: &TypeKey,
) -> Vec<(Direction, TypeRef)>
```

The same round trip, one layer out, ran wherever `key.to_type()` was fed straight to `reading_of` — which is literally `reading(&TypeKey::from_type(..))`. Those become `reading(key)`, the route #284 already moved jnigen's `convert_crossing` to; any spelling they still need comes off the reading.

44 `to_type()` call sites → 31. `TypeKey` itself is untouched — nothing is deleted until stage D.

## Two sites deliberately keep a spelling

Both now read it from the cell the registry already holds, so neither re-derives anything from a key:

- `order.rs`'s `plan_edges` needs real tokens for `extract_fn_trait_args`. Every node `visit_crossing` reaches has a cell (roots are the table's own keys; edges are filtered by `contains_key`), so the reading is always there.
- cbindgen's `convert_crossing` selector chain still takes `&syn::Type`. Moving it onto readings the way jnigen's went is its own change.

## Behaviour

`resolve.rs`'s `enqueue_edges_from` is deliberately called on keys with **no** cell. Keying rather than reading preserves that exactly — `immediate_edges` returns `Vec::new()` on a missing cell either way. This is why stage A keys the walk instead of switching those two sites to `reading()`.

## Verified

- `cargo test --all --all-features` — 631 lib tests, all green
- `examples/regen-check.sh` **byte-identical**, after `cargo clean --release` on every example crate so the rebuild was real

`regen-check.sh --with-zenoh-flat-jni` could not run: that sibling checkout is pinned to an older public API (`JniGen::new` no longer exists, plus argument-count drift) and does not build against `language-integration`. Pre-existing and unrelated — this PR changes no public surface.

### DIFF
--- a/prebindgen/src/api/core/registry/order.rs
+++ b/prebindgen/src/api/core/registry/order.rs
@@ -60,15 +60,23 @@ impl<M> Registry<M> {
             return;
         }
         let (dir, key) = node.clone();
-        let ty = key.to_type();
+        // Every node this walk reaches has a cell: the roots are the table's own
+        // keys, and each edge below is filtered by `contains_key`. So the
+        // spelling `plan_edges` needs is the reading the registry already
+        // stored — not one re-derived from the key (#291).
+        let plan_edges = self
+            .type_table(dir)
+            .get(&key)
+            .map(|cell| self.plan_edges(dir, cell.subject.syntax()))
+            .unwrap_or_default();
         let mut edges: Vec<Crossing> = self
-            .immediate_edges(dir, &ty)
+            .immediate_edges(dir, &key)
             .into_iter()
             // The structural edges arrive as readings, so the key is the
             // model's own answer rather than one re-derived from a spelling.
             .map(|(d, t)| (d, t.key()))
             .chain(
-                self.plan_edges(dir, &ty)
+                plan_edges
                     .into_iter()
                     .map(|(d, t)| (d, TypeKey::from_type(&t))),
             )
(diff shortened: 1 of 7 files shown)


milyin (commented):
Reviewed the full diff and found no blocking issues.

The stage-A boundary looks sound: `immediate_edges` now takes the identity it was immediately reconstructing, and still returns no edges for a missing cell, preserving the unresolved-diagnostic path. `visit_crossing` obtains callback-plan syntax from its direction-specific cell; its root/filtered-edge invariants ensure that cell exists. The cbindgen/JniGen migrations use `reading(key)` only where the prior `key.to_type() -> reading_of(...)` round trip already selected the registry reading, then recover syntax from that stored reading for the remaining syntax-based selectors/emitters.

Verification: the registry-focused suite passes locally (40 tests), `git diff --check` is clean, and all CI checks are green (`build` on 1.85/stable, Kotlin/JNI covertest, and ASan smoke).

PR #299, opened 2026-08-02
A declaration keeps the type it was written with (#291 stage B)

milyin:
Second of four on #291. **Stacked on #298** — review that one first; this PR's base will be retargeted to `language-integration` once it merges.

## What

`ptr_class!(Foo)` receives a real `syn::Type`, reduces it to a `TypeKey`, and throws it away. Everything that later needed those tokens back asked the **key** to reproduce them. That is backwards — the declaration is where the type came from.

Declarations now carry `Origin<syn::Type>` (the model's own convention for a node's tokens) at `SourceLocation::default()`, the sanctioned placeless location for something a build script authored rather than a captured file.

```rust
pub fn export_type(mut self, ty: Origin<syn::Type>) -> Self   // was: (key: TypeKey)
pub(crate) types:    HashMap<TypeKey, Origin<syn::Type>>      // Declared
pub replaces:        HashMap<TypeKey, Origin<syn::Type>>      // Decompositions
```

`export_type` now matches its sibling `cross`, which already took the spelling.

## The two sites a key genuinely could not serve

Both in `registry/scan.rs`, and both are why the core boundary had to move:

- the **qualified-declared-types diagnostic** needs multi-segment path *structure* for a type with no cell, so neither `reading()` nor a name accessor could serve it;
- the **declared-type scan** needs real tokens for `intern` — of a type that is in no table yet, so `reading()` has nothing to answer with. `intern` is `pub(in crate::api::core)` and deliberately takes tokens; that stays.

Both **canonicalize explicitly** at the point of use. That is not incidental: it is what the key was silently providing (`crate::Foo` must not read as a qualified path), and doing it in the open is the difference between preserving behaviour and hoping. The comments say so.

## Declare-phase consumers read their decl, not the registry

`build_expansions`, `build_deconstructors`, `convert_input_body`/`convert_output_body`, `build_sum_decons` and `validate_split_declarations` all run while only a `RegistryBuilder` exists, where `reading()` legitimately answers `None` for a not-yet-interned type. Swapping those to `reading()` would have been a **silent semantic change**, not a refactor.

## Two sites go the other way

Past the declare phase, where the sibling arm beside each already used `reading()`:

- `SpecKey::WholeFolder` in `derive_iface_spec` — contractually a pure function of its key, so a side channel was not open to it; it now defers on `None` exactly as the `Callback` arm above it does.
- cbindgen's callback structs read the argument types their declaration recorded, instead of rebuilding them from a `Vec<TypeKey>`. `CallbackKey` stays a list of identities — that is what the map is keyed by.

28 `to_type()` call sites → 9. Every one that remains is a name lookup (stage C1) or the idempotence test (stage D).

## Verified

- `cargo test --all --all-features` — 631 lib tests, all green
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release` on every example crate. This was the real question of this PR — whether a declaration's as-written spelling matches what the key was reproducing — and it does.

### DIFF
--- a/prebindgen/src/api/core/registry/declare.rs
+++ b/prebindgen/src/api/core/registry/declare.rs
@@ -139,8 +139,18 @@ impl<M> RegistryBuilder<M> {
 
     /// A type this binding **exports**: it crosses in both directions, and its
     /// body — a struct's fields, an enum's payloads — is scanned too.
-    pub fn export_type(mut self, key: TypeKey) -> Self {
-        self.registry.declared.types.insert(key);
+    ///
+    /// Takes the **type the declaration was written with**, like its sibling
+    /// [`Self::cross`], and derives the key here. It used to take the key alone,
+    /// which meant the scan had to recover tokens *from* the key to intern the
+    /// type and to diagnose its spelling — reasoning backwards from an identity
+    /// to a thing that already existed. A build script wrote `ptr_class!(Foo)`;
+    /// this is that `Foo` (#291).
+    pub fn export_type(mut self, ty: Origin<syn::Type>) -> Self {
+        self.registry
+            .declared
+            .types
+            .insert(TypeKey::from_type(&ty.syntax), ty);
         self
     }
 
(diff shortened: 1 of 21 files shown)


milyin (commented):
Reviewed the full delta against `edges-take-a-key`, including the declaration storage changes in core, the declare-phase JNI consumers, cbindgen callback/type emission, and the two post-declare registry lookups.

No blocking findings. The distinction is applied consistently: declaration-owned syntax is retained where the registry may not have interned a type yet, while scan explicitly canonicalizes at the two sites that previously inherited normalization from `TypeKey::to_type()`. The `WholeFolder` lookup also now defers cleanly when no reading exists.

Verification:
- `cargo test -p prebindgen --lib`: 558 passed
- `git diff --check`: clean

PR #300, opened 2026-08-02
A key answers what it is called (#291 stage C1)

milyin:
Third of four on #291. **Stacked on #299 → #298** — both green; bases get retargeted as each merges.

## What

Eight sites asked `to_type()` for a whole `syn::Type` and then threw all of it away but one ident. They were not asking for syntax; they were asking the key what it is *called*.

```rust
TypeKey::ident()      -> Option<syn::Ident>   // bare_path_ident's rule
TypeKey::short_name() -> Option<String>       // last segment, generics and all
```

**Two accessors, because the incumbent walks genuinely differ.** `bare_path_ident` refuses a type carrying generic arguments; the Kotlin class-name derivation reads `Publisher<'static>` as `Publisher` — a declaration writes the latter and means the class. One accessor would have had to pick a winner and silently change one set of call sites.

## Read off the canonical string, deliberately

`canon` is a token-stream rendering, so its tokens are space-separated — `Vec < u8 >`, `& Foo`, `a :: Foo`, `[u8 ; 4]` — which puts a path's head before the first `<` and its last segment after the last `::`, with `syn::parse_str::<syn::Ident>` as the total validator on the far end.

Reparsing the whole type instead would make a **name** depend on a serialize-then-reparse round trip, which is precisely the dependency #95 removed. Generic arguments are stripped *first*, because in `Vec<a::B>` the `::` belongs to the argument and the segment being named is still `Vec`.

## The warrant

`key_name_accessors_match_the_syn_walks` asserts both accessors equal to the walk each replaces across sixteen shapes: bare and qualified paths, generics, lifetime generics, references, mutable references, slices, arrays, tuples, unit, raw pointers, trait objects, fn pointers.

It also pins the one documented limit: a qualified-self path (`<T as Tr>::Item`) answers `None` rather than `Item`. It cannot reach here — `scan_declared_items` refuses a `qself` declaration — and refusing beats guessing for a shape this cannot read.

## Independent signal

The boundary ledger moved **120 → 117**, and `api/lang/jnigen/jni/kotlin_emit.rs` left it entirely. These were real source-syntax classification sites (`if let syn::Type::Path(tp) = &ty { … segments.last() … }`), not just call-site noise — which is the ledger's whole job to notice.

Three `to_type()` sites remain, all stage D: `option_depth`, `KotlinMeta::value_rust_key`, and the idempotence assertion that dies with the method.

## Verified

- `cargo test --all --all-features` — 634 lib tests, all green
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release` on every example crate

### DIFF
--- a/prebindgen/src/api/core/diagnostics.rs
+++ b/prebindgen/src/api/core/diagnostics.rs
@@ -15,7 +15,6 @@ use crate::api::core::{
     flat::{type_from_ident, Flat},
     prebindgen::NamePredicate,
     registry::TypeKey,
-    types_util::bare_path_ident,
 };
 
 /// What a binding claimed, so everything else can be reported.
@@ -79,7 +78,7 @@ pub(crate) fn unclaimed_report(flat: &Flat, claimed: &Claimed) -> Vec<String> {
     for key in sorted(claimed.ignored_types.iter().map(|k| k.as_str().to_owned())) {
         let named = TypeKey::parse(&key)
             .ok()
-            .and_then(|k| bare_path_ident(&k.to_type()))
+            .and_then(|k| k.ident())
             .is_some_and(|ident| flat.declared_type(&ident).is_some());
         if !named {
             out.push(format!(
--- a/prebindgen/src/api/core/flat/boundary.ledger
+++ b/prebindgen/src/api/core/flat/boundary.ledger
@@ -55,11 +55,10 @@
 4	api/lang/jnigen/jni/emit/convert.rs
 2	api/lang/jnigen/jni/emit/delivery.rs
 5	api/lang/jnigen/jni/emit/flat_input.rs
-17	api/lang/jnigen/jni/emit/names.rs
+16	api/lang/jnigen/jni/emit/names.rs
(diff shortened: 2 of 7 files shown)


milyin (commented):
[P1] Preserve path segments after generic arguments

`short_name()` truncates the canonical string at the first `<`, which loses any path segments that follow a generic segment. For example, `a::Foo<u8>::Bar` is a valid `syn::Type::Path`; the previous last-segment walk returns `Bar`, and `bare_path_ident` also returns `Bar` because the final segment itself has no arguments. The new implementation instead returns `Foo` from `short_name()` and `None` from `ident()` because the key contains `<`.

This changes the migrated Kotlin model lookups/name derivation: they can look up `Foo` or silently skip a declaration that previously resolved as `Bar`. Adding `"a::Foo<u8>::Bar"` to `SHAPES` reproduces the issue immediately in `key_name_accessors_match_the_syn_walks`.

Please make the string walk nesting-aware (so it finds the final top-level path segment and checks arguments on that segment only), or retain the two derived names when constructing the key from the canonical AST.

PR #301, opened 2026-08-02
A key is only an identity (#291 stage D, closes #291)

### LINKED ISSUE #291: Make `TypeKey` blind: a key should not be a back-door to `syn::Type`

milyin:
Last of four on #291. **Stacked on #300 → #299 → #298** — all green; bases get retargeted as each merges.

## The last three readers

- **`option_depth`** peeled `Option<…>` tokens to count layers the model had already counted. `TypeKind::Optional` is produced for exactly `Option<T>` at arity 1, and `optional_inner()` names the layer. It takes the reading now, and both callers hand over readings they were already sitting next to (one iterated `type_table(dir).keys()` when `.values()` had the answer).
- **`KotlinMeta::value_rust_key`** held a `TypeKey` and had exactly **one** reader, which immediately spent it on `to_type()`. It was never an identity here — only a detour through one — and both producers have the `syn::Type` in hand. Now `value_rust_type: Option<syn::Type>`, carrying the same canonical form it always yielded.
- **The idempotence assertion** re-keyed `to_type()`, which was really a claim about the parsed form a key kept beside its string. A key keeps no such thing; the next line's string round trip is the whole claim and it stays.

## Then the field

```rust
#[derive(Clone)]
pub struct TypeKey {
    canon: std::rc::Rc<str>,
}
```

`from_type` stops allocating the second `Rc`. `parse` still parses — to **validate** — and discards it. `Eq`/`Hash`/`Ord`/`Debug`/`Display` never read anything else, so nothing about identity, ordering, or error text moves.

## What this closes

Not a miscompilation — none was known. It is that the type system permitted a category of mistake: *spell a type nobody classified*. #280 sealed `TypeRef` so only the model may mint a reading, and a key that handed out tokens walked straight around that seal. The route from a key to syntax is `Conversions::reading` + `TypeRef::syntax`, and now it is the only one.

**44 call sites → 0**, across four PRs, with generated output byte-identical at every step.

## On the issue's "one genuine loss"

#291 predicted ~5 production sites turning infallible into `Option`. In the end **none did**. Stage B gave the declare-phase sites their own `syn::Type` (where `reading()` would have answered `None` and silently changed behaviour), stage C1 answered the name sites from the key itself, and the three here needed no new failure mode. The two sites that did move to `reading()` — `SpecKey::WholeFolder` and `crossing_keys` — each sit beside a sibling that already did, and defer on `None` exactly as it does.

## Verified

- `cargo test --all --all-features` — 634 lib tests, all green
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release` on every example crate

### DIFF
--- a/prebindgen/src/api/core/registry/key.rs
+++ b/prebindgen/src/api/core/registry/key.rs
@@ -7,16 +7,32 @@ use quote::ToTokens;
 /// Canonical type-shape key: identity is the token string of the
 /// **normalized** type ([`crate::api::core::flat::spelling::normalize_type`] —
 /// group/paren unwrap, `crate::`/`self::` and std-prelude path reduction;
-/// the complete equivalence rule set is documented there). The normalized
-/// parsed form is kept alongside the string, so [`Self::to_type`] is an
-/// infallible clone — no core invariant depends on serialize-then-reparse
-/// round trips (issue #95).
+/// the complete equivalence rule set is documented there).
+///
+/// # A key is an identity, and nothing else
+///
+/// It is what a table is indexed by. It is **not** a route to `syn::Type`: the
+/// only way to reach a type's syntax is
+/// [`Conversions::reading`](super::Conversions::reading) followed by
+/// [`TypeRef::syntax`](crate::api::core::flat::TypeRef::syntax), because a
+/// reading is what pairs a spelling with the classification that vouches for
+/// it.
+///
+/// This used to keep the parsed form beside the string and hand it out through
+/// `to_type()`, which let any holder of a key produce tokens for a type the
+/// model never classified — the same capability #280 sealed `TypeRef` against,
+/// granted by the key itself. A caller that wants tokens now has to have gotten
+/// them from somewhere that knows what they mean: the registry's reading, or
+/// the declaration that wrote them (#291).
+///
+/// What a key can still answer about itself is what it is **called** —
+/// [`Self::as_str`], [`Self::ident`], [`Self::short_name`] — because a name is
+/// not syntax.
 #[derive(Clone)]
 pub struct TypeKey {
-    /// Canonical token string — the identity `Eq`/`Hash` compare.
+    /// Canonical token string — the identity `Eq`/`Hash` compare, and the whole
+    /// of what a key is.
     canon: std::rc::Rc<str>,
-    /// The normalized parsed form the string was rendered from.
-    ty: std::rc::Rc<syn::Type>,
 }
 
 impl PartialEq for TypeKey {
@@ -66,6 +82,10 @@ impl std::error::Error for TypeKeyParseError {}
 
 impl TypeKey {
     /// Build a key by parsing the input as a type and normalizing.
+    ///
+    /// The parse is kept for **validation** and then discarded: a key that
+    /// cannot be a type is a mistake worth reporting at the declaration, and a
+    /// key that can is still only its canonical string.
     pub fn parse(s: &str) -> Result<Self, TypeKeyParseError> {
         let ty: syn::Type = syn::parse_str(s).map_err(|error| TypeKeyParseError {
             input: s.to_string(),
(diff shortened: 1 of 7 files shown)


milyin (commented):
Reviewed the complete stage-D delta against `a-key-answers-its-own-name`.

No additional findings in this PR itself. The three remaining migrations preserve their prior semantics: cbindgen counts `Option` layers from the model reading, `value_rust_type` carries the same canonical type previously recovered from `value_rust_key`, and removing the cached AST leaves identity/ordering/error rendering unchanged.

Verification:
- `cargo test -p prebindgen --lib`: 561 passed
- `git diff --check`: clean

This remains conditional on fixing the stacked PR #300 blocker: the current head inherits the `a::Foo<u8>::Bar` `short_name()`/`ident()` regression reported there. Once #300 fixes that accessor and this branch is rebased, I see no further blocker in #301.

PR #302, opened 2026-08-03
jnigen: the jobject decoder reads the element, and finds a wrong descriptor (#289)

milyin:
First of two on #289. The issue text is stale — #294 already converted the *flatten* path and its trailer said "closes #289", which only fired against `language-integration`. Current counts were 9 `option_inner_type` / 8 `reading_of`, not the 20/12 the issue states.

## What

`struct_input_body` — the whole-object `.jobject_input()` decoder — took the `syn::ItemStruct` and walked `syn::Fields::Named`, while its caller was **already holding** the `flat::Struct` that `struct_type()` returned and reached back down through `&s.origin.syntax`. `flat::Field::ty` is a `TypeRef` the model classified when it parsed the item.

```rust
pub(crate) fn struct_input_body(
    ext: &Declarations,
    s: &flat::Struct,                      // was: &syn::ItemStruct
    registry: &impl Conversions<KotlinMeta>,
) -> Option<(syn::Type, syn::Expr)>
```

The optional layer is asked **once** and reused, as `build_flat_struct_node` has since #294 — four independent path-segment tests in one loop iteration could disagree with each other. The name comes off `TypeKind::Named`, "is it a run" off `sequence_elem()`, the enum probe off `is_kotlin_enum_reading`; each has its precedent in `struct_plan.rs` (#288).

## ⚠️ Not output-preserving — one golden moved, and it is a fix

The issue says a moving golden is a found defect to be diagnosed. It is:

```
 "boxed"  Ljava/lang/Object;  →  Ljava/lang/Long;
```

`WrappedFields { boxed: Box<Option<i64>>, plain: Option<i64> }` is the fixture #294 added *because those two fields mean the same thing*. Kotlin declares both `Long?`. The old emitter asked JNI for `Ljava/lang/Object;` for `boxed` and `Ljava/lang/Long;` for `plain` — `option_inner_type` reads the last path segment, so `Box<Option<i64>>` answered "not optional", the descriptor chain fell through to its `Object` fallback, and the twins diverged. The #273 signature exactly.

**`GetFieldID` requires the field's exact declared descriptor**, so that lookup throws `NoSuchFieldError`.

Why nobody saw it: `JObject_to_WrappedFields_*` is emitted and **never called** — this fixture crosses via the flatten path, and its siblings (`JObject_to_Payload_*`) appear 7× each while this appears once. Any `.jobject_input()` data class with a wrapped optional field would have hit it.

## Also

- Two residuals #294 left *inside* the converted fn: an `is_kotlin_enum` on a spelling three lines from the reading, and a `reading_of` re-looking-up a `&TypeRef` already in hand.
- `FlatFieldNode::Value::direct_handle` becomes `Option<Box<syn::Type>>` carrying the handle target the plan peeled, replacing a `bool` beside a spelling the **renderer** re-peeled with the same last-segment test — a `Box<Option<T>>` handle field would have been handed the wrong `Box::from_raw` target. The node stays token-carrying: it is an emission IR, so the answer travels from where the reading was rather than a `TypeRef` moving into it.

## Counts, including the one that does not move

- **boundary ledger 117 → 116** (`flat_input.rs` 5 → 4). Only the hand-rolled `syn::Type::Path` Option test classified captured source; the four left classify `entry.destination` — a wire the adapter itself produced, legitimately its business.
- **jnigen census 9 → 4.**
- **`types_util` does NOT move, and was never going to.** The issue predicted it would shrink; that is now wrong. `option_inner_type` keeps callers in `struct_out.rs` ×2, `trait_impl.rs`, `fold.rs`, `fn_plan.rs`, `wrapper.rs`.

The remaining 4 are the sum side — PR2. It needs `Type::Variant`, because `Flat::enum_item` deliberately hands back only the `syn::ItemEnum`.

## Verified

- `cargo test --all --all-features` — 636 lib tests
- `examples/regen-check.sh` after `cargo clean --release` — clean apart from the one diagnosed descriptor

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -3570,7 +3570,7 @@ pub(crate) unsafe fn JObject_to_WrappedFields_f14f08c1<'env, 'v>(
             >>::from(format!("WrappedFields.id: {}", e)))? as _;
         let id = jlong_to_i64_fbf9a9bc(env, &__id_raw)?;
         let __boxed_raw: jni::objects::JObject = env
-            .get_field(v, "boxed", "Ljava/lang/Object;")
+            .get_field(v, "boxed", "Ljava/lang/Long;")
             .and_then(|val| val.l())
             .map_err(|e| <__JniErr as ::core::convert::From<
                 String,
--- a/prebindgen/src/api/core/flat/boundary.ledger
+++ b/prebindgen/src/api/core/flat/boundary.ledger
@@ -54,7 +54,7 @@
 4	api/lang/jnigen/jni/builder.rs
 4	api/lang/jnigen/jni/emit/convert.rs
 2	api/lang/jnigen/jni/emit/delivery.rs
-5	api/lang/jnigen/jni/emit/flat_input.rs
+4	api/lang/jnigen/jni/emit/flat_input.rs
 16	api/lang/jnigen/jni/emit/names.rs
 11	api/lang/jnigen/jni/emit/wrapper.rs
 3	api/lang/jnigen/jni/fold.rs
@@ -68,4 +68,4 @@
 2	api/lang/jnigen/jni/wire_access.rs
 2	api/lang/jnigen/util.rs
 
-# total: 117
+# total: 116
--- a/prebindgen/src/api/lang/jnigen/jni/emit/flat_input.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/flat_input.rs
@@ -283,9 +283,18 @@ pub(crate) fn struct_input_body(
         field_init.push(quote!(#fname_ident));
     }
 
+    // The struct's OWN delimiters, from the one place that chooses them.
+    // `flat::Struct` does not record whether its fields were named — that is
+    // spelling — so hard-coding braces here emitted `Unit {}` for
+    // `struct Unit;` and `Empty {}` for `struct Empty()`, neither of which is
+    // Rust. The `syn::Fields::Named` guard this walk replaced happened to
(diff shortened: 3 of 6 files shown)


milyin (commented):
[P1] Preserve the struct delimiter shape before iterating fields

Changing `struct_input_body` from `syn::ItemStruct` to `flat::Struct` removes the old `let syn::Fields::Named(..) = ... else { return None; }` guard. Checking `field.name` inside the loop is not equivalent for an empty unit or tuple struct: there is no field to reject, so the function now reaches the hard-coded braced initializer and emits `myflat::Unit {}` for `pub struct Unit;` (likewise `Empty {}` for `struct Empty()`). That Rust is invalid; previously these shapes returned `None` rather than emitting a converter that cannot compile.

I reproduced this with a `data_class!(Unit).jobject_input()` fixture: the generated converter contains `Ok({ myflat::Unit {} })` and the regression assertion fails.

Please retain an upfront named-fields shape check using `s.origin.syntax.fields`, or construct through the model’s delimiter-aware `Struct::spell` if unit/empty-tuple support is intended.

milyin (commented):
Re-reviewed the changes since c94f6a5.

The previous blocker is resolved. struct_input_body now constructs through flat::Struct::spell, so the model-owned delimiter spelling is preserved: Unit remains bare and EmptyNamed {} retains braces. This is the right boundary—field semantics still come from TypeRef, while constructor delimiters remain source spelling.

The new empty_structs_keep_their_own_constructor_delimiters regression covers both reachable empty shapes; tuple structs correctly remain outside this path because the model exposes them as Extern.

No remaining findings.

Verification:
- cargo test -p prebindgen --lib: 564 passed
- git diff --check: clean

PR #303, opened 2026-08-03
jnigen: the sum decoder reads the element too (#289)

milyin:
Second of two on #289. **Stacked on #302** — base retargets to `language-integration` once that merges.

## What

`sum_input_body` took the `syn::ItemEnum` and ran **two zips** to get back to per-field types: a `SumSpec` derived from the item, paired back against the item it was derived from. `Alternative` already *is* that pairing — name, index, and a `Vec<Field>` whose `ty` is a `TypeRef`. Both zips and the `SumSpec` go.

It needs a different accessor to get there: `Flat::enum_item` hands back only the `syn::ItemEnum`, deliberately — its own doc says a consumer acting on the Variant/Enum distinction should ask `declared_type`. Both `sum_input_body` and `build_flat_sum_field` do that now and match `Type::Variant`.

## The delimiters were the interesting part

The constructor came from a three-arm `syn::Fields` match. The obvious port — `Alternative::is_empty()` — is **wrong**, and the model says so itself: that is the *group* question, and `B`, `B()` and `B {}` are all empty by it. Rust demands the delimiters wherever the last two are named (`E::B` for `enum E { B() }` is a constructor *function*, not a value).

`Alternative::spell` is the designated helper — "the one place those delimiters are chosen, for match patterns and constructors alike" — and `Field::bind` shapes each init to match. So the three-arm match was standing in for something the model already owns.

## One derivation, not two

`sum_field_prop_name` now takes a `&syn::Member` rather than a `&SumField`. The member is the whole of what the name depends on, so a caller holding a `flat::Field` asks `Field::member()` and a caller holding a `SumField` reads its own — rather than a second convention that could drift from the sealed-interface emitter's.

## flat_input.rs is at zero

No `option_inner_type`, no `reading_of`, no `bare_path_ident`, no `pat_match_top`, no `SumSpec`, no `syn::Fields`. The file comes **off** the spelling census (4 → 0; 18 → 0 across #294 + #302 + this), which is what the census calls a migrated file.

**The boundary ledger does not move, and should not.** Its four remaining entries here classify `entry.destination` — a wire type the adapter itself produced, legitimately its business rather than a reading it should have asked for.

## SumSpec stays, for now

Retiring it was the agreed destination, but it still has callers in `struct_plan.rs`, `kotlin_emit.rs` and `sum_out.rs`. Doing it here would put the Kotlin emitters and the sum *output* path in a PR about input decoding. It also owns one thing the model does not — the leaf-naming convention — so that has to be rehomed rather than deleted. Worth its own PR.

## Verified

- `cargo test --all --all-features` — 636 lib tests
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release` — unlike #302, the sum path had no latent descriptor bug

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/flat_input.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/flat_input.rs
@@ -315,26 +315,28 @@ pub(crate) fn struct_input_body(
 /// design's rejected-alternative note about per-crossing JVM objects. Reading
 /// one field out of a `JObject` the caller already handed us costs nothing
 /// extra, so the asymmetry is real rather than an oversight.
+/// Takes the **element**, not the `syn::ItemEnum` it was parsed from (#289):
+/// `Alternative::fields` carries a `TypeRef` per payload, so the property read
+/// below asks the model instead of peeling tokens. It also retires the two zips
+/// this used to run — a `SumSpec` derived from the item, paired back against the
+/// item it came from — because `Alternative` already is that pairing.
 pub(crate) fn sum_input_body(
     ext: &Declarations,
-    e: &syn::ItemEnum,
+    v: &flat::Variant,
     registry: &impl Conversions<KotlinMeta>,
 ) -> Option<(syn::Type, syn::Expr)> {
-    use crate::api::core::types_util::SumSpec;
-
-    let key = TypeKey::from_ident(&e.ident);
+    let key = TypeKey::from_ident(&v.name);
     let cfg = ext.types.get(&key)?;
     let sum_cfg = cfg.sum()?;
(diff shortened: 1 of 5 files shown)


milyin (commented):
Reviewed the complete delta against `flat-input-reads-the-element`.

No additional findings in #303. The migration to `flat::Variant` removes both item zips without losing field order/addressing, and `Alternative::spell` correctly preserves the otherwise easy-to-lose distinction among `B`, `B()`, and `B {}`. The property decoder now consistently uses each payload’s existing reading.

Verification:
- `cargo test -p prebindgen --lib`: 563 passed
- `git diff --check`: clean

This is conditional on fixing the stacked #302 unit/empty-tuple struct constructor regression reported there; once #302 is corrected and this branch is rebased, I see no further blocker in #303.

PR #304, opened 2026-08-03
jnigen: a sum's match pattern keeps its own delimiters

milyin:
Found while scoping the `SumSpec` retirement (#289 follow-up). Standalone because it is a **bug fix**, not a refactor — it should not wait on the larger change.

## The defect

`encode_sum_group` built each arm's pattern by branching on the first field:

```rust
let pattern = match variant.fields.first().map(|f| &f.member) {
    None                     => quote!(#source::#vident),   // ← empty ⇒ bare
    Some(Member::Named(_))   => quote!(#source::#vident { … }),
    Some(Member::Unnamed(_)) => quote!(#source::#vident(…)),
};
```

For `enum E { B() }` and `enum E { B {} }` that emits `myflat::E::B`, which is **E0533** in pattern position: a zero-field tuple or struct variant still needs its delimiters.

Branching on the first field is structurally unable to answer it — an **empty** alternative has no first field to branch on. That is the same shape as the empty struct that had no field to refuse, which is how #302 came to emit `myflat::Unit {}`.

**Three instances of one defect class in this area now:**

1. #302 — a constructor, `struct Unit;` → `Unit {}`. Caught in review.
2. #303 — a constructor, `enum E { B() }` → avoided by using `Alternative::spell`.
3. **this** — a *pattern*, `enum E { B() }` → `E::B`. Nothing had found it.

## The fix

`Alternative::spell`, whose doc names the case exactly: *"the one place those delimiters are chosen — for match patterns and constructors alike, in either direction."* #303 used the constructor half; this is the pattern half. `Field::bind` shapes each binding.

Getting there needs the element rather than the item, so the arm list comes off `Flat::declared_type` → `Type::Variant` and walks `alternatives`. `enum_item` returns only the `syn::ItemEnum`, deliberately — its own doc says a consumer acting on the Variant/Enum distinction should ask `declared_type`. The tag becomes `alt.index`, which is what `SumVariant::tag` was a copy of.

This removes one of the five `SumSpec` call sites as a side effect; the other four are the follow-up PR.

## Verified

- `empty_sum_alternatives_keep_their_own_pattern_delimiters` covers unit / empty-tuple / empty-struct alternatives, and **fails against the old branch** (checked by reverting).
- `cargo test --all --all-features` — 638 lib tests
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release` — no in-tree fixture declares an empty non-unit alternative, which is exactly why this had no signal

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/sum_out.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/sum_out.rs
@@ -239,20 +239,23 @@ pub(crate) fn encode_sum_group(
         .expect("a sum plan carries its selector leaf");
     let tag_id = &obj_idents[tag_idx];
     // A unit variant contributes no leaf, so the arm list is driven by the
-    // enum's own variants, not by the grouped leaves.
-    let item_enum = registry.flat().enum_item(&ident).unwrap_or_else(|| {
-        panic!("jnigen sum unfold: no indexed enum `{ident}` for the decomposed sum")
-    });
-    let spec = crate::api::core::types_util::SumSpec::from_item_enum(item_enum);
+    // enum's own alternatives, not by the grouped leaves. `enum_item` hands
+    // back only the `syn::ItemEnum`, deliberately — a consumer that acts on the
+    // Variant/Enum distinction asks `declared_type` (#289).
+    let Some(crate::api::core::flat::Type::Variant(sum)) = registry.flat().declared_type(&ident)
+    else {
+        panic!("jnigen sum unfold: no indexed sum `{ident}` for the decomposed sum")
+    };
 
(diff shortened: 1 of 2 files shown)


milyin (commented):
Reviewed 39ced3f. No findings.

The fix is correctly placed at the model/spelling boundary: encode_sum_group now reads the canonical Variant alternatives, keeps the existing declaration-order tag and payload order, and delegates pattern syntax to Alternative::spell with Field::bind. That preserves bare, tuple, and named delimiters—including empty alternatives—without regressing raw identifiers.

Verification:
- focused empty-alternative regression: passed
- cargo test -p prebindgen --lib: 565 passed
- git diff --check: clean

PR #305, opened 2026-08-03
core: delete `SumSpec` — the model already describes a sum

milyin:
The #289 follow-up. **Stacked on #304** — base retargets to `language-integration` once that merges.

## What

`SumSpec`/`SumVariant`/`SumField` described a data-carrying enum as "a tag plus one leaf group per variant". `flat::Variant` describes the same thing and better: `Alternative` carries the name, the declaration-order index and a `Vec<Field>` whose `ty` is a classified `TypeRef`, where `SumField` kept a bare `syn::Type`. It was #211's own thesis sitting inside `api/core`.

Every live field had an exact model equivalent:

```
v.ident     → alt.name          v.tag    → alt.index
v.is_unit() → alt.is_empty()    f.member → field.member()
```

**Net −243 lines.**

## The blocker was not real, and the code said it was

I said twice — and `emit/sum_out.rs:53` and `kotlin_emit.rs:611` both say — that `SumSpec` "owns the leaf-NAMING convention, which is jnigen's own", so retiring it needed that rehomed first.

**Nothing read `SumField::name`.** jnigen names its slots with `sum_field_prop_name` + `sum_slot_fragment` (`struct_plan.rs`), a *different* convention. `SumField::ty` was dead for the same reason one layer down: every site already took `alt_field.ty` — the reading — off the element beside it. `SumSpec::{key, source}` were dead too. Four of nine fields had no reader.

The doc's other premise was also unmet: *"both adapters read one definition instead of growing a private one each"* — **cbindgen never used it**. The `#[allow(dead_code)]` on all three structs was the tell.

## The sites

Three of four already held the `&flat::Variant` and **zipped `SumSpec` back against `sum.alternatives`** — derived from the item, then re-paired with the element it was derived from. That is exactly the shape #303 removed from `flat_input.rs`. The fourth reached `Flat::enum_item` and moves to `declared_type` → `Type::Variant`, as #304 did.

Also removed: `kotlin_emit`'s `sum_field_property_name` (a second copy of `sum_field_prop_name` keyed on the deleted type), and a redundant `enum_item` lookup in `struct_plan` sitting beside the `declared_type` that did the real work.

## Counts do not move, and were not going to

`SumSpec` names no `syn::Type` variant, so the **boundary ledger never counted it**; the **spelling census** counts `option_inner_type`-family calls, which it never made. Saying so up front rather than predicting a drop — the last several PRs each had to report a count that stayed put.

The five deleted `types_util` fixtures tested the declaration-order tag; `flat/tests/acceptance.rs:934` already asserts that on `Alternative::index`.

## Verified

- `cargo test --all --all-features` — 633 lib tests (638 − 5 deleted fixtures)
- `cargo doc` — clean of the two intra-doc links the deletion orphaned (`pascal_to_snake`, `enum_discriminant_values`)
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release`


milyin (commented):
Reviewed f608faf against the exact #304 head. No findings.

The replacement is behaviorally consistent at all four call sites: flat::Variant already owns alternative order and indices, Alternative owns the field list, Field::member preserves named and positional addressing, and each Field carries the classified TypeRef previously taken from the parallel model. The shared sum_field_prop_name implementation is equivalent to the deleted duplicate, so Kotlin property and slot naming remain stable. The redundant descriptor and enum lookup can be removed without changing the wire tags, payload ordering, or generated surface.

Verification:
- cargo test --all --all-features: passed, including 633 prebindgen library tests and doc tests
- cargo doc --all --all-features --no-deps: succeeded
- git diff --check: clean

PR #306, opened 2026-08-03
core: a `SumTag` selector registers the sum it names (#282)

milyin:
Closes #282. The issue asked for one decision — is `LeafSource::SumTag` boundary-crossing data (registrable) or adapter-only metadata (intentionally unregistered)? **Decision: it gets a cell.**

## The issue text is stale; only registration was open

`sum_out.rs` stopped composing `TypeRef::named(enum_ident)` when #280 sealed the composers (`named` is `pub(super)` to `core::flat`), and the leaf now carries `Variant::type_ref()` — the reading the **declaration** stored. So "may the adapter compose here" is already closed. Registration was the open half.

## `require_output` could not serve it

It is `register_type_recursive(.., root = true)`, and a root **demands** a converter. A sum has no whole-value *output* converter, so requiring one fails resolution — pulling the tag's `i32` in that way is the original reason `has_converter()` exists at all.

So registration and demand had to separate:

```rust
/// Register `reading` as an output cell WITHOUT demanding a converter.
pub(crate) fn reference_output(&mut self, reading: &TypeRef)
```

named after the existing `RegistryBuilder::reference` vocabulary ("scanned, not required"), and both leaf loops now **split** on `has_converter()` instead of filtering by it. `ensure_entry` does `cell.root |= root`, so referencing a type the binding did declare cannot weaken it.

## The invariant, stated where it is checkable

> Every leaf's `out_ty` has a table cell; only a converter-bearing leaf is a root.

A **cell** says the type entered the pipeline, a **root** says the binding asked for it directly, an **entry** says one resolved — three claims, and a `SumTag` leaf makes only the first. Written on `has_converter`, `LeafSource::SumTag`, and `TypeRef` (whose doc already said a reading claims no converter, and now says it claims no cell either).

## What this buys — the old test proves it

The invariant held before **only because jnigen happens to declare the sum** via `export_type`. `unfold/tests.rs`'s assertion read `!...is_some_and(|c| c.root)`, which is *also true when the cell is absent* — and absent is exactly what it was, since that fixture's registry declares nothing. It passed for the wrong reason and could state neither half.

It now asserts the invariant over every leaf of the plan, and **fails against the old filter** (checked by reverting). Its message was wrong too: "a sum has no whole-value converter" is true only output-side.

Also worth noting: that fixture's tag leaf carries `out_ty: i32`, where the real emitter stores the **sum**. Asserting the invariant per-leaf rather than naming `Reading` sidesteps the divergence.

## The end-to-end claim the acceptance asks for

New test against a real `Registry` via `JniGen::registry()`: for a declared sum, a cell **both** ways, root cleared **both** ways, an input `entry` (the whole-`JObject` decoder) and **no** output entry. That asymmetry is the design — Rust → Kotlin is flattened, always.

**Not doing:** a source-scanning guard that `api::lang` never calls the `TypeRef` composers. They are `pub(super)` / `pub(in crate::api::core)`, so the compiler enforces it every build, and `ty.rs:96-100` says a scan test is the wrong instrument here.

## Verified

- `cargo test --all --all-features` — 634 lib tests
- `cargo fmt --check` with the CI config; `cargo doc` clean
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release`. This was the specific risk: `crossings()` seeds from every table key, so a *newly created* cell would add a crossing — but jnigen's sum already had one from its declaration, so nothing entered the order.

Ledger and census do not move, and were not going to: no `syn::Type` match, no spelling-helper call.

### DIFF
--- a/prebindgen/src/api/core/flat/ty.rs
+++ b/prebindgen/src/api/core/flat/ty.rs
@@ -59,6 +59,15 @@ use crate::SourceLocation;
 /// false by design for stored readings — `unrequire_output` leaves a cell whose
 /// converter genuinely cannot resolve, and a `SumTag` leaf never has one — so
 /// converter existence stays a lookup that answers `Option`.
+///
+/// It does **not** claim a registry cell either, and the two are separate
+/// questions. Holding a `TypeRef` means the model classified the type; whether
+/// it is in a type table is the registry's business, and the registry states it
+/// in three parts — a **cell** (the type entered the pipeline), a **root** (the
+/// binding asked for it directly), an **entry** (a converter resolved). A
+/// `SumTag` leaf's type makes the first and not the second, deliberately
+/// (#282); see
+/// [`Registry::reference_output`](crate::api::core::registry::Registry::reference_output).
 #[derive(Clone, Debug)]
 pub struct TypeRef {
     /// What the type means — the closed, destination-neutral classification.
--- a/prebindgen/src/api/core/registry/scan.rs
+++ b/prebindgen/src/api/core/registry/scan.rs
@@ -550,6 +550,27 @@ impl<M> Registry<M> {
         self.register_type_recursive(Direction::Output, reading, true);
     }
 
+    /// Register `reading` (and its nested positions) as an **output cell without
+    /// demanding a converter** — a type some plan *names* rather than one that
+    /// crosses.
+    ///
+    /// The third thing a table cell can mean, now said out loud. A cell records
+    /// that a type **entered the pipeline**; `root` records that the binding
+    /// asked for it *directly*; `entry` records that a converter resolved. This
+    /// makes the first without the second, which is exactly what a
+    /// [`SumTag`](crate::api::core::unfold::LeafSource::SumTag) selector needs:
+    /// it names *which* sum it chooses between, and that sum has no whole-value
+    /// output converter at all, so requiring one would fail resolution (#282).
+    ///
+    /// **Not [`require_output`](Self::require_output) with a flag.** That one is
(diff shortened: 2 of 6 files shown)


milyin (commented):
[P2] Make the fixture carry the production selector type

In prebindgen/src/api/core/unfold/tests.rs:1696, reading_sum_decon says it mirrors the JNI synthesis but gives the SumTag leaf out_ty = i32. Production synth_sum_leaves gives that leaf sum.type_ref(), so this test currently proves only that an arbitrary converter-free leaf is registered, not that the selector registers the sum it names. The new JniGen acceptance test cannot close that gap because sealed_class!(Reading) already creates the Reading output cell even with the old filtered loop.

Please change the fixture selector to tref(Reading). I verified that exact change passes on this PR and fails against the old filter with leaf `tag` registers its out_ty, directly pinning the behavior from #282 and removing the admitted fixture/production divergence.

Verification otherwise passed:
- cargo test --all --all-features: 634 library tests plus doc tests passed
- git diff --check: clean

— Codex (GPT-5)

milyin (commented):
Re-reviewed through c00f379. The previous P2 finding is resolved, with no remaining findings.

The fixture now gives the SumTag selector the production Reading out_ty, so sum_return_is_a_fixed_builder_plan directly pins that the selector registers the sum it names. This passes on the updated implementation and, as verified during the original review, fails against the old filtered registration. The JniGen acceptance test now also documents its separate declaration-state scope clearly.

Verification:
- focused sum_return_is_a_fixed_builder_plan regression: passed
- cargo test -p prebindgen --lib: 561 passed
- git diff --check: clean

— Codex (GPT-5)

PR #307, opened 2026-08-03
jnigen: one reading of the movability rule (#223 item 1)

milyin:
#223 item 1 — the one it calls *"worth doing regardless"*. Two divergences in the reach family, both latent, and neither unreachable for a reason the emitter states.

Note two of the issue's premises are **stale**: `wrapper.rs` no longer hand-rolls a reach (it calls `bind_hoists` + `reach_leaf_flat`, shared with `encode_plan_leaves`), and all four struct-field walks moved to `flat::Struct::fields`. What survives is smaller and sharper.

## (a) A second reading of `steps_are_movable`

`plan.rs:163-166` says of it:

> This is the one place the rule is written… **Two readings of it would drift, and the disagreement would be a borrow handed to an owning converter.**

`reach_leaf_flat` was the second reading, spelled `path.iter().all(PathStep::is_plain_field)`. **They already disagreed**: the rule permits a *trailing* optional field — a `None` arm still hands the whole `Option` over by value — and the restatement did not. So where `place_is_owned` granted an owned `out_ty` *on the strength of the rule*, the emitter projected a borrow into the owning converter that `out_ty` had selected. PR#221's P1, one path shape away.

Its comment **defended** the restatement: a trailing optional can't reach return delivery, since `single_return` routes a nullable leaf to callback delivery. That's true, and it is the shape of the hazard rather than a defence against it — a local restatement can disagree with the rule for as long as an invariant *somewhere else* keeps the disagreement unreachable.

## (b) A guard that inspected the wrong path

The optional-step assert asked its `path` **parameter**. `wrapper.rs` rebases onto a hoisted local and hands over the remaining suffix, `Hoisted::innermost` having stripped the prefix that bound it — optional step and all. So the guard passed **exactly when the hoist was the conditional one**, which is the case that cannot compose: an `Option<T>` local with a field read hung off it. It asks `leaf.path` now.

## (c) The missing test level

**No test called a reach function directly.** Every pin was an end-to-end string match on generated Rust — which is why a latent divergence had no failing test. Three unit tests now do, and each fails against the code it replaced. The movability one reports the defect verbatim:

```
a movable place is moved, not borrowed or cloned — got `(& (& __src . a) . b) . clone ()`
```

They ask `test_util::reading` rather than `Flat::classify`, which #280 sealed to `api::core` — a test under `api::lang` meeting that boundary is the boundary working.

## Verified

- `cargo test --all --all-features` — 637 lib tests
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release` — expected, since neither divergence is reachable today

Ledger and census unmoved: no `syn::Type` match added, no spelling-helper call.

Follow-up PR covers the three duplicates exploration turned up (`sum_slot_name` being a byte-for-byte copy of `sum_slot_fragment`; `classify_field` asking the spelling where `flat_input` asks the model; two camel-casers for one Kotlin property name). Item 3 — `UnfoldPlan` keeping its tree — stays deferred until #220 forces it, as the issue advises.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/delivery.rs
@@ -456,8 +456,16 @@ pub(crate) fn reach_leaf_flat(
     // somewhere to go. This derivation has none — it yields a plain Rust value,
     // not a `JObject` that could be null — so the shape is refused here rather
     // than composed into code that cannot type-check in the consumer's crate.
+    //
+    // Asked of the leaf's OWN path, not of `path`. The caller may hand a
+    // suffix: `wrapper.rs` rebases onto a hoisted local, and `Hoisted::innermost`
+    // strips the prefix that bound it — including any optional step inside it.
+    // Checking the parameter would therefore pass exactly when the hoist is the
+    // conditional one, which is the case that cannot compose (an `Option<T>`
+    // local with a field read hung off it). The full path is what the shape
+    // question is about.
     assert!(
-        !path.iter().rev().skip(1).any(PathStep::is_optional),
+        !leaf.path.iter().rev().skip(1).any(PathStep::is_optional),
         "jnigen unfold: leaf `{}` reaches through an optional step but is \
          delivered as a single return value, which has no `None` arm — this \
          shape needs callback delivery",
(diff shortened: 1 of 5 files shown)


milyin (commented):
Reviewed 606c662. No findings.

The change removes the duplicated movability interpretation at the correct point: reach_leaf_flat now uses steps_are_movable, matching the ownership decision that selected the leaf out_ty. Checking leaf.path for nested optional steps is also necessary because a hoist rebase can legitimately strip the conditional prefix from the path slice passed to the reach function. The three direct tests distinguish both old implementations and preserve the borrowed-output case.

Verification:
- focused delivery reach tests: 3 passed
- cargo test --all --all-features: passed, including 637 prebindgen library tests and doc tests
- git diff --check: clean

— Codex (GPT-5)

milyin (commented):
Re-reviewed through 4758f12. No findings.

The follow-up corrects the fixture semantics: identity and accessor-path leaves now use LeafSource::Accessor, matching the resolver, while the dedicated Field fixture pins the clone terminal treatment for synthesized data-class fields. This makes source load-bearing in the tests and ensures the movability regression exercises a reachable production shape.

Verification:
- focused delivery reach tests: 4 passed
- cargo test --all --all-features: passed, including 638 prebindgen library tests and doc tests
- git diff --check: clean

— Codex (GPT-5)

PR #308, opened 2026-08-03
jnigen: one name, one enum question (#223 cheap wins)

milyin:
The follow-up to #307. **Stacked on it** — base retargets to `language-integration` once that merges.

Three places where one question had two implementations — #223's own thesis, in the naming/classification layer it does not list. All found while checking whether item 2 ("one struct-field walk") was still tractable; it is **not**, because `classify_field` is output-only and needs *resolved* converters while `synth_value_struct_leaves` runs pre-resolve and `build_flat_struct_node` is input-side. These are the value that was left in that area.

## 1. `sum_slot_name` was a byte-for-byte copy

Of `sum_slot_fragment` — same lower-first-char plus `_` join. `kotlin_emit` called the copy at one site and the shared `sum_field_prop_name` at two others, so a slot name and the property inside it came from different files. Deleted.

## 2. `classify_field` asked the spelling where the model has the answer

It called `is_kotlin_enum` twice, on two spellings, where `flat_input` asks `is_kotlin_enum_reading`. `builder.rs:109-111` documents the difference: a `Box<Priority>` field is **false** for the first and **true** for the second, so a wrapped enum field classifies as a plain leaf and renders as its wire instead of the Kotlin enum class — the #273 family, output-side.

It asks **once** now, of `bare_ref` — the already-peeled reading sitting beside it. Optionality stays the caller's fact: `enum_probe` peels `Option` as well as borrows, so probing the unpeeled reading would make `Priority` and `Option<Priority>` indistinguishable and collapse two arms into one.

### ⚠️ A pre-existing gap found while trying to prove this

I added a `Box<Priority>` field to `WrappedFields` — the fixture #294 added for exactly this bare/wrapped pairing — to demonstrate the fix. **It does not resolve:**

```
Unresolved { key: TypeKey("Box < Priority >"), direction: Output }
```

Reverting the classification change leaves it failing identically, so this is a **pre-existing capability gap**, not this PR's to fix: a `Box<enum>` data-class field cannot be built at all today, in either classification. The fixture is therefore **not** in this PR, and the change stands on the question being the right one to ask rather than on a demonstration it cannot yet have.

It is also a clean instance of #223's own thesis — capability depending on *where the type appears* — and probably wants its own issue.

## 3. Two camel-casers named one Kotlin property

`render_data_class_source` **declared** the property with `kt_snake_to_camel`; `flat_input`'s access expression and its `GetFieldID` slot name used `util::snake_to_camel`, which additionally lower-cases the first character.

They agree for a conventional lower-snake field and only for that — a field spelled `Xyz` is declared `Xyz` and read as `xyz`, and `GetFieldID` for a name that is not the declared one **fails at runtime**. One `kotlin_property_name` now serves the declaration and both readers.

`snake_to_camel` stays where it names *parameters* — a different namespace with no declaration to match. `symbols.rs`'s mangling warning already used the kept caser and deliberately passes pre- and post-mangle names separately, so it needed no change.

## Verified

- `cargo test --all --all-features` — 637 lib tests
- `examples/regen-check.sh` **byte-identical** after `cargo clean --release` — expected for all three: the two functions were textually identical, no in-tree field name is unconventional, and the enum divergence needs a shape that does not currently resolve

Ledger and census unmoved. #2 removes a spelling-keyed question but `is_kotlin_enum` keeps other callers, so the census row does not drop.

### DIFF
--- a/prebindgen/src/api/lang/jnigen/jni/emit/flat_input.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/emit/flat_input.rs
@@ -35,8 +35,9 @@ pub(crate) fn struct_input_body(
         // field now, because the element models a field list rather than a
         // `syn::Fields` shape.
         let fname_ident = field.name.clone()?;
-        let fname = fname_ident.to_string();
-        let camel = mangle_kotlin_ident(&snake_to_camel(&fname));
+        // The name the property was DECLARED with — `GetFieldID` takes the
+        // slot's exact name, so this cannot be derived a second way.
+        let camel = kotlin_property_name(&fname_ident);
         let err_prefix = format!("{struct_name}.{camel}: {{}}");
         let raw_ident = format_ident!("__{}_raw", fname_ident);
 
@@ -1459,7 +1460,7 @@ fn build_flat_struct_node(
                 "only named-field structs can flatten",
             ));
         };
-        let fcamel = mangle_kotlin_ident(&snake_to_camel(&fident.to_string()));
+        let fcamel = kotlin_property_name(&fident);
         let child_native = format!("{native_prefix}_{}", fident);
         let field_ref = if nullable_context {
             format!("{access_prefix}?.{fcamel}")
--- a/prebindgen/src/api/lang/jnigen/jni/kotlin_emit.rs
+++ b/prebindgen/src/api/lang/jnigen/jni/kotlin_emit.rs
@@ -423,26 +423,6 @@ impl Declarations {
     }
 }
 
-/// Slot name of one variant field in the flattened `fromParts` signature:
-/// `<variantCamel>_<property>` (`periodicQueries_period`, `pair_v0`) — the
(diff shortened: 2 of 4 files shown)


milyin (commented):
Reviewed 004502c against the exact #307 head. No findings.

All three consolidations preserve or improve the intended semantics:
- sum_slot_fragment is byte-for-byte equivalent to the removed local helper and now owns the variant-slot convention everywhere;
- classify_field asks the classified reading once, after the caller separates optionality, so transparent enum wrappers cannot drift from flat_input while Vec remains a non-enum run;
- kotlin_property_name now serves both property declaration and JNI access/GetFieldID naming, eliminating the uppercase-leading mismatch without changing parameter namespaces.

The Box<enum> unresolved-output limitation described in the PR is pre-existing and remains cleanly outside this change.

Verification:
- cargo test --all --all-features: passed, including 637 prebindgen library tests and doc tests
- git diff --check: clean

— Codex (GPT-5)

PR #310, opened 2026-08-03
jnigen: the transparent bridge gets its outbound half (closes #309)

milyin:
Closes #309. Four commits: the missing capability, two defects found while building it, and the fixture that started it.

The model erases `Box` and `Cow` deliberately — `Box<Priority>` **is** `Priority` to every destination language. #294 gave the *input* selector a last resort for the case no layer arm claims (a wrapper over a **terminal**). `select_output_type` never got the twin.

## 1. The outbound arm

The dual inverts two lines, because the wrappers come **off** rather than going on:

```
input : let __inner = <inner>(env, v)?;   build_through_erased_wrappers(__inner)
output: let __inner = read_through(v);    <inner>(env, __inner)
```

`read_through_erased_wrappers` was already the operation. Everything else is direction-independent — inheriting the inner's `metadata` is what keeps `Box<Priority>` presenting as the Kotlin enum class rather than losing it behind the wrapper.

**One arm covers `Box<Handle>`, `Box<enum>` and `Box<DataClass>`** — `output_terminal` misses all three the same way, by keying on the *spelling*. The acceptance item answered, and asserted rather than argued.

## 2. ⚠️ The bridge skipped the inner's stages — and did not compile

`input_transparent_bridge` called the inner converter's function directly and left `pre_stages` empty, where every other composing arm uses `composed_inner_*`. A `convert!`-declared type reaches its Rust value *through* those stages, so a `Box` over one skipped them:

```rust
let __inner = jlong_to_u64_4384a5d6(env, v)?;
::std::boxed::Box::new(__inner)
//                     ^^^^^^^ expected `Duration`, found `u64`   [E0308]
```

Not a subtly wrong value — **the generated crate does not compile**. `boxed_duration_echo` is the fixture and it is load-bearing: revert this commit with it in place and the build fails with that error.

## 3. ⚠️ The inbound bridge has been resolving by alphabetical luck since #294

The deeper find. Whoever converts `Box<T>` delegates to `T`'s converter — a real dependency the `kind` walk cannot see, since `Box<T>` *classifies as* `T` and differs only in spelling. `subs` said "required"; nothing said "first". And `convert_with` is a **single pass in dependency order**.

So `Box<Payload>` resolved only because some other root's fields happened to pull `Payload` in earlier. Measured: rename `Priority` → `APriority` and `Box<APriority>` resolves while `Box<ZSample>` still does not, purely because `"ZSample"` sorts after `"Box < ZSample >"`. **A capability that depends on the alphabet is not one.**

`immediate_edges` now yields the stripped spelling as an edge. Goldens byte-identical — it replaces luck with an edge rather than changing output.

## 4. The fixture that could not be built

`WrappedFields` gains `Box<Priority>` beside `Priority`, matching the `Box<Option<i64>>` / `Option<i64>` pair it already carries. It now builds:

```kotlin
public data class WrappedFields(
    val id: Long, val boxed: Long?, val plain: Long?,
    val boxedEnum: Priority, val plainEnum: Priority,
)
```

`boxedEnum` presenting as `Priority` rather than its `Int` wire is **#308's** fix — which landed correct and undemonstrable, because no `Box<enum>` field could be built to show it. This is its first end-to-end evidence. The Kotlin exercise weighs the two enum fields against each other, so the claim is that wrapped and bare behave alike, not merely that the wrapped one compiles.

## Verified

Per commit, with each claim proven to bite (revert → fail → restore):

- `cargo test --all --all-features` — 639 lib tests
- clippy `-D warnings` on **both** 1.85.0 and stable, from the repo root
- `examples/regen-check.sh` after `cargo clean --release` — **byte-identical for commits 1 and 3**; commits 2 and 4 move goldens deliberately, by their fixtures alone
- `examples/covertest-kotlin$ ./gradlew run` — PASS, 49 sections, including both new exercises

Ledger and census unmoved: nothing here adds a `syn::Type` match or a spelling-helper call.

## Correction to the issue text

#309 says `every_erased_wrapper_has_ops` "already pins that each erased wrapper has both operations". It does not — its doc pins **membership, not capability**, and `Cow` deliberately carries `read: None`. The outbound bridge refuses `Cow` through `read` exactly as the inbound one refuses it through `build`.

### DIFF
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -577,6 +577,7 @@ fn main() {
                 .fun(fun!(blob_value_echo))
                 .fun(fun!(arrays_echo))
                 .fun(fun!(duration_optional))
+                .fun(fun!(boxed_duration_echo))
                 .fun(fun!(duration_boundary_echo))
                 // The converted analogue of `unsigned_emit`: a whole-value
                 // callback argument, which encodes on its own path rather than
--- a/examples/covertest-kotlin/kotlin/REPORT.md
+++ b/examples/covertest-kotlin/kotlin/REPORT.md
@@ -63,6 +63,7 @@ Base package: `io.prebindgen.covertest`
   - shaped by: return `BlobValue` decomposed → [stamp__secs, stamp__nanos, id, chunks] (Callback delivery)
 - `blob_value_new` — `fun blobValueNew(secs: Long, id: ByteArray, chunks: List<ByteArray>, onError: JniErrorHandler<BlobValue>): BlobValue`
   - shaped by: return `BlobValue` decomposed → [stamp__secs, stamp__nanos, id, chunks] (Callback delivery)
+- `boxed_duration_echo` — `fun boxedDurationEcho(value: ULong, onError: JniErrorHandler<ULong>): ULong`
 - `boxed_elem_id_sum` — `fun boxedElemIdSum(ps: List<Payload>, onError: JniErrorHandler<Long>): Long`
 - `boxed_latest` — `fun <R> boxedLatest(a: SummaryVault, onError: JniErrorHandler<R?>, build: SummaryBuilder<R>): R?`
   - shaped by: return `Summary` decomposed → [count, total] (Callback delivery)
--- a/examples/covertest-kotlin/kotlin/generated/io/prebindgen/covertest.kt
+++ b/examples/covertest-kotlin/kotlin/generated/io/prebindgen/covertest.kt
@@ -940,6 +940,8 @@ internal object CovNative {
         errorSink: Any,
     ): Any?
 
+    external fun boxedDurationEcho(value: Long, errorSink: Any): Long
+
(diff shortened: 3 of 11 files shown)


milyin (commented):
Reviewed 923164f. No findings.

The outbound transparent bridge is the correct mirror of the existing input fallback: it removes the erased wrapper, delegates through the complete inner output chain, and preserves the inner destination metadata and niches. The new `immediate_edges` dependency also closes the ordering hole by placing the stripped converter before the wrapped spelling that delegates to it. Using `composed_inner_input` fixes the staged-conversion case without changing the single-stage route.

Coverage is appropriately layered: the acceptance test exercises enum, handle, and data-class terminals in both directions; the generated fixture compiles the staged `Box<Duration>` path; and the Kotlin runtime checks wrapped and bare enum presentation together.

Verification:
- focused terminal-bridge test: passed
- `cargo test --all --all-features`: passed, including 639 prebindgen library tests and doc tests
- Kotlin/JNI covertest: PASS, all 49 sections
- regeneration left the worktree clean
- `git diff --check`: clean

— Codex (GPT-5)

PR #313, opened 2026-08-04
core: a spelling is for spelling — seal the model's syntax

milyin:
The `TypeKind` pivot put every source fact in the model. The door stayed
open: `TypeRef::syntax()` handed out a `&syn::Type`, `Origin::syntax` was
a public field, and any consumer could still take the source apart
instead of asking. The boundary ledger *measured* that — 116 sites — and
its own header listed three kinds of classification it cannot see.

So the model stops handing out nodes:

    Origin::syntax   pub -> pub(super)
    Origin::spell()  -> TokenStream    output — free
    Origin::as_syn() -> &S             THE escape — counted

`TypeRef::syntax()` is **replaced** by `spell()` and `as_syn()`, and
`Element`/`Type::syntax()` become `as_syn()`. Removing the old name is
the point: each of its call sites had to say which of the two it meant,
and the compiler asked one by one — 260 of them.

**No `ToTokens` on `Origin`.** That trait is what made `quote!(#node)`
compile, and it hands `to_token_stream().to_string()` over with it —
a classifier's input in a spelling's clothing. A `TokenStream`
interpolates just as well one `let` earlier. It does not make a token
string impossible (`spell().to_string()` reaches one, and `domain.rs`
classifies a `syn::Type` a *build script* supplied, which the seal never
covers); it makes one **say so**.

## The ledger counts the door, not just the grep

    # total classification sites: 116   (unchanged — nothing reclassified)
    # total escapes: types:       115   MUST reach zero
    # total escapes: items:        28   expected to persist

A file with no escapes cannot classify source syntax *however it is
written*, so the blind spots that had no watched site — an ident compared
by name, a helper handed the node, a syn enum outside `WATCHED` — are
gated behind the same call. That is what widening `WATCHED` was trying to
approximate.

The bucket is read off the receiver: `origin` means a captured item's own
node, anything else means a type. Two over-counts land in the type bucket
deliberately, both in the safe direction — adapter declarations reuse
`Origin` for a placeless location, and carrying a spelling into an
adapter-owned `syn::Type` field is not classification either. The header
says so rather than implying it away.

## What the escape count is made of

Three kinds, and only the first is classification: taking a node apart
(what the first section already sees), keying by spelling
(`TypeKey::from_type` where `TypeRef::key()` is the answer), and carrying
a spelling into a `syn::Type`-typed adapter field (`ConverterImpl::subs`
/ `produced`). Each is a follow-up the count now names. This PR rewrites
no classifier.

**Did not move**: every generated artifact byte-identical, classification
counts unchanged, 643 lib tests + 22 doctests green, clippy + fmt clean
on 1.85.0 and stable.

### DIFF
--- a/prebindgen/src/api/core/flat/boundary.rs
+++ b/prebindgen/src/api/core/flat/boundary.rs
@@ -128,16 +128,28 @@ const HEADER: &str = "\
 # below count what the first section can only measure — every place that still
 # has a node to take apart.
 #
-#   ## escapes: types   `ty.as_syn()`         — MUST reach zero
-#   ## escapes: items   `f.origin.as_syn()`   — expected to persist
+#   ## escapes: types   `ty.as_syn()`, `stripped_syntax()`, `to_syn()`
+#                                             — MUST reach zero
+#   ## escapes: items   `f.origin.as_syn()`, `enum_item()`
+#                                             — expected to persist
 #
 # A type escape is a source type the model should have been able to answer for;
 # the classification sites above are the subset that visibly does classify. An
 # item escape is a captured item's own node, which an emitter re-stating a whole
 # item legitimately needs until items grow modelled accessors.
 #
-# The bucket is read off the RECEIVER: `origin` means the item's node, anything
-# else means a type. Two over-counts land in the type bucket on purpose, both in
+# The scan counts FOUR doors — `as_syn`, `stripped_syntax`, `to_syn`,
+# `enum_item` — and counts each by NAME rather than by call shape, so UFCS
+# (`TypeRef::as_syn(&ty)`) and a function item (`let read = TypeRef::as_syn;`)
+# are counted like a method call. A shape-matched rule missed both, which is a
+# ratchet that can be stepped around. `escape_surface_is_closed` keeps the list
+# honest: it reads the model's own surface and fails if a public method hands
+# out a structural syn node under a name the scan does not count — which is how
+# three of the four doors were found.
+#
+# The bucket is read off the NAME, then the RECEIVER: `enum_item` hands out an
+# item; for `as_syn`, a receiver of `origin` (or an `Origin::` qualifier) means
+# the item's node, anything else means a type. Two over-counts land in the type bucket on purpose, both in
 # the safe direction — adapter declarations reuse `Origin` for a placeless
 # location (`decl.rust_type`), and carrying a spelling into an adapter-owned
 # `syn::Type` field is not classification either. Both are follow-ups the count
@@ -312,7 +324,43 @@ impl Counts {
     }
 }
 
-/// Which bucket an `as_syn()` call belongs to, read off the **receiver**.
+/// Every route from the model to a structural `syn` node.
+///
+/// `as_syn` is the door the seal opened deliberately; the other three predate it
+/// and hand out a node just as completely, so the census covers them or it is
+/// not a census. [`escape_surface_is_closed`] is what stops a fifth from
+/// appearing quietly.
+///
+/// | | hands out | why it exists |
+/// |---|---|---|
+/// | `as_syn` | the node itself | the escape |
+/// | `stripped_syntax` | `syn::Type` | the spelling under a transparent wrapper |
+/// | `to_syn` | `syn::Type` | the round-trip that checks the lowering |
+/// | `enum_item` | `&syn::ItemEnum` | a declared enum's own item |
+const ESCAPES: &[&str] = &["as_syn", "stripped_syntax", "to_syn", "enum_item"];
+
+/// The syn types a consumer can **take apart**. An ident, a lifetime or a member
+/// is a leaf: there is no shape to match on, so handing one out is not a door.
+const STRUCTURAL: &[&str] = &[
+    "Type",
+    "Expr",
+    "Item",
+    "ItemEnum",
+    "ItemStruct",
+    "ItemFn",
+    "ItemConst",
+    "Fields",
+    "Variant",
+    "Field",
+    "Path",
+    "Signature",
+    "FnArg",
+    "ReturnType",
+    "Pat",
+    "GenericArgument",
+];
+
+/// Which bucket an escape belongs to — by name, then by **receiver**.
 ///
 /// Outside `flat/`, a receiver named `origin` is always a captured item's node —
 /// an `ItemFn`, an `ItemStruct`, a `Field`, a `Variant` — because `TypeRef`'s own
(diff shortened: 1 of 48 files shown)


prebindgen/src/api/core/flat/boundary.rs
```diff
@@ -304,10 +375,25 @@ fn count(stream: TokenStream, aliases: &[String], n: &mut usize) {
             && is_sep(&toks, i + 1)
             && matches!(toks.get(i + 3), Some(TokenTree::Ident(_)))
         {
-            *n += 1;
+            n.classify += 1;
             i += 4;
             continue;
         }
+        // `<receiver> . as_syn ( )` — the escape, bucketed by receiver.
+        if matches!(&toks[i], TokenTree::Ident(id) if *id == "as_syn")
```
milyin: The escape census can be bypassed with UFCS. `TypeRef::as_syn(&ty)` and `Origin::as_syn(&origin)` produce the same node, but this branch requires a preceding `.` and an empty argument group. A consumer can also bind `let read = TypeRef::as_syn;` and call it, adding a new source-syntax escape with no ledger drift. Since the new invariant says `as_syn` is the one counted door, please count UFCS/function-item references too and add scanner tests for them (or reshape the API/check so every valid spelling is covered).

The all-features library suite passes locally (643 tests), and all CI checks pass.

— Codex (GPT-5)

prebindgen/src/api/core/flat/boundary.rs
```diff
@@ -431,68 +636,277 @@ fn scanner_recognizes_the_shapes_that_matter() {
...
+                    if STRUCTURAL.contains(&id.to_string().as_str())))
+                || matches!(&toks[i], TokenTree::Group(g) if names_a_node(g.stream()))
+        })
+    }
+    for item in items {
+        match item {
+            syn::Item::Impl(im) => {
+                for it in &im.items {
+                    if let syn::ImplItem::Fn(f) = it {
+                        let is_method =
```
milyin: This still leaves an uncounted door because the receiver check treats every associated function as harmless. An associated function can take the model explicitly:

```rust
impl TypeRef {
    pub fn leak(this: &Self) -> &syn::Type { this.as_syn() }
}
```

Outside `flat`, `TypeRef::leak(&ty)` exposes the held node, but `collect_doors` skips it because the first argument is `FnArg::Typed`, and the escape scan does not count `leak`. Passing that result to an existing classifier/helper adds neither an escape nor a classification-site ledger change. Please cover associated functions that accept/borrow a model value (with a regression test), or otherwise reject this API shape.

The same closed-surface claim also rests on the manually exhaustive `STRUCTURAL` list; for example, a receiver method returning `&[syn::Attribute]` is currently invisible because `Attribute` is absent. Inverting this to flag every explicit `syn::` return except a small allowlist of intentional leaf types would make the guard materially harder to bypass.

Validation at `29daeae`: 644 all-feature library tests and 22 doctests pass; formatting and all CI jobs pass.

— Codex (GPT-5)

prebindgen/src/api/core/flat/boundary.rs
```diff
@@ -431,68 +635,349 @@ fn scanner_recognizes_the_shapes_that_matter() {
...
+                && matches!(toks.get(i + 3), Some(TokenTree::Ident(id))
+                    if !LEAF.contains(&id.to_string().as_str())))
+                || matches!(&toks[i], TokenTree::Group(g) if names_a_node(g.stream()))
+        })
+    }
+    /// A function the caller could have written itself: no receiver, and a
+    /// non-leaf node already among its inputs.
+    fn is_transformer(sig: &syn::Signature) -> bool {
+        let has_receiver = matches!(sig.inputs.first(), Some(syn::FnArg::Receiver(_)));
+        let takes_a_node = sig.inputs.iter().any(|arg| match arg {
```
milyin: The transformer exemption is still bypassable because `takes_a_node` is existential and does not establish where the return came from:

```rust
pub fn leak<'a>(
    model: &'a TypeRef,
    _decoy: &syn::Type,
) -> &'a syn::Type {
    model.as_syn()
}
```

This has no receiver and takes a non-leaf node, so `is_transformer` returns true and `collect_doors` drops it; `leak` is not in `ESCAPES`, so a consumer can again obtain the model node with no ledger drift. The premise in the comment—that everything returned was already reachable from the node input—does not follow when other inputs can carry model state. Since there are only three intended transformers, an explicit transformer allowlist (or an equally strict proof that no other input can carry model state) would keep the exception on the safe side; please pin the mixed-input case as a regression.

There is a second spelling bypass in `names_a_node`: it recognizes only literal `syn::T`. `use syn::Type as SynType; impl S { pub fn leak(&self) -> &SynType { ... } }` is invisible to the surface guard. Please add alias handling/regression coverage before relying on the claim that every non-leaf `syn` return is a door.

Validation at `dded28d`: 645 all-feature library tests, 22 doctests, focused boundary tests, formatting, and all CI jobs pass.

— Codex (GPT-5)

prebindgen/src/api/core/flat/boundary.rs
```diff
@@ -431,68 +673,415 @@ fn scanner_recognizes_the_shapes_that_matter() {
...
+                        }
+                    }
+                }
+            }
+            syn::Item::Mod(m) => {
+                if let Some((_, items)) = &m.content {
+                    collect_doors(items, at, aliases, out);
+                }
+            }
+            _ => {}
```
milyin, line 1058: The direct item alias is covered now, but the surface proof still has naming and API-kind holes.

First, `syn_imports` does not see a crate alias or a local type alias:

```rust
use syn as syntax;
type Node = syntax::Type;
impl S { pub fn leak(&self) -> &Node { ... } }
```

The flattened run for `use syn as syntax` contains no uppercase type, and the return names only `Node`, so `hands_out_a_node` is false. This is ordinary Rust naming rather than a convention-breaking lowercase `syn` type. Please either resolve these aliases or reject such alias forms inside `flat`, with regressions.

Second, `collect_doors` only visits `Item::Fn` and `ImplItem::Fn`. A public field can reopen the exact capability this PR closes (`pub node: syn::Type` on a public model struct), and a public trait method plus its inherited-visibility impl is also invisible:

```rust
pub trait Leak { fn leak(&self) -> &syn::Type; }
impl Leak for TypeRef { fn leak(&self) -> &syn::Type { self.as_syn() } }
```

Neither shape reaches `is_door`, and calls/field access use names outside `ESCAPES`. The closed-surface guard should cover externally visible structural fields and trait methods too, or narrow its stated guarantee accordingly.

Validation at `7d733b2`: 645 all-feature library tests, 22 doctests, focused boundary tests, formatting, and all CI jobs pass.

— Codex (GPT-5)

prebindgen/src/api/core/flat/boundary.rs
```diff
@@ -431,68 +696,561 @@ fn scanner_recognizes_the_shapes_that_matter() {
...
+        if is_cfg_test_item(&item_attrs(item)) {
+            continue;
+        }
+        match item {
+            syn::Item::Fn(f) if is_door(&f.vis, &f.sig) => {
+                out.push((f.sig.ident.to_string(), at.to_string()));
+            }
+            // A crate rename or a local alias puts a node behind a name this
+            // check cannot follow, so the name itself is reported.
+            syn::Item::Type(t)
```
milyin, line 992: The alias arm still leaves two ordinary alias forms unguarded.

First, it rejects only aliases whose own visibility escapes `flat`. A private alias can still hide a node in a public signature:

```rust
type Node = syn::Type;
impl S { pub fn leak(&self) -> &Node { ... } }
```

`Item::Type` is skipped because `Node` is inherited/private, and the method return contains no recognized node name. The alias is transparent, so callers do not need to name it; I verified the equivalent minimal Rust pattern compiles and its public result can be consumed as the underlying public type. Since the chosen policy is to report rather than resolve aliases, every non-leaf local alias needs reporting regardless of alias visibility (test-only exclusions aside).

Second, associated type aliases are not visited. A concrete escape is:

```rust
impl std::ops::Deref for TypeRef {
    type Target = syn::Type;
    fn deref(&self) -> &Self::Target { self.as_syn() }
}
```

The impl method is checked, but `&Self::Target` names no node; `ImplItem::Type` is ignored, so consumers obtain the node through `&*ty` with no escape name or ledger drift. The same applies to `TraitItem::Type` defaults/bindings. Please cover non-leaf `TraitItem::Type` and `ImplItem::Type` aliases and pin the Deref-shaped case.

Validation at `834cbf9`: 645 all-feature library tests, 22 doctests, focused boundary tests, formatting, and all CI jobs pass.

— Codex (GPT-5)

prebindgen/src/api/core/flat/boundary.rs
```diff
@@ -431,68 +699,617 @@ fn scanner_recognizes_the_shapes_that_matter() {
...
+                for v in &en.variants {
+                    fields_of(&v.fields, &en.ident, true, out);
+                }
+            }
+            syn::Item::Mod(m) => {
+                if let Some((_, items)) = &m.content {
+                    collect_doors(items, at, aliases, out);
+                }
+            }
+            _ => {}
```
milyin, line 1058: The alias and associated-type cases are covered, but this fallback still leaves a pre-expansion macro hole. `collect_doors` ignores `Item::Macro`, plus `ImplItem::Macro` and `TraitItem::Macro` inside the arms above. For example:

```rust
macro_rules! leak {
    () => {
        impl TypeRef {
            pub fn leak(&self) -> &syn::Type { self.as_syn() }
        }
    };
}
leak!();
```

The walker sees only macro items, while code outside `flat` can call `ty.leak()` without writing any name from `ESCAPES`, so neither side of the ledger changes. `include!` can produce the same result. Because `flat` currently has no production item-generating macros, please fail closed on item macros (including impl/trait item macros), or inspect their expansion, and pin at least top-level and impl-item regressions.

I verified the focused boundary tests, all 645 library tests, all 22 doctests, formatting, diff whitespace, and the current CI checks; they are otherwise clean.

— Codex (GPT-5)

milyin: Follow-up after checking this against the actual tree: this is a completeness limitation of the prospective meta-test, not a defect present in the PR or current model. Production `core/flat` contains no `macro_rules!`, `include!`, or other item-generating macro source, and this PR introduces none. I therefore do **not** consider this merge-blocking and would not require macro rejection/expansion machinery in #313. Fail-closed handling would be optional future hardening only if the guard is intended to prove the invariant across API-generation patterns that the repository does not use.

— Codex (GPT-5)

PR #314, opened 2026-08-04
Umbrella: the model answers every source question (finishes #211's L5)

milyin:
Umbrella for finishing what [#229](https://github.com/milyin/prebindgen/pull/229) started: **every source fact comes from the model, and nothing outside `core::flat` takes captured Rust apart.**

#229 built the model and made the components consume it. It closed its own stages and left L5's last two bullets open — "extend the ledger's `WATCHED`" and "close or accept the blind spots". This umbrella owns those, and the work they turned out to name.

## What is being done

`core::flat` parses captured Rust once into elements: a closed model paired with the syntax each node was built from. The rule for every consumer is

> **Classify off `kind`, spell with `spell()`.**

The remaining work is to make that rule *true* rather than *asked for*. Three things stand in the way, and they are different problems:

1. **The model has to be able to answer.** A consumer that reaches for the syntax because the classification lost something is not the consumer's fault. Largely done — `TypeKind` is now the accepted subset of `syn::Type`, lifetimes, wrappers and generic arguments included, and `TypeKind::to_syn()` round-trips it against the source tokens.
2. **The syntax has to stop being handed out by default.** Done in #313: `Origin`'s syntax is private, `spell()` yields tokens, `as_syn()` yields the node and says so.
3. **The consumers have to be migrated.** Not done, and this is the long pole: 122 places still take a node.

## How it is measured

`prebindgen/src/api/core/flat/boundary.ledger` — three committed counts, regenerated with `UPDATE_BOUNDARY_LEDGER=1 cargo test -p prebindgen boundary_ledger`. A count that moves fails the build, so **the diff is the decision** and a migration cannot land invisibly.

| Section | Now | Done when | What it counts |
|---|---:|---|---|
| classification sites | 116 → **49** | **irreducible — accounted for, S45** | a variant of `syn::Type`/`syn::Expr` named outside `core::flat` — what a consumer visibly *does* with a node |
| escapes: types | 122 → **0** ✅ | **zero — met, S38** | a route from the model to a type node — what a consumer *can* do |
| escapes: items | 43 → **4** ✅ | **at its floor — accounted for, S46** | a captured item's own node: an emitter re-stating a whole item legitimately needs one — but *reading a fact off* one is a missing accessor, which is how S18 took nine of them |

**`escapes: types` is zero as of S38 — the goal is met.** Every consumer outside
`core::flat` reads `kind`, spells `spell()`, or looks up a `TypeKey`; there is no
longer a route from the model to a type node anywhere in the crate. The ledger
header now says the count must *stay* there, and what a new row would mean.

`api/core` got there at S33 — S8 claimed it and left seven, in `order.rs`,
`expand.rs`, `key.rs`, `diagnostics.rs` and `scan.rs`, which is itself a result:
a claim the ledger did not check is a claim that drifts.

**`classification sites` (49) is also at its floor, and that floor is enforced.**
Every remaining site reads a wire this adapter composed, a build-script
declaration, a generated converter's own signature, or the array-length
whitelist — never source syntax. S45 made that an accounted `FLOOR` table with a
test that fails on an unaccounted file *and* on a stale entry, so a new
classifier has to declare which population it joins.

**`escapes: items` (4) is at its floor too, as of S46**, and accounted the same
way. There is exactly one legitimate reason to take an item node — re-stating a
whole captured item verbatim, tokens and all — so `ITEM_FLOOR` is a list of
reasons rather than an enum, and taking an item to read a *fact* off it is a
missing accessor by definition. The four: `const_path_alias` re-emitting a
captured `const` (×2, both adapters), a `Guard`'s `const _`, and the C mirror
re-stating an `EnumValue`'s discriminant *as written*.

S46 found two that did **not** qualify, hiding inside a population the header
had called legitimate without saying which members were: `registry/scan.rs`
walked `sig.inputs` / `sig.output` and cloned a whole `syn::ItemConst` to read
`.ty`. Both read the element now, and `intern_recursive` — the node door into
the type table — is deleted for want of a caller.

## Where this stands

| count | at its floor | floor enforced |
|---|---|---|
| `escapes: types` = 0 | ✅ S38 | ✅ ledger ratchet |
| `classification sites` = 49 | ✅ | ✅ `FLOOR`, S45 |
| `escapes: items` = 4 | ✅ S46 | ✅ `ITEM_FLOOR`, S46 |

Every count is at its floor, and **every floor is enforced rather than
asserted** — which is the distinction this transition kept learning the hard
way. S8 claimed `api/core` was clean and it held seven escapes for twenty-five
stages. S38's header edit was reverted by the next regeneration, because the
header is generated from `boundary.rs` and I edited the output. S46's own two
sites hid inside a population described as legitimate without naming its
members. Each time the fix was the same: make the claim something the build
checks.

(Counts as of S46, all merged. The type count went 122 → 126 → 114 → 107 → 91 → 82 → 74 → 69 → 67 → 66 → 65 → 58 → 56 → 55 → 56 → 52 → 51 → 47 → 41 → 35 → 31 → 26 → 23 → 18 → 11 → 7 → 4 → 2 → 1 → 0: S2 revealed a population before the stages after it removed that and more.)

A pattern worth naming, since three stages have now hit it: **a site that asked
`syn` often did so because the old `TypeKind` could not answer.** S3's name
builders, S6's `type_kind`, S7's form predicates — each carried a comment
explaining why the spelling was necessary, and each of those comments was
describing the classification that S0 replaced. Reading them is how the
remaining population gets triaged.

From S10 on, the **classification** count is the one moving: deleting a helper
that took a node (`option_inner_ref_mutability`, `is_string_type`,
`peel_leading_ref`) pays more than the one escape it was reached through, because
each such helper was several watched sites.

**The item bucket can fall too**, which S11 found: it was expected to persist
until items grow modelled accessors, and the other way it drops is an item read
that existed only because a *type* question was being asked of the signature.

**A duplicate classifier is worth more than its count.** S17 found two
implementations of enum discriminant numbering — the model's and a copy in
`types_util` that the adapters used — which is the #210 shape the ledger exists
to catch, living inside `api/core` where the ledger does not look. It cost two
classification sites to delete and removed a way for the two answers to
disagree; the copy had already got the overflow case wrong.

**What "irreducible" is starting to look like.** `trait_impl.rs`'s five remaining
classification sites all match on `produced` — the spelling the converter
selector *composes* and hands in. Nothing captured is being classified there, and
no reading exists to ask instead. That is the shape the classification count is
converging on, and the point at which an entry stops being owed.

The second is the real measure, and it exists because the first can only measure. Since the model's syntax is sealed, a file with **no escapes cannot classify source syntax however it is written** — which is what widening `WATCHED` was trying to approximate, and it covers the blind spots that had no watched site to widen (an ident compared by name, a helper handed the node, a syn enum nobody listed).

### The escape count is not 122 classifiers

Three populations share it, and only the first is what the ledger's top section already sees:

* **taking a node apart** — the migration proper: ask the model instead;
* **keying by spelling** — `TypeKey::from_type(t.as_syn())` where `t.key()` is the answer. Identity, not parsing;
* **carrying a spelling** into an adapter-owned `syn::Type` field — `ConverterImpl::subs` / `produced`, which is L5's own bullet.

So the number over-counts classification on purpose. Over-counting the bucket that must reach zero is the safe direction; each population is a follow-up the count names rather than hides.

### A rising count can be the right answer

A field typed `syn::Type` is a **node cache the census cannot see**: it costs one
escape where it is filled and none at all where it is read, so N consumers read a
node for free and the ledger counts 1. Replacing that field with a `TypeRef`
removes the one and reveals the N — and those reads were always happening.

S2 is the first stage where that shows: `api/core` drops 15 escapes, jnigen gains
24, and every revealed read is a name-building classifier that was hiding behind
a plan field. The rule, now stated in the ledger header:

> A rising count **after a field is de-cached** is the measure becoming true. A
> rising count with **no field removed** is the thing this ledger exists to catch.

So "escapes: types → zero" is the goal, and the path to it is not monotone.

S3 is the other half of that: the 12 reads S2 revealed in `iface.rs` were one
population — three name-building functions — and taking them off the spelling
dropped **both** counts (escapes 126 → 114, classification 116 → 110) plus a row
from jnigen's own spelling census. Reveal, then remove.

### Two guards, because a count is only as good as its door

* `escapes_are_counted_by_their_receiver` — the census counts the **name**, not a call shape, so UFCS (`TypeRef::as_syn(&ty)`) and a function item (`let read = TypeRef::as_syn;`) count like a method call.
* `escape_surface_is_closed` — reads the model's own surface and fails if anything externally visible hands out a non-leaf `syn` node under a name the census does not count. Functions, public fields, trait methods and their impls, and aliases (`use syn as syntax`, `type Node = syn::Type`, `type Target = syn::Type`) alike.

Both allowlists sit on the **exception** side — `LEAF` says what is not a node, `TRANSFORMERS` says who is excused from being a door — so being wrong raises a false alarm instead of leaving a silent hole. Four of the five doors the census counts were found by the guard rather than by design.

### And what it does not measure

Token-string classification. `spell().to_string()` reaches a string, and the one live instance (`core/domain.rs` matching `"i8" | "f32"`) reads a `syn::Type` a **build script** supplied, which was never the model's. The ledger header says so rather than implying it away.

## Stages

| Stage | Owns | State |
|---|---|---|
| S0 | `TypeKind` is the accepted syntax subset; the readings replace the folds | **done** — [#312](https://github.com/milyin/prebindgen/pull/312), in `main` |
| S1 | Seal the syntax; count the doors | **done** — [#313](https://github.com/milyin/prebindgen/pull/313) |
| S2 | A declaration carries identity, a plan carries a reading — the `syn::Type` fields in `api/core`'s plans | **done** — [#315](https://github.com/milyin/prebindgen/pull/315) |
| S3 | The name builders `subject_short` / `subject_package` / `whole_value_name` read the kind — the 12 reads S2 revealed | **done** — [#316](https://github.com/milyin/prebindgen/pull/316) |
| S4 | `ConverterImpl::subs` → `Vec<TypeKey>`: they were identities, and the resolver keyed them on arrival | **done** — [#317](https://github.com/milyin/prebindgen/pull/317) |
| S5 | A key is asked for, not built — `x.key()`, `CtorSig::target`, `SpecKey::whole_folder` | **done** — [#318](https://github.com/milyin/prebindgen/pull/318) |
| S6 | jnigen's own carriers: `FlatFieldNode::rust_ty`, and `type_kind` taking a `TypeKey` | **done** — [#319](https://github.com/milyin/prebindgen/pull/319) |
| S7 | A question about the form is a question about the kind — the predicates that asked `syn` because the old `TypeKind` could not answer | **done** — [#320](https://github.com/milyin/prebindgen/pull/320) |
| S8 | `api/core` reaches zero: `Layered::core`, `accessor_signature`, a peeled core in `iface` | **done** — [#321](https://github.com/milyin/prebindgen/pull/321) — but it did **not** reach zero; see S33 |
| S9 | The leaf-vec element list is a set of identities, and its producer reads the model | **done** — [#322](https://github.com/milyin/prebindgen/pull/322) |
| S10 | `PlanParam::ty` carries its reading; the wrapper's mutability probes read the kind | **done** — [#323](https://github.com/milyin/prebindgen/pull/323) |
| S11 | The sum-position checks read `ret.fallible_parts()` instead of `sig.output` — the first fall in the **item** bucket | **done** — [#324](https://github.com/milyin/prebindgen/pull/324) |
| S12 | The slice-of-sums refusal reads the model — the twin walk #322 left behind | **done** — [#325](https://github.com/milyin/prebindgen/pull/325) |
| S13 | A lookup takes the reading, a reconstruction takes the identity — `leaf_ty_is_prim`, `sum_reconstruct` | **done** — [#326](https://github.com/milyin/prebindgen/pull/326) |
| S14 | cbindgen's naming family takes a `TypeKey` — every emitted C symbol | **done** — [#327](https://github.com/milyin/prebindgen/pull/327) |
| S15 | `struct_fields` reads the `Struct` element instead of its item | **done** — [#328](https://github.com/milyin/prebindgen/pull/328) |
| S16 | The scan interns readings and takes elements — `registry/scan.rs` halves both buckets | **done** — [#329](https://github.com/milyin/prebindgen/pull/329) |
| S17 | One discriminant numbering, and the element kind replaces `assert_only_unit_variants` | [#330](https://github.com/milyin/prebindgen/pull/330) |
| S18 | An element answers for its own docs — the item bucket falls by a quarter | **done** — [#331](https://github.com/milyin/prebindgen/pull/331) |
| S19 | cbindgen's unit-enum lookup is the shape check | **done** — [#332](https://github.com/milyin/prebindgen/pull/332) |
| S20 | A signature is a parameter list — `overloads.rs`'s three `sig.inputs` walks | **done** — [#333](https://github.com/milyin/prebindgen/pull/333) |
| S21 | `prim_array_of` reads the extent — the original map's own L4 bullet, outliving the program that named it | **done** — [#334](https://github.com/milyin/prebindgen/pull/334) |
| S22 | The plan keeps the reading it was handed — `struct_plan`'s three derived spellings | **done** — [#335](https://github.com/milyin/prebindgen/pull/335) |
| S23 | The constant refusal reads the element; one peel serves both paths | **done** — [#336](https://github.com/milyin/prebindgen/pull/336) |
| S24 | The string audit reads `alternatives` — `enum_variants` becomes `enum_alternatives(&TypeKey)`, and the `enum_item` door leaves `cbindgen/emit.rs` | **done** — [#337](https://github.com/milyin/prebindgen/pull/337) |
| S25 | cbindgen's field policies take readings — the whole per-field chain (`struct_fields`, `mirror_field_wire`, `payload_*`, the ownership audit); **then** the three tagged-union walks read `alternatives`, `payload_enum` replaces `enum_item` + `assert_payload_enum`, and `Alternative::spell` + `Field::bind` replace `variant_pattern` / `variant_ctor`. Deletes `types_util::enum_shape`, the last `syn::Fields` classifier | **done** — [#338](https://github.com/milyin/prebindgen/pull/338) |
| S26 | The JNI wrapper shapes take a reading — the five `t1_ty = t1.as_syn()` chains, and the `option_input` / `option_output` / `nullable_kind_for*` / `reject_vec_of_handle` / `build_fn_type` / `opaque_leaf_meta` helpers they feed | **done** — [#339](https://github.com/milyin/prebindgen/pull/339) |
| S27 | A lookup key is not a spelling — both transparent bridges' `stripped_syntax` → `stripped_key`; `selector.rs`'s `output_slice`, `&[T]` element, and `fallible_parts` (whose syntactic fallback becomes unreachable by construction) | **done** — [#340](https://github.com/milyin/prebindgen/pull/340) |
| S28 | An identity is a key, taken off the element — `SpecKey::callback`, `peel_receiver_key` (four classification sites), and the `class_members` signature walk they gated | **done** — [#341](https://github.com/milyin/prebindgen/pull/341) |
| S29 | The acceptance check the model already made — the array-length whitelist was six `syn::Expr` forms wider than `lower_array_len` accepts; `validate_constant_fn` reads the `Function` element | **done** — [#342](https://github.com/milyin/prebindgen/pull/342) |
| S30 | The source path of a declared type is its identity — `src_ty_of(&TypeKey)`, the three declared-type emission loops, both `convert.rs` targets, and the three value-opaque helpers they gated | **done** — [#343](https://github.com/milyin/prebindgen/pull/343) |
| S31 | The interface leaf reads — `leaf_iface_param` → `unfold_leaf_kt` → `enum_probe_type`, and `LeafDesc::Whole::ty`. Deletes the spelling twin its own doc said was waiting on exactly these two fields | **done** — [#344](https://github.com/milyin/prebindgen/pull/344) |
| S32 | What a Rust arm erases to, and one more node cache — `overloads.rs`'s `rust_type_erased` / `arm_erased_sig`, and `flat_input.rs`'s `direct_handle` field | **done** — [#345](https://github.com/milyin/prebindgen/pull/345) |
| S33 | `api/core` reaches zero **for real** — `plan_edges` returns crossings, `constructed_value` returns the reading, `type_from_ident` loses its last caller, `scan.rs` canonicalizes off the declaration's key | **done** — [#346](https://github.com/milyin/prebindgen/pull/346) |
| S35 | The return target and the declared payload are identities — `fn_plan`'s three candidates are all readings; `declared_opaque_payload_inner` returns the key it computed | **done** — [#348](https://github.com/milyin/prebindgen/pull/348) |
| S36 | `produced` stops being a spelling — `Produced::{Reading, Composed}`, so `bridge_layers` reads `erased_wrappers` and the `canonical` argument disappears from four helpers and five locals | **done** — [#349](https://github.com/milyin/prebindgen/pull/349) |
| S37 | The declare-phase read is an identity too — `build_sum_decons` off `Origin::key().ident()`. **`jnigen` and `api/core` are both zero** | **done** — [#350](https://github.com/milyin/prebindgen/pull/350) |
| S39 | The C wrapper emitter reads the element — `emit_function_wrapper` / `emit_inputs` / `alias_preflight` off `Function::params` and `ret`, and `alias_slot` off `TypeKind` | **done** — [#352](https://github.com/milyin/prebindgen/pull/352) |
| S46 | The scan reads, and the item floor is accounted too — `scan_fn_signature` / the const scan off the element, `intern_recursive` deleted, and `ITEM_FLOOR` + its test. **All three counts at their floor, all three floors enforced** | **done** — [#359](https://github.com/milyin/prebindgen/pull/359) |
| S45 | The classification floor is accounted for, and **checked** — `FLOOR` names per file which of four non-source populations its sites read (Wire / Declaration / GeneratedSignature / AcceptedLength), and a test fails both on an unaccounted file and on a stale entry. Also corrects S38: the header text belongs in `boundary.rs`'s `HEADER`, not the generated `.ledger` | **done** — [#358](https://github.com/milyin/prebindgen/pull/358) |
| S44 | The return surface reads — `ReturnSurface::classify` / `classify_return` off the reading; `annotate_borrow_with_lifetime` deleted (it was reached only from the composed-type door, where nothing is a borrow). **jnigen's classification count is irreducible: wires and build-script declarations only** | **done** — [#357](https://github.com/milyin/prebindgen/pull/357) |
| S43 | A conversion function is a signature — `one_param` / `convert_single_param_any` / `fn_ret` off the `Function` element, in **both** adapters; `result_ok_type` + `result_err_type` → `fallible_parts` | **done** — [#356](https://github.com/milyin/prebindgen/pull/356) |
| S42 | The JNI symbol checks read — `symbols.rs`'s three name lookups off `Alternative`, `Field` and `EnumValue`; `kotlin_emit`'s fn ident. **`enum_item` loses its last production caller** | **done** — [#355](https://github.com/milyin/prebindgen/pull/355) |
| S41 | The source path recurses on readings — `src_ty_deep` off `TypeKind::Ref` / `Slice`, and `emit_inputs`' slice arms take the parameter's reading | **done** — [#354](https://github.com/milyin/prebindgen/pull/354) |
| S40 | The C shape lowering reads — `lower_shape` / `encode_value` / `output_is_fallible` recurse on readings, and `value_ty` comes off `fallible_parts`. Deletes `is_option_type`, `is_vec_type`, `is_unit`, `cow_slice_elem` | **done** — [#353](https://github.com/milyin/prebindgen/pull/353) |
| S38 | **`escapes: types` reaches ZERO.** cbindgen's selector chain takes readings — all thirteen arms, and `convert_crossing` with them. Every substitution was a peer an earlier stage had already added | **done** — [#351](https://github.com/milyin/prebindgen/pull/351) |
| S34 | The JNI terminals and bridges read — `build_input_fn` / `build_output_fn` split into a tokens core plus a reading twin, and everything they gated: `lookup_*`, `opaque_handle_*`, `override_kotlin_name`, `primitive_*`, `cow_bytes_output` | **done** — [#347](https://github.com/milyin/prebindgen/pull/347) |
| S34 | `produced` becomes a reading — `let syntax = ty.as_syn()` in both jnigen selectors, feeding `input_wrapper_shape` / `output_wrapper_shape`, and cbindgen's twin in `convert_with`. Blocked on the one arm that composes a `Vec<T>` the adapter cannot mint a reading for (#280), and on `bridge_layers` / `build_from_canonical` / `is_canonical_spelling` | not started |
| S29 | jnigen's `input_terminal` / `output_terminal` — one `let ty = reading.as_syn()` each, feeding ~15 downstream calls (`opaque_handle_input`, `build_input_fn`, `override_kotlin_name`) | not started |
| S16 | cbindgen's field consumers take readings — `is_string`, `data_field_wire`, `payload_field_wire`; then `struct_fields` returns `(Ident, TypeRef)` and `src_ty` follows | not started |
| S17 | The adapters' remainder, now all **chain** migrations — a helper takes a node, so its callers must: `trait_impl`'s converter locals (14), `iface`'s `leaf_iface_param` / `unfold_leaf_kt` (7), `selector.rs` (6), cbindgen's naming chain (6) | not started |

| S19 | `Niches { value: syn::Expr, matches: syn::Expr }`; `post_process_item(&mut syn::Item)` | not started |
| S4b | The classifiers that remain: read the model, or state why the site is adapter-synthesized. `leaf_iface_param`, `selector.rs`, `emit/flat_input.rs` are the dense ones | not started |
| S5 | Item accessors — doc attributes, `scan_fn_signature` — so the item bucket can shrink too | not started |

S2 was scoped as "keying stops being a spelling operation" and that turned
out to be 7 sites; keying folded into it for free instead. What the stage
actually removed is the `syn::Type` **fields**, and that is where the rest of
the program's shape came from — see the node-cache note below.

## Review protocol

Each stage PR states its own exit, and every one of them so far has had the same first line:

* **Must not move** — every generated artifact byte-identical (`examples/regen-check.sh`). A behaviour-preserving migration that changes an artifact is a bug, not a surprise.
* **The ledger diff is the claim** — which counts moved, and why.
* Gate: `cargo test -p prebindgen --all-features`, clippy + `cargo fmt --check` on **both** 1.85.0 and stable, then the regen check.


docs/language-integration.md
```diff
@@ -627,11 +627,37 @@ growing back.
...
+      `enum_item`, `type_from_ident`) by NAME rather than by call shape, so UFCS
+      and a function item count like a method call — and
+      `escape_surface_is_closed` reads the model's own surface so a sixth cannot
+      appear quietly, which is how four of the five were found. It asks the
+      question the safe way round — every `syn` return is a door unless its type
+      is a leaf (`Ident`, `Lifetime`, `Member`, `Index`) and unless the function
+      is one of three named transformers — and it reads more than functions,
+      because more than a function can hand out a node: a public field, a trait
+      method and its impl, and an alias that hides the name
+- [ ] Drive **escapes: types** to zero. The population is three kinds, and only
```
milyin, line 649: The final umbrella still commits the S1-era status here: the preceding paragraph reports 115 type / 28 item escapes, and this line leaves the type-zero objective unchecked. At this head the enforced ledger says 49 classification sites, **0** type escapes, and **4** item escapes, with both remaining floors accounted. Since #314 presents the migration as complete, please update this section to the final counts/status (or label the old figures explicitly as a historical S1 snapshot); otherwise the checked-in integration guide contradicts both the PR and `boundary.ledger` about whether the main objective is finished.

Validation at `60600da`: 640 all-feature library tests and 22 doctests pass; in-repo generated artifacts are byte-identical; diff whitespace and all current CI jobs are clean.

— Codex (GPT-5)

PR #360, opened 2026-08-04
C1: the emission callbacks carry a capability

milyin:
First stage of replacing the boundary census with a **capability**. Follows umbrella #314, which reached its floor by counting; this changes the mechanism.

## Why not keep counting

Three holes, and the first two are not fixable by a better census:

1. **It is already walked around.** Six sites launder a reading back into a `syn::Type` via `spell()` → `parse_quote!` — `cbindgen/mod.rs:568` `spelled()`, `jni/builder.rs:1753` `spelled_ty()` (15 callers between them), and four others. None names a counted door, so `escapes: types = 0` is true of the *names*, not of the capability.
2. **It cannot see out-of-crate adapters.** `scan_tree` walks this crate's `src/`. `flat` is public API and `lib.rs` invites third-party adapters; for those the boundary is enforced by nothing at all.
3. **It costs ~1,940 lines** — `boundary.rs` 1510, `boundary.ledger` 163, jnigen's `spelling_census` 264, plus two hand-edited floor tables.

## The mechanism

```rust
pub struct Emit { _seal: () }

impl Emit {
    pub(in crate::api::core) fn new() -> Self { … }   // the whole enforcement
}
```

Private field, no public constructor: the only way to hold one is to be handed one. `write_rust` mints exactly one and hands a borrow to the callbacks whose job is producing Rust — `prerequisites`, the five `on_*`, `post_process_item`. Nothing else in the pipeline gets one, so adapter code that classifies, plans, names or validates cannot compile a call to a door — **in this crate and in anyone else's**.

That is the property counting cannot have.

## What moved in this stage

The **item** doors only. Types come in C3.

| `Emit` method | replaces |
|---|---|
| `item` / `type_item` | `Element::as_syn`, `Type::as_syn` |
| `const_alias` / `const_verbatim` | `const_path_alias(c.origin.as_syn(), …)`, `c.origin.spell()` |
| `guard` | a `Guard`'s `const _`, spliced verbatim |
| `discriminant` | an `EnumValue`'s `= 0x07`, **as written** |

`const_path_alias` is sealed to `api::core` alongside them — it takes a `&syn::ItemConst` no adapter can obtain any more. `Emit::const_alias` takes the *element* instead, which is what handing over a whole item to read four fields off it was asking for.

## The census skips `emit.rs`

It already skips `api/core/flat/` because "the language is where classification is supposed to live". By the same rule it now skips `api/core/emit.rs`: the capability module holds every door **on purpose**, so counting it would report the design as drift. What the two escape sections still measure is *consumers* — and the point is that a consumer cannot reach a door.

## Measurement

| | before | after |
|---|---:|---:|
| classification sites | 49 | 49 |
| escapes: types | 0 | 0 |
| escapes: items | 4 | **0** |

```
api/core/prebindgen.rs            1 -> 0
api/core/write.rs                 1 -> 0
api/lang/cbindgen/trait_impl.rs   1 -> 0
api/lang/jnigen/jni/trait_impl.rs 1 -> 0
```

`ITEM_FLOOR` is now empty and its test passes vacuously — S46 built it two stages ago to account a population that this stage dissolves. Both it and `FLOOR` are deleted at the end of this arc, together with the ledger.

## Public API

`Emit` is exported as `prebindgen::core::Emit`: an out-of-crate adapter implements `Prebindgen`, so it receives one and must be able to name the type. That is also how the boundary reaches it — the seven callback signatures are a breaking change for any such adapter, and adding `_emit: &Emit` is the whole migration.

## Gate

- `cargo test -p prebindgen --all-features` — 640 lib, 22 doctests
- ledger regenerated, diff above
- clippy `--all-targets --all-features -D warnings` on stable **and** 1.85.0
- `examples/regen-check.sh` — every generated artifact byte-identical

### DIFF
--- a/prebindgen/src/api/core/emit.rs
+++ b/prebindgen/src/api/core/emit.rs
@@ -1,0 +1,187 @@
+//! [`Emit`] — the capability to render captured Rust syntax.
+//!
+//! # Why this exists
+//!
+//! The model pairs every element with the syntax it was built from, and
+//! generated Rust has to **spell** that syntax: a converter's signature says
+//! what the source said. But *reading* the same syntax to decide what a type
+//! means is the thing [#211](https://github.com/milyin/prebindgen/issues/211)
+//! removed — a decision belongs to `kind`, which cannot disagree with itself
+//! the way a spelling can.
+//!
+//! Those two are the same capability if the model simply hands syntax out, so
+//! for a long time the difference was kept by *counting*: a token census froze
+//! how many places named a door, and a moving count failed the build. That
+//! worked and had three holes. It could be walked around without moving
+//! (`spell()` → `parse_quote!` recovers a node while naming no counted door).
+//! It could not see an out-of-crate adapter at all, because it scanned this
+//! crate's `src/`. And it cost about two thousand lines to maintain.
+//!
+//! So the difference is a **capability** instead. Syntax is reachable only
+//! through this type, this type cannot be constructed outside `api::core`, and
+//! core hands one out only to the callbacks whose job is producing Rust.
+//! Adapter code that classifies, plans, names or validates never receives one,
+//! and a call to a door from there does not compile — in this crate and in
+//! anyone else's.
+//!
+//! # Where one comes from
+//!
+//! [`Prebindgen::on_function`](super::prebindgen::Prebindgen::on_function) and
+//! its four peers, [`prerequisites`](super::prebindgen::Prebindgen::prerequisites),
+//! [`post_process_item`](super::prebindgen::Prebindgen::post_process_item), and
+//! the closure [`RegistryBuilder::convert_with`](super::registry::RegistryBuilder::convert_with)
+//! calls. Nothing else. If a helper needs an `&Emit`, that is the helper saying
+//! it emits; if threading one to it feels wrong, it is probably deciding
+//! something and wants the model instead.
+//!
+//! # The residual
+//!
+//! [`Emit::spell`] yields a `TokenStream`, so emission code can re-parse it and
+//! take the node apart. That is deliberate — emission is where syntax belongs —
+//! and closing it would mean an emission IR for Rust, mirroring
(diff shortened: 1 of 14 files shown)


prebindgen/src/api/core/emit.rs
```diff
@@ -0,0 +1,187 @@
...
+        ty.stripped_syntax()
+    }
+
+    /// A captured item, verbatim — attributes, visibility and body included.
+    ///
+    /// The legitimate reason to reach for an item at all: an emitter re-stating
+    /// one as written. Reading a *fact* off an item is a missing accessor, and
+    /// the model is where it belongs.
+    pub fn item(&self, e: &Element) -> syn::Item {
+        e.as_syn()
```
milyin, line 175: `Emit` is not yet the capability boundary claimed here, because this delegates to an original door that remains public. At this head an out-of-crate adapter can still call `Element::as_syn()`, `Type::as_syn()`, or `function.origin.as_syn()` directly; it can also call `function.origin.spell()` and parse the tokens back into an `ItemFn`. I compiled those exact calls from a separate crate without ever naming or receiving `Emit` (and the new `IdentityExt` test itself ignores `_emit` and emits through `f.origin.spell()`).

That makes the new `escapes: items = 0` misleading: the census now skips `core/emit.rs`, but the public capability surface has not changed—only the four in-tree call sites moved. Please either seal the original item routes in this stage (for example, core-restrict `Element::as_syn` / `Type::as_syn` and the relevant `Origin` access, leaving type-level proxies for the later C3 migration) and add an out-of-crate compile-fail check, or retain the census/status as transitional until the stage that actually makes these calls fail. The module docs should also not yet say `convert_with` closures receive `Emit`; their current signature still receives only `(&Crossing, &Building)`.

Validation at `6de7999`: 640 all-feature library tests, formatting, diff whitespace, and all current CI jobs pass. The external no-`Emit` bypass crate also compiles successfully.

— Codex (GPT-5)

PR #361, opened 2026-08-04
Umbrella: the boundary becomes a capability, not a count

milyin:
Successor to [#314](https://github.com/milyin/prebindgen/pull/314). That umbrella moved every consumer off captured Rust syntax and onto the model, and *measured* it with a token census committed as `boundary.ledger`. This one replaces the measurement with something the compiler enforces, and deletes the machinery as it becomes redundant.

## Why the census cannot be the permanent answer

It did its job — 122 type escapes to 0 over 46 stages, artifacts byte-identical throughout. But as a *mechanism* it has three problems, and the first two are not fixable by a better census.

**It can be walked around without moving.** Six sites launder a reading back into a `syn::Type` via `spell()` → `parse_quote!` — `cbindgen/mod.rs` `spelled()`, `jni/builder.rs` `spelled_ty()` (15 callers between them), and four others. None names a counted door, so `escapes: types = 0` is true of the *names* and not of the capability. The ledger header already lists token-string classification as a blind spot it "cannot close"; this is its structural cousin, and it arrived during the migration itself, not from an adversary.

**It cannot see out-of-crate adapters at all.** `scan_tree` walks this crate's `src/`. `flat` is public API and `lib.rs` invites third-party adapters — for those, the boundary is enforced by nothing.

**It costs ~1,940 lines**: `boundary.rs` 1510, `boundary.ledger` 163, jnigen's `spelling_census` 264, plus two hand-edited floor tables that each need a same-commit entry.

## The mechanism

```rust
pub struct Emit { _seal: () }

impl Emit {
    pub(in crate::api::core) fn new() -> Self { … }   // the whole enforcement
}
```

Private field, no public constructor: the only way to hold one is to be handed one. Core mints one in `write_rust` and passes a borrow to the callbacks whose job is producing Rust. Everything else — classification, planning, naming, validation — never receives one, so a call to a door from there **does not compile**, in this crate and in anyone else's.

That is the property counting cannot have. It also makes the boundary self-documenting: a helper that needs an `&Emit` is saying it emits.

## Scope

Deliberately **not** an emission IR for Rust mirroring [`api::gen::kotlin`](https://github.com/milyin/prebindgen/tree/main/prebindgen/src/api/gen/kotlin) (declaration model → typed expression AST → renderer). Generated bodies stay `quote!`. What moves is *who can reach captured syntax*, not *how code is built*.

## Stages

| # | Owns | State |
|---|---|---|
| C1 | `Emit` exists; the **item** doors sealed behind it, proved by `compile_fail` doctests that build as their own crate | **done** — [#360](https://github.com/milyin/prebindgen/pull/360) |
| C2 | `Display for TypeRef` — 22 diagnostic sites stop spelling | **done** — [#362](https://github.com/milyin/prebindgen/pull/362) |
| C3 | The converter path carries the token. **18 of ~30 arms turned out not to need it**; found three scalar spellings that were identities, and a declare-time probe that wanted one | **done** — [#363](https://github.com/milyin/prebindgen/pull/363) |
| C3b | A plan carried a spelling so decisions could peel it — the #273 mistake, in a plan field. Removing it stopped the capability leaking into validation and Kotlin rendering | **done** — [#364](https://github.com/milyin/prebindgen/pull/364) |
| C4 | An overload collides by identity, not by spelling — the latent bug this umbrella named up front | **done** — [#365](https://github.com/milyin/prebindgen/pull/365) |
| C5 | The **type** half sealed. Found `synthetic_getter` minting a `flat::Function` inside the adapter, which is what dragged the capability into validation | **done** — [#366](https://github.com/milyin/prebindgen/pull/366) |
| C6 | Delete the census — 1932 lines, net −1945 | **done** — [#367](https://github.com/milyin/prebindgen/pull/367) |

## Where it ended

| | start | end | enforced by |
|---|---|---|---|
| routes to a captured **item** | 43 escapes historically, 4 at C1 | none | compiler |
| routes to a captured **type** | 122 historically, 0 by census at C1 | none | compiler |
| adapter `spell()` calls | 69 | **0** | `Emit` |
| enforcement machinery | 1932 lines | 8 `compile_fail` doctests | rustc |

Every route from the model to captured syntax runs through one type an adapter
must be **handed** and cannot construct. Emission callbacks receive it;
classification, planning, naming and validation do not, and a call from there
does not compile — in this crate and in anyone else's.

## What threading it found

The capability was not just a stricter boundary; it was a diagnostic. Each of
these was invisible to a census, because none of them names a counted door:

* **An overload comparison keyed on a spelling** (C4). `ErasedJvmType` is a
  `String` compared with `==`; two spellings of one type would not collide, and
  the JVM would reject the class. Latent only because parse-time normalization
  made spellings agree — a dependency it could not see.
* **A plan field that existed so decisions could peel it** (C3b).
  `option_inner_type` on a spelling is the #273 mistake, and it sat where every
  consumer of a plan inherited it. That is why threading leaked into validation.
* **An adapter minting a model element** (C5). `synthetic_getter` built a
  `flat::Function`, which means spelling — so validation and Kotlin rendering
  were asked for an emission capability to obtain a `Function` they only wanted
  to read. #280 says that construction is the model's.
* **Three wire policies spelling a scalar** (C3) to get what `ScalarKind`
  already names.
* **18 of ~30 converter arms needing nothing at all**, because #314 had already
  moved them onto `key()`. Their signatures now say so.

## What C1 found

The review of #360 is the argument for this umbrella in miniature. The first version of `Emit` delegated to doors that stayed `pub` — so the capability was a convention, the four in-tree call sites had merely moved, and `escapes: items = 0` measured the census skipping `emit.rs` rather than anything being closed. A reviewer compiled the bypass crate to show it.

**The census would never have caught that.** The fix was sealing, and the check is now a `compile_fail` doctest — the only thing in this repo that builds against the published API the way a real consumer does.

One nuance that fell out of it: `Origin::spell` could not be blanket-sealed. On `Origin<syn::ItemFn>` its tokens re-parse to the item, so it is the item door renamed; on `Origin<syn::Type>` it is a *declaration*'s own type, written by a build script and never captured, which #280 leaves the model no reading for. It is sealed on the blanket impl and re-opened narrowly as `declared_spelling()`, named for what it is.

## Review protocol

Unchanged from #314, and every stage has met it:

* **Must not move** — every generated artifact byte-identical (`examples/regen-check.sh`).
* **The ledger diff is the claim**, until C6 deletes it.
* Gate: `cargo test -p prebindgen --all-features`, clippy + `cargo fmt --check` on **both** 1.85.0 and stable, then the regen check.


prebindgen/src/api/core/emit.rs
```diff
@@ -0,0 +1,324 @@
...
+///
+/// ```compile_fail
+/// # use prebindgen::core::flat;
+/// fn leak(f: &flat::Function) -> proc_macro2::TokenStream { f.origin.spell() }
+/// ```
+///
+/// A type's spelling, sealed as of C5 — the last route, and the one the census
+/// could only ever *count*:
+///
+/// ```compile_fail
```
milyin, line 171: The type half is not sealed: this compile-fail set omits the most direct existing door, `TypeRef::as_syn()`, which is still `pub fn` and returns the captured `&syn::Type`. I compiled an external crate containing `fn leak(t: &TypeRef) -> &syn::Type { t.as_syn() }` successfully without an `Emit`.

There are two more live route classes the same external check exposes:

* `Struct::spell`, `Alternative::spell`, and `EnumValue::spell` remain public even though they read retained source delimiters; `Emit::{struct_shape,alternative_shape}` merely delegate to them. Production adapter code still calls these methods directly outside `Emit`, so the stated rule that a helper needing captured spelling must carry `&Emit` is not enforced.
* `Display for TypeRef` currently calls the sealed `self.spell()` verbatim. `syn::parse_str(&ty.to_string())` succeeds externally, recreating the exact spelling through a route the compiler permits. Calling that unsupported does not make the claimed capability property structural; either render a model-derived diagnostic form or narrow the claim accordingly.

Please seal the direct node/shape methods, route legitimate emission through `Emit`, and add compile-fail cases for them before deleting the census. At minimum `TypeRef::as_syn` is the exact door this umbrella says no longer exists.

Validation at `20f8e1a`: 631 all-feature library tests, 30 doctests (including the eight present compile-fail cases), formatting, diff whitespace, and all current CI jobs pass. The external bypass crate also compiles successfully.

— Codex (GPT-5)

PR #368, opened 2026-08-05
C7: seal what C5 claimed, and check it mechanically

milyin:
Addresses the review on [#361](https://github.com/milyin/prebindgen/pull/361). Every point was correct; the central one is bad enough to lead with.

## C5 claimed the type half was sealed. It was not.

`TypeRef::as_syn` — the single most direct door, the one this umbrella is named for — was still `pub` after C5, and C6 deleted the census on top of that claim. The reviewer compiled `fn leak(t: &TypeRef) -> &syn::Type { t.as_syn() }` from an external crate.

**The cause matters more than the fix.** At C5 I did verify from an external crate — but I wrote the check myself and tested the three methods I had just changed. I verified my diff, not my claim. That is exactly the S8 failure this project already recorded (`api/core holds zero type escapes`, true of nothing, unchecked for twenty-five stages), committed by me two stages after writing it up.

## Sealed, each confirmed `E0624` from a real external crate

| | |
|---|---|
| `TypeRef::as_syn` | the door C5 missed |
| `Flat::enum_item` | one of the five original doors |
| `{Struct,Alternative,EnumValue}::spell` | the delimiters the source wrote |
| `flat::spell` (the module) | so `fields` cannot be reached around them |
| `flat::type_from_ident` | **deleted** — a public `syn::Type` factory with no caller since S33, which the plan said to delete and I did not |

`Emit::shape` replaces the three shape methods, bounded by a `pub(in crate::api::core)` `Shaped` trait: an out-of-crate consumer can call `shape` on the three elements, and cannot name the trait to route around it.

## One thing the reviewer did not ask for, and I got wrong first

I initially sealed `Field::member` and `Field::bind` too. They read the field's `name` and `index` — model facts, no captured syntax — and sealing them dragged the capability back into Kotlin data-class rendering, which was the C3b/C5 symptom all over again. Reverted, with the reason written next to them.

## `Display` renders the identity

```rust
-write!(f, "{}", self.spell())
+write!(f, "{}", self.key().as_str())
```

As reported, delegating to `spell()` handed the captured spelling back out through `format!`, so `syn::parse_str(&ty.to_string())` reconstructed it exactly. It renders the canonical form now — readable in a `panic!`, and a round trip lands on a *normalized* type rather than the source's own tokens.

## The actual fix: `flat::surface`

C6 deleted the census's `escape_surface_is_closed` on my argument that visibility makes it unnecessary. **That argument was wrong**, and this PR is the proof: visibility enforces the methods someone remembered to seal, and cannot notice the one they missed.

Restored as a ~110-line test that reads `flat`'s own public surface and fails on anything returning a `syn` type — with the census's two hard-won rules kept, because both were learned from real bypasses:

* an inverted `LEAF` allowlist (a `syn` type nobody considered is a door until someone argues otherwise);
* `TRANSFORMERS` exempt by **name and shape** — exactly one parameter, itself a node — because a name alone is defeated by `fn leak(model: &TypeRef, _decoy: &syn::Type) -> &syn::Type`.

On its first run it found `enum_item`, `type_from_ident` and all three shape methods. Four of those I was not looking for.

## Copilot

Both applied: the `convert_with` rustdoc examples in `registry/{mod,declare}.rs` still showed the two-argument closure, and `emit.rs`'s module doc still said the converter path had not been threaded.

## Verification

The check I should have run at C5 — eleven bypasses, all now `E0624` or `E0599`:

```
4 error[E0624]: method `as_syn` is private
2 error[E0624]: method `spell` is private
1 error[E0624]: method `stripped_syntax` is private
1 error[E0624]: method `to_syn` is private
3 error[E0599]: no method named `spell` found for {Struct,EnumValue,Alternative}
```

## Gate

- `cargo test -p prebindgen --all-features` — 632 lib (one new), 34 doctests (11 `compile_fail`, up from 8)
- clippy `--all-targets --all-features -D warnings` on stable **and** 1.85.0
- `examples/regen-check.sh` — every generated artifact byte-identical

### DIFF
--- a/prebindgen/src/api/core/flat/surface.rs
+++ b/prebindgen/src/api/core/flat/surface.rs
@@ -53,6 +53,10 @@ fn doors(file: &syn::File) -> Vec<String> {
         let syn::ReturnType::Type(_, t) = ty else {
             return None;
         };
+        names_a_type(t)
+    }
+    /// A `syn` mention in any type position, not just a return.
+    fn names_a_type(t: &syn::Type) -> Option<String> {
         let toks: Vec<proc_macro2::TokenTree> =
             quote::ToTokens::to_token_stream(t).into_iter().collect();
         for (i, tt) in toks.iter().enumerate() {
@@ -62,7 +66,6 @@ fn doors(file: &syn::File) -> Vec<String> {
             if *id != "syn" {
                 continue;
             }
-            // `syn :: <Name>` — two `:` puncts then the name.
             if let Some(proc_macro2::TokenTree::Ident(name)) = toks.get(i + 3) {
                 let n = name.to_string();
                 if !LEAF.contains(&n.as_str()) {
@@ -71,6 +74,12 @@ fn doors(file: &syn::File) -> Vec<String> {
             }
         }
         None
+    }
+    /// `Origin<..>` is the seal itself — every element carries one, and its own
+    /// accessors are what this file exists to keep sealed.
+    fn is_origin(t: &syn::Type) -> bool {
+        let toks = quote::ToTokens::to_token_stream(t).to_string();
+        toks.starts_with("Origin ") || toks.starts_with("Origin<")
     }
     fn is_transformer(sig: &syn::Signature) -> bool {
         TRANSFORMERS.contains(&sig.ident.to_string().as_str())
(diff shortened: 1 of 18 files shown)


prebindgen/src/api/core/flat/surface.rs
```diff
@@ -0,0 +1,328 @@
...
+    fn names_a_node(ty: &syn::ReturnType) -> Option<String> {
+        let syn::ReturnType::Type(_, t) = ty else {
+            return None;
+        };
+        names_a_type(t)
+    }
+    /// A `syn` mention in any type position, not just a return.
+    fn names_a_type(t: &syn::Type) -> Option<String> {
+        let toks: Vec<proc_macro2::TokenTree> =
+            quote::ToTokens::to_token_stream(t).into_iter().collect();
```
milyin, line 61: [P2] Recurse into token groups when detecting syntax nodes

This only iterates the top-level TokenTrees. Parenthesized and tuple types are represented by a TokenTree::Group, so a concrete door such as pub fn leak() -> (syn::Type,) is missed: names_a_type sees the group, never sees syn, and returns None. The new item-kind mutation test therefore proves the listed item forms only, not the stated any-type-position invariant. Please recurse into groups (or inspect the syn::Type tree directly) and add a grouped-type mutation case.

— Codex (GPT-5)

PR #370, opened 2026-08-05
Split prebindgen into layers: Phase A + kotlin-codegen carve-out

milyin:
Splits the monolithic `prebindgen` crate into layers. This is **Phase A + B0** — all preparation plus the one carve-out that was independent of the rest. The remaining carves (`prebindgen-flat`, `prebindgen-registry`, `prebindgen-c`, `prebindgen-jni`) follow in a later PR.

Stacked on `c8-delete-the-guards`.

## Why

One 75k-line crate carried everything, with three consequences:

- A source crate (`zenoh-flat`) that only calls `init_prebindgen_out_dir()` compiled the whole JNI adapter and linked `jni`.
- A shipped binding library (`zenoh-flat-jni`) depended on the entire generator — `syn`, `quote`, `prettyplease`, `toml` — purely to reach `JniBindingError` / `box_j*` / `CachedIfaceMethod` at run time.
- Two layers that should be independent formed dependency cycles: `flat → registry` and `cbindgen → jnigen`.

**Both cycles are now gone.**

## What changed

| Commit | Effect |
|---|---|
| **B0** | `kotlin-codegen` → [its own repo](https://github.com/milyin/kotlin-codegen). Zero dependencies, publishable to crates.io today. |
| **A1** | `TypeKey` → `flat`. Breaks `flat → registry`. |
| **A2** | `prebindgen-jni-runtime` + `prebindgen-c-runtime` carved out; emitted paths repointed. |
| **A3** | Shared decl vocabulary → `core::decl`. Breaks `cbindgen → jnigen`. |
| **A5** | `unstable-cbindgen` feature deleted. |
| **A4** | `Emit` → `flat`. `api/core/flat/` now has **zero** code references to any core sibling. |

A shipped binding library now depends on ~350 lines instead of 75k: `prebindgen-jni-runtime` pulls only `jni`, `prebindgen-c-runtime` pulls nothing.

## Two corrections to the original plan

**The expand decls are shared, not JNI-only.** `fun!` expands to `FunctionDecl`, and `example-cbindgen/build.rs` already writes `convert!(Millis).input(fun!(millis_from_raw))` — so `FunctionDecl` must reach the C adapter, and it holds `ExpandParamDecl` / `ExpandReturnDecl`. Safe, because all four name only `TypeKey`, `Origin<syn::Type>`, and `syn` types — never a Kotlin or JNI type.

**`const_path_alias` moved with `Emit`.** It sat beside the `Prebindgen` trait but is pure syntax rendering with no pipeline dependency, and `Emit::const_alias` is its only caller. Leaving it would have made `flat` reach up into the registry layer.

## Deliberately deferred

**The visibility relaxations.** The plan had them in A4, but done in-place `pub(in crate::api::core)` still compiles, so relaxing now would weaken the seal early with no compiler check that it was done right. At carve time the compiler names exactly which items need `pub`. This is where over-publicizing is the real hazard, so it belongs under maximum forcing function.

## One thing the split genuinely costs

`Emit::new()` is `pub(in crate::api::core)` — *only core may mint the capability*. Once core becomes `flat` + `registry`, that is **inexpressible**: registry must mint, and Rust has no cross-crate friend. It becomes `pub` at carve time.

The important half survives: `as_syn` / `spell` / `stripped_syntax` stay private **inside `flat`**, so "classify off `kind`, spell with `spell()`" still holds. What is lost is only the secondary control over *who mints*. `emit.rs`'s module doc currently asserts the stronger claim and should be amended when that lands.

## Verification

- **535 lib + 34 doc + 20 + 10 passing, 0 failures** at every commit
- `clippy` + `fmt --check` clean on **both 1.85.0 and stable**
- `regen-check.sh` **byte-identical** — which is what proves A3/A4 were pure moves
- A2's golden change verified as cosmetic: identical after normalizing the rename, whitespace, and commas (the residue is `prettyplease` re-wrapping lines the longer crate name pushed past its width)
- `cargo tree`: `prebindgen-c-runtime` has no dependencies; `prebindgen-jni-runtime` has only `jni`

Dropping `unstable-cbindgen` also fixes a real correctness gap: a bare `cargo test` used to silently skip ~60 cbindgen tests, and the committed goldens were feature-sensitive. CI's duplicate test line is gone.

## CI note

`kotlin-codegen` is a sibling path dependency until it reaches crates.io, so each job now checks out both repos into `$GITHUB_WORKSPACE` and runs from the `prebindgen` one — the same layout local development uses, so the path dep resolves unchanged with no `Cargo.toml` rewriting. A resolve job pins one `kotlin-codegen` SHA per run.

**This PR is the first real exercise of that wiring.**

## Follow-ups (not in this PR)

- `zenoh-flat-jni/kotlin/generated/.prebindgen-kotlin-output` → `.kotlin-codegen-output` (stale ownership marker; failure mode is a loud refusal to write, never data loss)
- `zenoh-flat-jni` / `zenoh-flat-c` repointed to the runtime crates
- The remaining four carves


.github/workflows/rust.yml
```diff
@@ -23,15 +23,44 @@ env:
...
+  resolve-kotlin-codegen:
+    name: Resolve kotlin-codegen main
+    runs-on: ubuntu-latest
+    outputs:
+      sha: ${{ steps.resolve.outputs.sha }}
+    steps:
+    - name: Resolve kotlin-codegen main to a commit
+      id: resolve
+      run: |
+        sha=$(git ls-remote https://github.com/milyin/kotlin-codegen.git refs/heads/main | cut -f1)
```
milyin: **Pin kotlin-codegen across runs, not only across jobs.** This resolves mutable `main` on every workflow invocation, so rerunning this exact prebindgen commit later can compile a different kotlin-codegen revision—or fail after an unrelated breaking change. Because this PR removes the in-tree implementation, that external revision is now part of the source needed to reproduce the build. Please record an immutable expected SHA (or tag/rev) in the repository and have CI check out that value; updates can then be deliberate and reviewed alongside compatibility changes.

— Codex (GPT-5)

milyin: Agreed, and fixed in 14e3403 — the reproducibility argument is the right one: since this PR removes the in-tree emitter, the kotlin-codegen revision is now part of the source needed to build a given commit, so resolving mutable `main` meant an old commit could compile against something it was never tested with.

CI now reads `KOTLIN_CODEGEN_REV` (currently `201497cc`), pinned in the workflow's `env:` block, and every job checks out that exact commit. Bumping it is a reviewed one-line diff that can land alongside whatever compatibility change it needs.

That also let me delete the `resolve-kotlin-codegen` job outright — it existed only to keep three jobs from independently resolving `main` mid-run, which pinning makes moot. Net: one fewer job, and reproducible.

Local development still tracks kotlin-codegen `main` via `download_repositories.sh`, matching how the other sibling repos already work — fluid for dev, pinned for CI.

prebindgen/src/api/lang/jnigen/jni/builder.rs
```diff
@@ -2182,26 +2182,6 @@ pub(crate) fn is_wire_type(ty: &syn::Type) -> bool {
...
-        .iter()
-        .take(path.segments.len() - 1)
-        .map(|s| s.ident.to_string())
-        .collect::<Vec<_>>()
-        .join("::")
-}
-
 pub(crate) fn default_err_type() -> syn::Type {
     syn::parse_quote!(__JniErr)
 }
```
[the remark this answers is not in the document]
milyin: This one is a false positive — the code compiles, and CI is green on both 1.85.0 and stable.

`local_path_prefix` is still referenced in `builder.rs` (lines 1260 and 1284), but its definition moved to `api/core/decl.rs` and reaches those call sites through the existing re-export chain:

- `api/core/decl.rs` — the definition
- `api/lang/jnigen/jni/decl.rs:18` — `pub(crate) use crate::api::core::decl::{declared_origin, local_path_prefix, ConvertSpec, LocalField, LocalVariant};`
- `api/lang/jnigen/jni/mod.rs:658` — `pub use decl::*;`
- `builder.rs:9` — `use super::*;`

So the unqualified calls resolve unchanged. That re-export was deliberate: keeping `jni`'s `pub use decl::*;` intact is what holds `prebindgen::lang::*` byte-identical across this refactor, which is the property the whole commit is trying to preserve.

The one call site that *did* need qualifying is in the C adapter, `cbindgen/trait_impl.rs:1570`, which now says `crate::api::core::decl::local_path_prefix(path)` — it has no such glob import, and breaking its dependency on `jnigen` is the point of the commit.

PR #390, opened 2026-08-06
Let a value form carry an Option<sum> field (#220)

milyin:
Closes #220.

## The asymmetry

`Option<sum>` was refused as a **value-form** field while the same field on a **data class** worked — the `fromParts` bridge emits a `<field>__present` flag beside its `<field>__tag` (`PlanFieldKind::Sum { optional }`). Same shape, one position over, different answer.

Rejecting was right when #213 landed: without gating, the emitter composes `match &(&__vf).result` on an `Option<Sum>` against bare variant patterns — `E0308` in the consumer crate. A named error beat emitting that.

## The issue's premise is out of date

#220 proposes teaching core a present-flag notion on the output side, dual to `FlatLeaf::is_present_flag`. **Core does not need one.** The *conditional value form* work already built absence into the leaf model, and it covers this case unchanged — `prebindgen-registry/src/unfold.rs::flatten` today already:

- marks the field's last path step optional — `PathStep::field(last, opt && decomposed)`, and `decomposed` is true for a sum (`FieldDecon::Leaves`);
- ORs the field's own `opt` into every synthesized leaf's `nullable`.

And everything downstream of that flag is already correct: the tag boxes so JVM `null` cannot alias tag `0` (`leaf_is_prim`), `leaf_slot` knows each slot's wire default, `iface.rs` types the selector `Int?` and the group nullable, and `kotlin_emit.rs` puts `null -> null` ahead of the real tags. All of it JVM-tested today through `ledgerEach`.

**So this PR changes no core, no `iface.rs`, no `kotlin_emit.rs`.**

## The one function

`delivery.rs`'s sum-segment loop folded the selector's path with `fold_steps`, which has no optional handling — and said so:

> `// no step on it is optional, since an optional sum is refused where the leaves are built.`

A sum segment can't take the per-leaf treatment `reach_leaf` gives an ordinary optional field: its leaves aren't independent, only one group is live. So the **whole segment** gates as one tuple bind whose absent arm carries every slot's default — the shape a conditional hoist already emits, applied to an optional step inside the segment's own path:

```rust
let (__cb0_obj1, __cb0_obj2, __cb0_obj3): (JObject, JObject, JObject) = {
    let __so1: &::core::option::Option<_> = &(&__vf0).outcome;   // coerced, #268
    match __so1 {
        Some(__sg1) => { /* today's encode_sum_group, unchanged */ (…) }
        None => (JObject::null(), JObject::null(), JObject::null()),
    }
};
```

`encode_sum_group` is reused untouched. The optional step follows `reach_leaf`'s own division: a **field** step goes through `bind_as_option` so the destructure doesn't care how the source spelled the optional (#268); a **call** step keeps its direct match, since an owned position can't be coerced that way.

`Vec<sum>` stays refused — variable arity genuinely has no fixed layout.

## Coverage

- `a_sum_field_behind_option_or_vec_is_rejected_by_name` → `a_vec_sum_field_is_rejected_by_name`; the `Option` row and its rationale are gone.
- **New** `an_optional_sum_field_gates_its_whole_segment` — the coercion site, the tuple-bound absent arm, the tag defaulting to JVM null rather than a raw `jint`, and Kotlin's `null -> null`.
- **New** `a_bare_sum_field_takes_no_gate` — the dual, so the optional path isn't entered when there's nothing to gate.
- `examples/covertest-kotlin` gains `Probe`, whose value form has an `Option<Lookup>` field. The JVM harness pins the distinction the boxing exists for:

  ```kotlin
  probeNew(9, -2, …) { seq, tag, _, _ -> "$seq:$tag" }  // "9:null"  — no sum at all
  probeNew(9,  0, …) { seq, tag, _, _ -> "$seq:$tag" }  // "9:0"     — a PRESENT Lookup.Absent
  ```

  A raw `jint` selector could not tell those apart, which is the whole reason no present flag is needed. **50 sections pass** under `./gradlew run`.

  Deliberately kept off `ReportStruct`: `Report` is embedded twice in `Ledger`, so a field there would add 3 positional slots to `reportEach` and 6 to `ledgerEach` and bury the change under signature churn.

## Verified

`cargo test --all`; clippy `-D warnings` and `cargo fmt --check` (CI's exact config) on **1.85.0 and stable**; `examples/regen-check.sh` clean. Generated output is **additive only** — 866 insertions, no deletions, no existing declaration touched — and `regen-check.sh --with-zenoh-flat-jni` is byte-identical, since zenoh-flat's `ReplyStruct.result` is a bare `ReplyResult`.

`docs/sum-types.md` §4.3 now records that both output paths gate `Option<sum>`, by two different means — a present flag on the `fromParts` bridge, the selector's own nullability in a leaf list.

### DIFF
--- a/prebindgen-jni/src/jni/emit/delivery.rs
+++ b/prebindgen-jni/src/jni/emit/delivery.rs
@@ -949,8 +949,9 @@ pub(crate) fn encode_plan_leaves(
         let leaf = &plan.leaves[seg.start];
         let (base, base_is_ref, path, _) = rebase(leaf);
         // The value to `match` on. The selector's own path reaches the sum
-        // (empty when the sum IS the value); no step on it is optional, since
-        // an optional sum is refused where the leaves are built.
+        // (empty when the sum IS the value), and a step on it MAY be optional:
+        // the refusal that used to guarantee otherwise is gone (#220), which is
+        // what the gate below exists for.
         //
         // A plain field chain is borrowed DIRECTLY (`&base.a.b`) rather than
         // through the base (`&(&base).a.b`). The two are the same value, but
@@ -976,7 +977,30 @@ pub(crate) fn encode_plan_leaves(
                 // statement of it.
                 let opt_e = fold_steps(&qualify, &path[lead..=k], projected, false);
                 let bind = format_ident!("__sg{}", seg.start);
-                let inner = fold_steps(&qualify, &path[k + 1..], quote!(#bind), true);
+                // The tail is empty for every shape that reaches here: a sum
+                // leaf's path stops AT the sum. Stated because two things below
+                // would otherwise be quietly wrong for a non-empty one — a
+                // second optional step would compose `match &Option<..>` against
+                // bare variant patterns, which is the E0308 the deleted
+                // `builder.rs` assert used to pre-empt by name.
+                debug_assert!(
+                    path[k + 1..].is_empty(),
+                    "sum segment: a step after the gated one ({} left) — the tail \
+                     is assumed empty here",
+                    path.len() - k - 1,
+                );
+                // What the `match` binds, asked of the step rather than assumed:
+                // the FIELD branch scrutinizes `&Option<_>` (that is what
+                // `bind_as_option` is for), so ergonomics binds `&Sum` — a
+                // borrow. Only an owned-yielding CALL binds an owned value. The
+                // literal `true` disagreed with `reach_leaf`, which passes
+                // `false` for its analogous recursion.
+                let inner = fold_steps(
+                    &qualify,
+                    &path[k + 1..],
+                    quote!(#bind),
+                    path[k].yields_owned(),
+                );
                 (inner, Some((k, opt_e, bind)))
             }
         };
--- a/prebindgen-jni/src/jni/emit/delivery.rs
+++ b/prebindgen-jni/src/jni/emit/delivery.rs
@@ -958,8 +958,28 @@ pub(crate) fn encode_plan_leaves(
         // rejects once a sibling leaf has moved another field out of it — and
         // borrowing this field while sibling fields move is exactly what a
         // consuming value form does.
-        let (matched, lead) = project_leading_fields(&base, base_is_ref, &path);
-        let matched = fold_steps(&qualify, &path[lead..], matched, false);
+        let (projected, lead) = project_leading_fields(&base, base_is_ref, &path);
+        // An `Option<sum>` field gates the WHOLE segment, not each slot (#220).
+        // A sum's leaves are not independent — only one group is live per value
+        // — so absence cannot be the per-leaf `null` `reach_leaf` gives an
+        // ordinary optional field. It is one tuple bind whose `None` arm carries
+        // every slot's default, which is the shape a conditional value form's
+        // hoist already emits below; this applies it to an optional step inside
+        // the segment's own path.
+        let opt_at = (lead..path.len()).find(|&i| path[i].is_optional());
+        let (matched, gate) = match opt_at {
+            None => (fold_steps(&qualify, &path[lead..], projected, false), None),
+            Some(k) => {
+                // Through the optional step INCLUSIVE, then the rest off the
+                // binding — the same split `reach_leaf` makes, so the borrow in
+                // front of it stays the ordinary rule rather than a second
+                // statement of it.
+                let opt_e = fold_steps(&qualify, &path[lead..=k], projected, false);
+                let bind = format_ident!("__sg{}", seg.start);
+                let inner = fold_steps(&qualify, &path[k + 1..], quote!(#bind), true);
+                (inner, Some((k, opt_e, bind)))
+            }
+        };
         let (group_stmts, group_args) = encode_sum_group(
             ext,
             registry,
@@ -969,6 +989,42 @@ pub(crate) fn encode_plan_leaves(
             fail,
             emit,
         );
+        let group_stmts = match gate {
+            None => group_stmts,
+            Some((k, opt_e, bind)) => {
+                let ids: Vec<&syn::Ident> = obj_idents[seg.clone()].iter().collect();
+                let slots: Vec<Slot> = plan.leaves[seg.clone()]
+                    .iter()
+                    .map(|l| leaf_slot(registry, l))
+                    .collect();
+                let tys = slots.iter().map(|s| &s.ty);
+                let defaults = slots.iter().map(|s| &s.default);
+                // A FIELD step composes to a borrow, so it goes through a
+                // coercion site and the destructure stops caring which
+                // representation the source spelled the optional as (#268). A
+                // CALL yields its own owned value, whose payload downstream may
+                // move, so it keeps the direct match — the same division
+                // `reach_leaf` makes.
+                let (prelude, scrutinee) = if path[k].is_field() {
+                    let opt_bind = format_ident!("__so{}", seg.start);
+                    (bind_as_option(&opt_e, &opt_bind), quote!(#opt_bind))
+                } else {
+                    (TokenStream::new(), opt_e)
+                };
+                quote! {
+                    let (#(#ids,)*): (#(#tys,)*) = {
+                        #prelude
+                        match #scrutinee {
+                            ::core::option::Option::Some(#bind) => {
+                                #group_stmts
+                                (#(#ids,)*)
+                            }
+                            ::core::option::Option::None => (#(#defaults,)*),
+                        }
+                    };
+                }
+            }
+        };
         // The whole segment — its slot declarations and its `match` — is
         // routed like any other leaf under the same form.
         match hoisted.conditional(&leaf.path) {
--- a/prebindgen-jni/src/jni/emit/delivery.rs
+++ b/prebindgen-jni/src/jni/emit/delivery.rs
@@ -977,17 +977,29 @@ pub(crate) fn encode_plan_leaves(
                 // statement of it.
                 let opt_e = fold_steps(&qualify, &path[lead..=k], projected, false);
                 let bind = format_ident!("__sg{}", seg.start);
-                // The tail is empty for every shape that reaches here: a sum
-                // leaf's path stops AT the sum. Stated because two things below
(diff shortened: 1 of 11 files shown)


prebindgen-jni/src/jni/tests/value_form.rs
```diff
@@ -535,23 +537,227 @@ fn a_sum_field_behind_option_or_vec_is_rejected_by_name() {
...
+    let kotlin = gen
+        .write_kotlin(&dir.join("kotlin"))
+        .expect("write_kotlin")
+        .iter()
+        .map(|p| std::fs::read_to_string(p).unwrap())
+        .collect::<Vec<_>>()
+        .join("\n");
+
+    assert!(
+        !rust.contains("::core::option::Option::None => ("),
```
milyin: **This negative assertion is vacuous — the test cannot fail.**

`prettyplease` renders a tuple-valued match arm as `::core::option::Option::None => {` + newline + `(`, never `None => (` on one line. I verified this empirically: dropping this exact assertion into `an_optional_sum_field_gates_its_whole_segment` (which *does* emit the gate) makes it pass there too.

So the dual test the PR advertises as pinning "the optional path isn't entered when there's nothing to gate" would stay green even if the gate were emitted for a bare sum field. Only the `!kotlin.contains("null -> null")` check below carries weight.

The sibling test already gets this right by asserting on the whitespace-stripped text (`rc`); this one should do the same.

```suggestion
        !rust.split_whitespace().collect::<String>().contains("::core::option::Option::None=>{("),
```

prebindgen-jni/src/jni/emit/delivery.rs
```diff
@@ -958,8 +958,28 @@ pub(crate) fn encode_plan_leaves(
...
-        let matched = fold_steps(&qualify, &path[lead..], matched, false);
+        let (projected, lead) = project_leading_fields(&base, base_is_ref, &path);
+        // An `Option<sum>` field gates the WHOLE segment, not each slot (#220).
+        // A sum's leaves are not independent — only one group is live per value
+        // — so absence cannot be the per-leaf `null` `reach_leaf` gives an
+        // ordinary optional field. It is one tuple bind whose `None` arm carries
+        // every slot's default, which is the shape a conditional value form's
+        // hoist already emits below; this applies it to an optional step inside
+        // the segment's own path.
+        let opt_at = (lead..path.len()).find(|&i| path[i].is_optional());
```
milyin, line 970: **Stale comment directly above this line now contradicts the code** (lines 951-953, just outside the hunk):

> // The value to `match` on. The selector's own path reaches the sum
> // (empty when the sum IS the value); **no step on it is optional, since
> // an optional sum is refused where the leaves are built.**

That is exactly the invariant this PR deletes in `builder.rs`, and the very next statement (`let opt_at = …`) exists because a step on it *can* now be optional. In a codebase where these comments are the primary record of why the emitter is shaped the way it is, leaving this one in place will send the next reader looking for a refusal that is gone.

prebindgen-jni/src/jni/emit/delivery.rs
```diff
@@ -958,8 +958,28 @@ pub(crate) fn encode_plan_leaves(
...
+        let (matched, gate) = match opt_at {
+            None => (fold_steps(&qualify, &path[lead..], projected, false), None),
+            Some(k) => {
+                // Through the optional step INCLUSIVE, then the rest off the
+                // binding — the same split `reach_leaf` makes, so the borrow in
+                // front of it stays the ordinary rule rather than a second
+                // statement of it.
+                let opt_e = fold_steps(&qualify, &path[lead..=k], projected, false);
+                let bind = format_ident!("__sg{}", seg.start);
+                let inner = fold_steps(&qualify, &path[k + 1..], quote!(#bind), true);
```
milyin: Two things about the tail after the gated step, both latent today but worth stating since the removed `builder.rs` assert was the only *named* diagnostic in this area:

1. **`owned = true` disagrees with `reach_leaf`.** In the FIELD branch below, `#bind` comes from matching a `&::core::option::Option<_>`, so match ergonomics binds it as `&Sum` — a borrow, not an owned value. `reach_leaf` passes `false` for its analogous recursion (`reach_leaf(.., quote!(#nested), /*base_is_ref=*/true, ..)` → `fold_steps(.., false)`). With `true`, a `Call` step in `path[k + 1..]` would be composed as `&(&Sum)` and hand the accessor `&&Sum` → E0308 in the consumer crate. `true` is only right for the CALL branch (owned `Option<T>` scrutinee); one flag is serving two branches that disagree.

2. **`opt_at` finds only the first optional step**, and `fold_steps` has no optional handling, so a *second* optional step in the tail would compose `match &Option<..>` against bare variant patterns — the same `E0308` the deleted assert was written to pre-empt, now with no named error.

Both are unreachable right now only because `synth_sum_leaves` gives every sum leaf an empty `path`, so `path[k + 1..]` is always empty. That invariant isn't stated anywhere here. A `debug_assert!(path[k + 1..].is_empty(), ..)` (or just passing `path[k].yields_owned()` instead of the literal `true`) would keep this honest.

prebindgen-jni/src/jni/tests/value_form.rs
milyin: Fixed in 7925993, with the whitespace-stripped form.

One thing worth recording, because it changes what the test is worth: the assertion is now **able** to fail — the needle `::core::option::Option::None=>{(` is exactly what the sibling test asserts is *present* in the gated emission — but I could not construct a mutation that makes it fail **in this fixture**. Forcing the predicate (`find(|_| true)`) leaves the test green, because for a bare sum field `project_leading_fields` consumes the `result` step as a plain field, so `lead == path.len()` and the search range is empty before the predicate is ever consulted.

So the dual’s real guard against the gate firing is structural rather than asserted, and only the Kotlin `null -> null` check is falsifiable by touching the gate. The fix is still worth having — it is now a claim about the output instead of a claim about `prettyplease`’s line breaking — but it should not be read as pinning more than it does.

prebindgen-jni/src/jni/emit/delivery.rs
milyin: Fixed in 7925993. It now reads:

> The value to `match` on. The selector's own path reaches the sum (empty when the sum IS the value), and a step on it MAY be optional: the refusal that used to guarantee otherwise is gone (#220), which is what the gate below exists for.

Agreed on the reasoning — a comment asserting a refusal that the same PR deletes is worse than no comment, because it reads as a reason not to look.

prebindgen-jni/src/jni/emit/delivery.rs
milyin: Both taken in 7925993, and your framing of (1) is better than mine — I had flagged the same flag as "inert today", but `reach_leaf` passing `false` for the analogous recursion is the argument that settles it: two branches were sharing a literal that only one of them makes true.

Fixed as `path[k].yields_owned()` rather than a branch on `path[k].is_field()`. `yields_owned` is `Call { owned: true }` and nothing else, so it answers correctly for both arms from the step itself — the FIELD arm gets `false` (the `&Option<_>` scrutinee binds `&Sum`) and an owned-yielding CALL gets `true` — and there is no second place stating which arm is which.

For (2) I took the `debug_assert!(path[k + 1..].is_empty(), ..)` over my own suggestion of asserting no *second* optional step: emptiness is the invariant that actually holds and it subsumes the other. It names the E0308 the deleted `builder.rs` assert used to pre-empt, so the diagnostic that left `builder.rs` has a home. 253 lib tests, both Kotlin examples and `emitcheck` run without it firing.

Independently of the reachability question, I probed the shape that would put a second optional step in one path — `Option<Child>` where `Child`'s own value form holds `Option<sum>` — through `emitcheck`. It is refused upstream by name ("a value form nested under another one that is reached through `Option` — conditional hoists do not nest"), so the assert documents an invariant that two separate mechanisms currently keep.

prebindgen-jni/src/jni/emit/delivery.rs
```diff
@@ -949,17 +949,61 @@ pub(crate) fn encode_plan_leaves(
...
+                // would otherwise be quietly wrong for a non-empty one — a
+                // second optional step would compose `match &Option<..>` against
+                // bare variant patterns, which is the E0308 the deleted
+                // `builder.rs` assert used to pre-empt by name.
+                debug_assert!(
+                    path[k + 1..].is_empty(),
+                    "sum segment: a step after the gated one ({} left) — the tail \
+                     is assumed empty here",
+                    path.len() - k - 1,
+                );
```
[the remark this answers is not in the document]
milyin: Both halves taken, in b2b3e0c.

**`assert!`, not `debug_assert!`** — and the argument is stronger than "invariants should be enforced": a build script inherits the consumer's profile, so `debug_assert!` is absent from exactly the release build of a downstream crate where a mis-emission is most expensive to diagnose. That defeats the reason I gave for adding it — the deleted `builder.rs` assert was the only *named* diagnostic here, and a debug-only replacement hands back a bare E0308 in generated code the author did not write. It now matches `reach_leaf`'s own optional-step assert a few hundred lines up, message shape included.

**The condition** — agreed, with the reasoning worth stating because it cuts against my own commit message. Emptiness is true today (a sum leaf's path stops at the sum) but it is not the safety condition: a non-optional tail composes correctly through `fold_steps`, the more so now that the fold takes `path[k].yields_owned()` rather than a literal. Asserting emptiness would refuse a shape that works. A *second optional* step is the one that composes `match &Option<..>` against bare variant patterns, so that is what is asserted:

```rust
assert!(
    !path[k + 1..].iter().any(PathStep::is_optional),
    "jnigen unfold: leaf `{}` reaches its sum through TWO optional steps — the \
     segment gate has one `None` arm, so the second would be matched as if it \
     were the sum itself",
    leaf.name,
);
```

For the record on reachability: the shape that would trip it — `Option<Child>` where `Child`'s own value form holds `Option<sum>` — is refused upstream by name ("a value form nested under another one that is reached through `Option` — conditional hoists do not nest"), measured through `emitcheck`. So this assert is a second lock on a door that is already shut, which is the right weight for it — not a load-bearing guard, but the named error if the first lock ever opens.

255 lib tests, `regen-check.sh` byte-identical, clippy `-D warnings` + `fmt --check` on 1.85.0 and stable.

PR #395, opened 2026-08-11
Assert the C bindings' behaviour, not just that they build (#387)

### LINKED ISSUE #387: Need functional test
Current test only verifies the fact that build passes. But there is no behavior test for C and Kotin bindings, how the data flows across the boundaries

milyin:
Closes #387.

The Kotlin half of #387 turned out to be covered already: `covertest-kotlin` is a
52-section assert harness and the `covertest` CI job runs it on a real JVM. The C half
was not — `smoke.c` exercised the tagged union thoroughly and left the rest of the
declared surface compile-checked only, and `perftest-c`'s `correctness()` asserts ran
only when somebody benchmarked by hand.

### `smoke.c`: five new sections

| section | what crosses |
|---|---|
| handle lifecycle | state written through the `&mut` route, read back through the borrow accessors; a clone that diverges from the original |
| error channel | a rejected `&str` parse, a domain error, and an `operation_t` discriminant no variant has — the union tag's validate-before-materialise rule one type down |
| owned returns | a `char *` and a malloc'd `double[]` + length, released by the single universal `example_free`; an empty `Vec` is the `(NULL, 0)` pair, not a block |
| closure struct | `call` once per item in order, `drop` exactly once — the context is a malloc'd block, so LeakSanitizer enforces the second half rather than the count alone |
| by-value types | `foo_t` / `inside_foo_t` both ways, and `caption_t`, whose owned `char *` field a data struct leaves to the caller |

### `smoke-asan.sh`: run `perftest-c` too

Its `correctness()` already asserted the five parameter-passing semantics of the by-value
`.repr_c_struct()` surface; this is about executing them in CI under the leak detector,
not about the timings (meaningless under ASan at `PERFTEST_N=200`, so the measured block
goes to `/dev/null` and only the asserts and any sanitizer report decide the exit status).

Doing so immediately found a leak: `bench_put_by_take` seeded its payload with a label
that the first loop iteration overwrote unfreed. Seeded NULL now — the loop provides one
per iteration anyway.

### Verified locally

- `examples/smoke-asan.sh` — both programs clean under ASan/LSan/UBSan
- `examples/covertest-kotlin && ./gradlew run` — `PASS - 52 sections`
- no Rust touched, no generated-file drift

### DIFF
--- a/examples/example-cbindgen/c/smoke.c
+++ b/examples/example-cbindgen/c/smoke.c
@@ -1,6 +1,10 @@
 /*
- * C smoke test over the prebindgen-generated `example_flat` C ABI, focused on
- * the **tagged union** (`Shape`, declared `.tagged_union()`).
+ * C smoke test over the prebindgen-generated `example_flat` C ABI: does the
+ * data actually arrive intact, in both directions, for every kind of thing the
+ * C adapter emits? Its longest section is the **tagged union** (`Shape`,
+ * declared `.tagged_union()`); the rest of the declared surface — the opaque
+ * handle, the `char **e` error channel, the owned `char *` / array returns, the
+ * closure struct, and the by-value structs and enums — follows it.
  *
  * A data-carrying Rust enum crosses by value as a `#[repr(C)]` enum with payload
  * variants, which cbindgen renders as a tag + `union`:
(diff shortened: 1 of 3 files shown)


milyin:
Closing: its work is on `main`, split across two merged PRs.

- The **leak it found** — `bench_put_by_take` seeding a payload label that the
  first iteration overwrites unfreed — and the sanitizer runner change that
  catches it landed as #699. `examples/smoke-asan.sh` now compiles and runs both
  `example-cbindgen`'s `c/smoke.c` and `perftest-c`'s `c/perftest.c` through one
  helper, so `correctness()` and LeakSanitizer see the by-value
  `.repr_c_struct()` surface in CI rather than only when somebody benchmarks by
  hand.
- The **five `smoke.c` sections** — handle lifecycle, error channel, owned
  returns, closure struct, by-value types — landed as #700. They pass on `main`
  unchanged.

What is not carried over is this branch's `smoke.c` base: `main` gained three
callback-argument sections of its own since August, so the recovered sections
were appended to that file rather than the branch's version replacing it. The
branch is left in place.

— Claude Code (Opus 5)

PR #396, opened 2026-08-11
Stop leaking a JSONL file per rustc invocation (#201)

### LINKED ISSUE #201: it seems that prebindgen generated files increasingly taking space
this size is not normal (this is prebindgnen-marked project)
```
[1] Cleaning: ./PREBINDGEN/zenoh-flat
     Removed 686092 files, 156.1GiB total
```

milyin:
Closes #201.

## What leaked

`#[prebindgen]` named each capture `{group}_{pid}_{thread}.jsonl`. Every rustc process therefore created another full capture. The build script clears its prebindgen directory only when Cargo reruns that script, so repeated `cargo check`, `build`, `test`, `clippy`, and rustdoc compilations could keep adding JSONL files to an unchanged Cargo build-output directory.

For a crate the size of `zenoh-flat` (310 marked items), each repeated capture is roughly 0.5 MiB per build-hash directory.

## What this PR fixes

Capture names are now keyed to the compilation unit rather than the rustc process:

- Derive the id from **every** `-C metadata` / `--codegen metadata` component. rustc parses this option with `parse_list`, so occurrences accumulate, each contributes its whitespace-separated words, and `StableCrateId::new` sorts them; the id mirrors that exactly.
- Hash the components into a fixed 16-digit hexadecimal identifier before using them as a filename component. This also makes arbitrary valid metadata such as `foo/bar` or a Windows-style backslash safe.
- When rustdoc supplies no metadata, hash its OS-native command line **minus the arguments that only run the tests or render diagnostics** (`--test-args`, `--test-run-directory`, `--runtool`, `--runtool-arg`, `--nocapture`, `--color`, `--error-format`, `--json`, `--diagnostic-width`), so a doctest filter no longer mints a new capture.
- Reset each compilation-unit capture once per process, then append all records for that unit to the same file.
- Hold a per-capture-file mutex across both the first reset and every append, preventing concurrent macro expansion threads from unlinking or interleaving records.
- Propagate reset, lock, and append failures as source-positioned compiler diagnostics.
- Serialize each JSONL batch before opening the file and write the batch in one operation.

The number of files depends on which distinct units Cargo asks rustc/rustdoc to compile, but it is bounded per unit rather than growing per invocation.

Both halves of the id deliberately err towards *more* ids. An id finer than the unit costs a stale file; an id shared by two units lets one unit reset the other's capture and silently drop items, which is the failure that breaks a downstream build. Accordingly the rustdoc filter is a deny-list of provably run-time-only options rather than an allow-list of compile-identifying ones: an unrecognized option stays in the id.

No reader change is needed: `Source` already discovers groups by the prefix before the first underscore and deduplicates records by name plus cfg.

## Review fixes (latest revision)

Both blockers found on `fea397d5` are fixed and covered by regressions.

**1. Repeated `-C metadata` components were collapsed** (#396 (comment) r3787045196). Reproduced end-to-end, with silent record loss, using nothing more exotic than `RUSTFLAGS="-Cmetadata=custom"` — which reaches every unit, so the lib and test units of `example-flat` shared one trailing component:

```
$ RUSTFLAGS=-Cmetadata=custom cargo test -p example-flat --lib && \
  RUSTFLAGS=-Cmetadata=custom cargo build -p example-flat
```

with a temporary `#[cfg(test)] #[prebindgen] pub fn test_only_probe()` in `example-flat`:

| | capture files in the shared `out/prebindgen` | `test_only_probe` |
| --- | --- | --- |
| `fea397d5` | `default_35de3c917bb9a9c3.jsonl` (44 records) | **lost** — the build unit reset the test unit's file |
| this revision | `default_d1c6e7b1eed6ba16.jsonl` (45), `default_9de8e1b72630ef9d.jsonl` (44) | kept |

The list semantics were re-verified directly against rustc, on 1.85.0 (legacy symbol hash) and 1.97.1 (v0 crate disambiguator), with a probe crate:

| arguments | 1.85.0 | 1.97.1 |
| --- | --- | --- |
| `-Cmetadata=b` | `h67633aed03fee1fb` | `CseyYMGf8ElKw_` |
| `-Cmetadata=a -Cmetadata=b` | `h5a1047a0533b34ea` | `Cs1XVTM2TReHw_` |
| `-Cmetadata=b -Cmetadata=a` | `h5a1047a0533b34ea` | `Cs1XVTM2TReHw_` |
| `-Cmetadata='a b'` | `h5a1047a0533b34ea` | `Cs1XVTM2TReHw_` |
| `-Cmetadata=a -Cmetadata=` | `h3dcdd8ed3cea6839` | `Cs2CqKFEsjyPW_` |
| `-Cmetadata=a` | `h3dcdd8ed3cea6839` | `Cs2CqKFEsjyPW_` |

So components accumulate, values split on whitespace, order does not matter (rustc sorts), and an empty value contributes nothing — which is what the id now implements, and which also settles the separate empty-value report: an empty occurrence adds no component, and a metadata list that is empty overall falls through to the argument fallback.

**2. The rustdoc fallback hashed runtime-only arguments** (#396 (comment) r3779811135). `cargo test --doc alpha` and `--doc beta` reach rustdoc as `--test-args alpha` / `--test-args beta`, which do not change what is compiled. Those options are now dropped before hashing. Three full lifecycle rounds (`check`, `build`, `test`, two doctest runs with a *different* filter each round, `cargo test --doc -- --nocapture`, all-target clippy, `doc`) leave a byte-for-byte identical file set after every round, each capture complete at 44 / 3 records.

## Follow-up: stale Cargo build-hash directories

This PR itself does **not** remove stale `target/*/build/<pkg>-<hash>/` directories. The remaining prebindgen-owned accumulation is now addressed by stacked PR #398 (issue #397).

PR #398 adds an explicit `cargo prebindgen clean` command. It acquires all discovered Cargo profile locks before mutation, removes only state-validated `out/prebindgen` payloads across old build hashes, and invalidates each exact producer so the capture is regenerated on next use. It deliberately leaves Cargo's `<pkg>-<hash>` directories and all unrelated artifacts intact.

Current split:

- Fixed here (#396): unbounded JSONL growth inside one unchanged `<pkg>-<hash>/out/prebindgen` directory.
- Fixed by the stacked follow-up (#398): explicit, concurrency-safe cleanup of prebindgen payloads across retained build hashes.
- Intentionally Cargo-owned: the surrounding `<pkg>-<hash>` directories themselves.

## Verification

- The two regressions above, plus: separator-containing metadata for both slash styles; a concurrent first-write/reset test with 32 synchronized writers; an incremental-rebuild check confirming rustc re-expands every `#[prebindgen]` after a source edit, so a reset capture is always rewritten complete.
- `cargo test --all --all-features`
- `cargo clippy --all-targets --all-features -- --deny warnings`
- Focused tests and strict Clippy on Rust 1.85.0 and 1.97.1
- CI rustfmt configuration and `git diff --check`
- `examples/regen-check.sh`: committed generated output remains byte-identical.

### DIFF
--- a/prebindgen-proc-macro/src/lib.rs
+++ b/prebindgen-proc-macro/src/lib.rs
@@ -37,9 +37,11 @@
 //! See also: [`prebindgen`](https://docs.rs/prebindgen) for the main processing library.
 //!
 use std::{
-    collections::HashSet,
+    collections::HashMap,
+    ffi::OsString,
     hash::{Hash, Hasher},
-    sync::{Mutex, OnceLock},
+    path::{Path, PathBuf},
+    sync::{Arc, Mutex, OnceLock},
 };
 
 use prebindgen::{get_prebindgen_out_dir, Record, RecordKind, SourceLocation, DEFAULT_GROUP_NAME};
@@ -139,8 +141,16 @@ impl Parse for PrebindgenArgs {
     }
 }
 
-/// Groups whose JSONL file this process has already reset.
-static STARTED_GROUPS: OnceLock<Mutex<HashSet<String>>> = OnceLock::new();
+#[derive(Default)]
+struct CaptureState {
+    initialized: bool,
+}
+
+type SharedCaptureState = Arc<Mutex<CaptureState>>;
+
+/// One lock per capture file. Macro expansions can run on several rustc
+/// threads, but unrelated groups do not need to block one another.
+static CAPTURE_STATES: OnceLock<Mutex<HashMap<PathBuf, SharedCaptureState>>> = OnceLock::new();
 
 /// Identify the compilation unit this expansion belongs to.
 ///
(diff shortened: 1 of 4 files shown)


prebindgen-proc-macro/src/lib.rs
```diff
@@ -135,51 +139,59 @@ impl Parse for PrebindgenArgs {
...
-    if let Some(p) = JSONL_PATHS.with(|path| path.borrow().get(group).cloned()) {
-        return p;
+    let path = get_prebindgen_out_dir().join(format!("{group}_{}.jsonl", unit_id()));
+    let first_write = STARTED_GROUPS
+        .get_or_init(Mutex::default)
+        .lock()
+        .map(|mut started| started.insert(group.to_string()))
+        .unwrap_or(false);
+    if first_write {
+        let _ = std::fs::remove_file(&path);
```
milyin, line 192: The initialization is not synchronized with the first append. `started.insert(...)` releases the mutex before `remove_file`: thread A can mark the group started, thread B can append its first record, and then thread A can unlink the file, silently losing B’s record and producing an incomplete capture. The sequential regression test cannot exercise this interleaving.

Please keep initialization and writes under a per-group lock (which also removes the need to rely on concurrent append/`write_all` atomicity), and propagate an unexpected removal error instead of ignoring it.

— Codex (GPT-5)

PR #401, opened 2026-08-13
Make Emit a registry-owned callback key

### LINKED ISSUE #375: Restore the Emit capability seal across the crate boundary
## What happened

Splitting the pipeline into `prebindgen-registry` (#374) broke the `Emit` capability. `Emit` exists so that captured Rust syntax is reachable **only** through a token handed to emission callbacks — code that *decides* cannot reach what code that *emits* must spell. That was enforced by `pub(in crate::api::core)`, with `compile_fail` doctests proving it from outside the crate.

A module-path seal cannot cross a crate boundary, and Rust has no friend visibility. When the pipeline moved out, the items its emission path calls had to become `pub`.

`docs/crate-split.md` predicted only `Emit::new` would open, and that `as_syn` / `spell` / `stripped_syntax` would all stay private. That prediction was wrong: the registry calls `spell` too.

## What opened

| Item | Why the registry needs it |
|---|---|
| `Emit::new` | the registry mints the capability for emission callbacks |
| `TypeRef::spell` | `write_rust`'s emission spells the captured type |
| `Origin::spell` | same, at the item level |
| `Flat::enum_item` | enum lowering reads the declared item |
| `Flat::classify` | the registry classifies declared types during scan |
| `Flat::add_local_function` | binding-local fns are synthesized before scanning |
| `TypeRef::{borrowed, optional, scalar}` | plan engines construct derived readings |
| `types_util::ident` | shared helper |

## What did NOT open — the load-bearing part

**Every route to the `syn` node itself** is still `pub(in crate::api::core)`:

- `TypeRef::{as_syn, stripped_syntax}`
- `TypeKind::to_syn`
- `Element::as_syn`, `Type::as_syn`, `Origin::as_syn`
- the three `spell(head, parts)` shape methods

So "classify off `kind`" still holds structurally. What opened is `spell()`, which yields a `TokenStream` — and `emit.rs` already documented re-parsing one to recover a node as **the accepted residual**:

> `Emit::spell` yields a `TokenStream`, so emission code can re-parse it and take the node apart. That is deliberate — emission is where syntax belongs — and closing it would mean an emission IR for Rust […] which is a much larger piece of work.

The regression is therefore precise: *that residual no longer requires holding an `Emit`.* Any consumer of `prebindgen` can now call `spell()` directly.

## Collateral

Four `compile_fail` doctests asserted the closed form of `TypeRef::spell`, `TypeRef::scalar`, `Flat::enum_item` and `Origin::spell`. They no longer prove anything and were converted to plain doctests. The remaining ~8 `compile_fail` examples still guard the node-level doors and still pass.

## Proposed fix

A **public-but-unconstructable token**, threaded through the widened signatures:

```rust
// in prebindgen (flat), public so the signatures can name it,
// unconstructable so only a blessed crate can produce one
pub struct Sealed(());

impl Sealed {
    #[doc(hidden)]
    pub const fn __mint_for_registry() -> Self { Sealed(()) }
}
```

`prebindgen-registry` calls `__mint_for_registry` once and holds the result; `spell(&self, _: &Sealed)` and friends then require it. A third-party adapter can name `Sealed` in a signature but cannot make one, so the capability is compiler-enforced again and the `compile_fail` doctests come back.

Open questions for whoever picks this up:

1. **Is a `#[doc(hidden)]` public minting fn honest?** It is reachable by anyone who reads the source. Stronger options: a private-module `Sealed` trait with a blanket impl, or moving `flat` back inside `prebindgen-registry` (which restores the seal exactly, at the cost of no standalone flat crate — considered and deferred in #374).
2. **Does the token belong on every widened item, or only the `spell` pair?** `classify` / `add_local_function` / the `TypeRef` constructors are not syntax doors; they may simply be legitimate public API now.
3. **Should `Emit` itself carry the token** rather than each method taking one, so the ergonomics stay `emit.spell(ty)`?

## Not urgent

Nothing is unsound and no generated output changes. This is an API-hygiene regression: an invariant that used to be compiler-enforced is now a documented convention. Tracked in `docs/crate-split.md`, which records both the correction and this follow-up.

milyin:
## Goal

Two jobs inside a language adapter want the captured syntax — the `syn` nodes
`#[prebindgen]` recorded from the source crate — for opposite reasons.

Generated Rust has to **spell** a type exactly as the source wrote it, or the emitted
converter's signature does not compile in the source crate's scope. Deciding what a type
*means* must never read that syntax. `prebindgen-flat`, the crate that parses the
captured records, classifies each type once — into `TypeKind` and the `TypeKey` identity
beside it — and an adapter that pattern-matches the `syn` node instead can reach a
different answer than `prebindgen-flat` did, silently, in generated code that still
compiles. Issue #211 removed that second reading from the adapters: only
`prebindgen-flat` matches on captured syntax.

`Emit` is what makes the difference a compiler check rather than a review habit. Captured
syntax is reachable only through an `Emit`, and only the pipeline's emission callbacks are
handed one. An adapter's planning, naming and validation code never holds one, so a call
that would re-read the source from there does not compile.

## What is broken

Until the crate split (#374) that check held by visibility: `Emit`'s constructor and every
spelling method were `pub(in crate::api::core)`, and the whole pipeline lived under that
path.

The split put the captured syntax and the pipeline in different crates: the syntax in
`prebindgen-flat`, the pipeline in `prebindgen-registry`, which is what hands the
capability to the callbacks a language adapter — `prebindgen-c`, `prebindgen-jni` —
implements. Rust has no cross-crate friend visibility. So `Emit::new`,
`TypeRef::spell` and `Origin::spell` all had to become `pub`, and any crate depending on
`prebindgen-flat` could spell captured syntax without ever holding an `Emit`. #375 tracks
that regression.

## What this does

The one capability type becomes two: a rendering protocol in the crate that owns the
syntax, and a key in the crate that decides where rendering is allowed.

- `prebindgen_flat::RustEmitter` is a public object-safe trait. Its default methods are
  the single implementation of rendering — spelling a type, re-emitting a captured item
  verbatim, a discriminant as written, and a struct or variant shape with the delimiters
  the source wrote.
- `prebindgen_registry::Emit` is a tuple struct with a private field and a `pub(crate)`
  constructor. It implements `RustEmitter`, and `Deref<Target = dyn RustEmitter>`
  forwards the calls, so adapter call sites keep writing `emit.spell(ty)` and the
  registry crate does not restate the method list.
- `prebindgen-flat`'s `flat::emit` module is crate-private, and `prebindgen-registry`
  does not re-export `RustEmitter` beside the `prebindgen-flat` modules it does
  re-export.

An adapter depending only on `prebindgen-registry` therefore can neither name the trait
nor construct an `Emit`. It renders only where the registry's `write_rust` or
`RegistryBuilder::convert_with` hands it a `&Emit`. Compile-fail doctests pin both
restrictions.

The direct access routes the split opened are closed again. `TypeRef::spell` and `Origin::spell`
are `pub(crate)`, and `Flat::enum_item` — a raw-item accessor kept only for two registry
tests — is deleted; those tests match on `Flat::declared_type` instead. The other items
#375 lists as widened stay public: `Flat::classify`, `Flat::add_local_function`,
`TypeRef::{borrowed, optional, scalar}` and `types_util::ident` build `prebindgen-flat`
values rather than hand out syntax, and an independent parser needs them.

## The residual

A different collector can depend on `prebindgen-flat` directly and implement
`RustEmitter` for a key of its own. That is deliberate: `prebindgen-flat` is usable
without the registry, and drawing an emission boundary is the collector's decision. It also means
an adapter could do the same to reach syntax outside a callback, and Rust cannot prevent
that. Doing so takes a direct dependency in `Cargo.toml` plus a visible
`impl RustEmitter`, and a step in the `build` job of the Rust workflow greps the two
in-tree adapters for such an impl and fails the run.

## API changes for adapter authors

- On the `Prebindgen` trait an adapter implements, the five emission callbacks,
  `prerequisites` and `post_process_item` are handed `&prebindgen_registry::Emit`, as is
  the closure `RegistryBuilder::convert_with` takes.
  `prebindgen_flat::Emit` no longer exists; `prebindgen_registry::Emit`, previously a
  re-export of it, is now the registry's own type.
- The generic `emit.shape(item, head, parts)` splits into `shape_struct`,
  `shape_alternative` and `shape_enum_value`, because a generic method cannot appear on
  an object-safe trait. The `prebindgen-c` and `prebindgen-jni` call sites move to the
  named methods.
- `Emit::member` and `Emit::bind` are dropped. Both forwarded to `Field::member` and
  `Field::bind`, which are public because they read the field's name and index rather
  than syntax, and no call site went through `Emit`.
- `prebindgen-flat`'s `testing` feature is removed. `Emit::for_test` moves to
  `prebindgen-registry`, behind the `testing` feature the adapter test suites already
  enable for the `Registry` test constructors.

## Other cleanup

- `UnfoldError::ReturnTypeMismatch`'s `actual` field now carries the canonical type
  identity that `Display for TypeRef` produces, matching the already-canonical `declared`
  field beside it, instead of the captured spelling.
- `Origin`, the wrapper pairing a `prebindgen-flat` node with the syntax it was built
  from, keeps its
  derived `Debug`, documented as diagnostic output rather than a structured route to that
  syntax.
- Registry tests that assert on spellings go through a test-only `SpellForTest`
  extension trait over `Emit::for_test`, so they exercise the same key production
  callbacks receive.
- `docs/crate-split.md` and the README crate table describe the protocol/key boundary,
  and the split's phase table no longer carries #375 as an open follow-up.

## Validation

Run on the branch head, all green:

- `cargo test --workspace --all-features` — unit, integration and doctests, including the
  compile-fail ones this change adds
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `cargo fmt --all --check` with the workflow's rustfmt config
- `RUSTDOCFLAGS=-Dwarnings cargo doc -p prebindgen-flat -p prebindgen-registry --all-features --no-deps`
- The CI guard: its `git grep` finds nothing on the branch, and matches once an
  `impl RustEmitter` is planted in `prebindgen-jni`

Closes #375


milyin:
@copilot resolve the merge conflicts in this pull request

milyin (commented):
Re-reviewed current head 1b0afc5 after the merge-conflict resolution. No blocking findings.

The follow-up fully addresses the earlier review: flat::emit is private while the direct flat collector API remains available at the crate root; the registry deliberately omits the protocol re-export; Emit(()) now tests field privacy directly; the adapter policy guard is in CI; and the documentation accurately separates the compiler-checked default path from the explicit direct-flat-dependency residual. The merge resolution also correctly retained both SpellForTest and the new capture-directory helper in test_util.rs.

Validation on the current head:
- cargo test --workspace --all-features — green, including all new compile-fail doctests
- cargo clippy --workspace --all-features --all-targets -- --deny warnings — green
- git diff --check origin/main...HEAD — clean
- generated artifacts leave the worktree clean
- GitHub Rust workflow — green

Looks ready to merge.

— Codex (GPT-5)

.github/workflows/rust.yml
```diff
@@ -37,6 +37,13 @@ jobs:
     - uses: actions/checkout@v4
       with:
         path: prebindgen
+    - name: Keep Rust emission authority in registry callbacks
+      run: |
+        if git grep -n -E "impl .*RustEmitter for" -- prebindgen-c prebindgen-jni; then
```
milyin: This guard is the only enforcement behind the PR description's residual ("a step in the `build` job of the Rust workflow greps the two in-tree adapters for such an impl and fails the run"), and two things get past it.

The regex requires `impl`, the trait name and `for` on one source line. rustfmt wraps an impl header that exceeds the width limit, so an adapter written as

```rust
impl<M: SomeLongBound>
    prebindgen_flat::RustEmitter for JniGen<M> {}
```

passes the step. I ran both forms against this branch inside `prebindgen/`: the single-line `impl prebindgen_flat::RustEmitter for Foo {}` is caught (exit 0), the wrapped form is not (exit 1, step green).

Second, the pathspec names `prebindgen-c` and `prebindgen-jni` literally. `git grep` exits 1 — indistinguishable from "no match", so the step passes — when a pathspec matches no tracked file. Renaming or moving either adapter crate turns the guard into a silent no-op instead of a failure.

Both evasions still need the prerequisite the PR description names: a direct `prebindgen-flat` entry in the adapter's `Cargo.toml`, since neither adapter depends on that crate today. Grepping the two manifests for that dependency name is one line, is immune to how the `impl` is formatted, and fails loudly rather than silently if a manifest path stops existing.

prebindgen-flat/src/flat/mod.rs
```diff
@@ -638,29 +638,6 @@ impl Flat {
...
-    #[allow(dead_code)]
-    pub fn enum_item<N: Name + ?Sized>(&self, name: &N) -> Option<&syn::ItemEnum> {
-        match self.declared_type(name)? {
-            Type::Variant(v) => Some(v.origin.as_syn()),
-            Type::Enum(e) => Some(e.origin.as_syn()),
-            _ => None,
-        }
-    }
-
     /// Module name of every captured source, in first-seen order.
```
milyin: Deleting `Flat::enum_item` here leaves three comments in `prebindgen-jni` that still describe it as a live method on `Flat`.

- `prebindgen-jni/src/jni/emit/sum_out.rs:238` — "`enum_item` hands back only the `syn::ItemEnum`, deliberately"
- `prebindgen-jni/src/jni/emit/flat_input.rs:1067` — the same sentence
- `prebindgen-jni/src/jni/symbols.rs:317` — "`enum_item` answered for both shapes by handing back the whole `syn::ItemEnum`"

The first two are present tense, so a reader who follows them to check the claim looks for a method that no longer exists. The third is already past tense and reads correctly. Rewording the first two the way the third is worded, or pointing them at `declared_type`, keeps them checkable.

PR #402, opened 2026-08-13
Umbrella: the shape matrix — enumerate what crosses, per position and target (#198)

milyin:
This body mirrors [`docs/shape-matrix.md`](https://github.com/milyin/prebindgen/blob/shape-coverage/docs/shape-matrix.md) on this branch, which is the one place step state is edited. Change the doc, then re-sync this body — never the other way round.

**Draft, and long-lived.** Child PRs target **`shape-coverage`**, not `main`.

Umbrella for [#198](https://github.com/milyin/prebindgen/issues/198), itself part of [#399](https://github.com/milyin/prebindgen/issues/399).

## The problem

Which Rust shapes survive the trip to C or to Kotlin — `Option<&T>`, `Vec<Option<T>>`, a payload-carrying enum inside a struct — was knowable only by reading the generators, or by writing the Rust and finding out. A user discovered a gap by hitting it; a reviewer could not tell whether a change closed a hole or moved it; and a regression that quietly dropped a shape looked like nothing at all, because no test knew the shape was supported.

The answer is a **generated table** — one row per (shape, position, target), the answer produced by running the real generators. There is deliberately no second authority on legality anywhere in this work: a second opinion can disagree with the first, and then neither is trustworthy.

## What is on the branch now

`examples/shape-matrix` enumerates **412 cells** (206 per target) across four axes and runs each through the real `prebindgen-c` and `prebindgen-jni`, then through rustc, then through cbindgen or the Kotlin compiler, and — for three of them — through actual execution.

| Target | header | kotlin | rustc | bad kotlin | bad rust | rejected | panic | n/a |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| C | 84 | 0 | 0 | 6 | 0 | 51 | 37 | 24 |
| Kotlin/JNI | 0 | 118 | 0 | 1 | 1 | 48 | 10 | 24 |

Each target ends at its own last stage, and the report says so rather than levelling to the shorter ladder: a C cell goes on to cbindgen and tops out at `header`, a JNI cell goes on to the Kotlin compiler and tops out at `kotlin`.

**Four axes:** the value shape in a position; the **call** (aliasing is a property of a call, not of a value); the **declaration policy** (the same Rust declared as something else); and the target.

**Four gates**, each of which fails rather than asks a reviewer to notice:

- the **type axis is derived**, not listed — `tag_of` is an exhaustive match over `TypeKind`, so a new accepted Rust form stops the crate compiling until it has a fixture;
- the **declaration axis** likewise, over `prebindgen_jni::ClassDecl`;
- **`REPORT.md` is committed** and diffed by `regen-check.sh`, so every changed answer is a reviewed diff;
- **`GUARANTEES.md`** holds a floor per cell so a cell may rise freely and cannot fall silently. Raising is automatic; lowering is a hand edit, on purpose.

## What it found

The table was never the deliverable. The tickets are, and **every one is fixed**:

| Issue | Finding | Fixed by |
|---|---|---|
| [#410](https://github.com/milyin/prebindgen/issues/410) | JNI emits an unqualified `Cow` into the consumer's scope | [#423](https://github.com/milyin/prebindgen/pull/423) |
| [#411](https://github.com/milyin/prebindgen/issues/411) | JNI decodes an exclusive-borrow parameter as a shared or plain value | [#426](https://github.com/milyin/prebindgen/pull/426) |
| [#412](https://github.com/milyin/prebindgen/issues/412) | C moves out of a raw pointer for an `Option<T>` by-value parameter | [#425](https://github.com/milyin/prebindgen/pull/425) |
| [#413](https://github.com/milyin/prebindgen/issues/413) | C returns of borrowed elements: `&[T]` drops the value, `Vec<&T>` maps over an `unsafe fn` | [#427](https://github.com/milyin/prebindgen/pull/427) |
| [#414](https://github.com/milyin/prebindgen/issues/414) | C qualifies std `Option` into the source module, and leaves the declared type bare | [#424](https://github.com/milyin/prebindgen/pull/424) |
| [#429](https://github.com/milyin/prebindgen/issues/429) | JNI's sum-payload builder converts the wire value as if it were the leaf | [#432](https://github.com/milyin/prebindgen/pull/432) |
| [#430](https://github.com/milyin/prebindgen/issues/430) | JNI's `data_class` factory builds a handle through the constructor #404 made private | [#431](https://github.com/milyin/prebindgen/pull/431) |
| [#433](https://github.com/milyin/prebindgen/issues/433) | JNI declares a boxed slot for an `Option<handle>` payload the encoder writes as a primitive | [#434](https://github.com/milyin/prebindgen/pull/434) |
| [#428](https://github.com/milyin/prebindgen/issues/428) | C calls the composite marker for an `Option<&T>` **callback argument** | [#435](https://github.com/milyin/prebindgen/pull/435) |
| [#437](https://github.com/milyin/prebindgen/issues/437) | C builds an array for a `Vec` **callback argument** and never emits the helper — the prerequisite gate reads a function's return type only | open |
| [#438](https://github.com/milyin/prebindgen/issues/438) | JNI's **callback** adapter skips a value's `Option` and collection layers — #429 in the direction #432 did not fix | open |
| [#191](https://github.com/milyin/prebindgen/issues/191) | a third of C's refusals and a fifth of JNI's arrive as panics — a measured number, not a defect | open |

Beyond the eight cells those two new tickets account for, **one cell** is short of the top of its target's ladder for a reason that is not a defect: `unit_enum__ret_as_ptr__jni`, where a fieldless enum declared as a handle trips the generated alignment assertion, because a two-variant enum is 1-aligned and bit 0 is the closed tag. The refusal is right and the vocabulary is what is wrong — it arrives as a compile error in the consumer's build rather than as a rejection at declaration time, which is #191's subject.

### Where the findings came from, and what that says

Two of the nine were found by a cell. **Four were found by a stage added later**, over cells that were already passing: the compiler found #411 and #412, the Kotlin compiler found #429 and #430. That is the argument for building stages rather than more cells, and it is the same result every stage of this work has produced.

The last two were found by **neither**, and each names something the crate cannot see:

- **#433** came from a JVM run. A Kotlin `external fun` and a Rust `#[no_mangle]` symbol are checked against each other by nothing until a JVM looks one up — which is what step **2e** is for, and why it is next rather than optional.
- **#428** came from a hand-written fixture. A **callback argument** was not one of the four positions this crate enumerated, so the C column read 0 broken cells for two merges while that defect was live. It is the mirror image of #413, which the matrix *did* find in return position — the same composite, lowered structurally one way and left to a marker the other.

  That position is step **7**, and it has landed. On its first run it found **#437** and **#438** — the second being #429 in the one direction #432 did not reach, because two emitters each carry their own copy of the rule. Which is the axis's argument restated: a position the corpus omits is a hole no other axis can cover for it.

Three behavioural claims are **established by running the code**, not by inspecting it: the alias guard fires on a call naming one resource twice, spares one that does not, and runs *before* ownership moves — the last proven by reclaiming the resource afterwards, which would be a double free if a converter had run.

## Steps

| # | Step | State |
|---|---|---|
| 1 | The enumerator, both targets, committed report + regen gate | **landed** (#400) |
| 2 | Receipts: rustc accepts the emitted Rust | **landed** (#403) |
| 2b | Kotlin emitted, and cbindgen asked for the header | **landed** (#405) |
| 2c | The runtime — C side, in pure Rust | **landed** (#409) |
| 2d | The Kotlin compiler | **landed** |
| 2e | The JVM runtime | not started — needs a cdylib per cell and a harness to enter it |
| 7 | The callback-argument position | **landed** |
| 3 | The guarantee ratchet — a floor per cell | **landed** (#406) |
| 4 | The call axis | **landed** (#407) |
| 5 | The declaration-policy axis | **landed** for the JNI vocabulary (#408); C's own declarators blocked on #192 |
| 6 | Plan-level invariants | blocked on #192 / #193 |

Step 2d turned out to be a smaller commitment than it looked: the two CI jobs that run the matrix install a pinned `kotlinc`, and each cell's Kotlin gets its own package so one compiler pass covers the corpus. What remains is **2e** — a JNI wrapper is entered from a JVM, which needs each cell built as a cdylib and loaded, where the C side's wrappers were ordinary Rust functions a `cargo test` could call — and the plan-layer rework (**5-C**, **6**).

2e is also the only stage that can settle the claim neither compiler makes: that the two halves agree about a signature. #433 is that claim failing, found by hand because no cell runs — so 2e is no longer a hypothetical.

**7** was the other kind of gap — not a missing stage over the corpus but a missing *position in it* — and it is now landed: 40 new cells, two defects on the first run, every stage above it applying unchanged.

## A recurring result worth stating

Every stage that added a *stronger* kind of evidence immediately invalidated some of the weaker evidence beneath it — usually this harness's, not the generator's:

- the compile stage found three fixtures that had been reporting confident wrong answers, including borrow-returning cells that could never have compiled;
- the runtime stage found that a `()`-returning C binding **aborts** on an alias rejection, because a panic cannot cross `extern "C"` — so the guard is unobservable in-process, and the fallible spelling is the only one whose rejection a caller can act on;
- the policy axis found that one earlier finding belonged to the by-value path rather than to `Option`;
- the Kotlin compiler found four cells sitting at the top of JNI's ladder whose Kotlin is not a program;
- a JVM run — outside this crate, in `covertest-kotlin` — found a cell at the *new* top whose two halves disagree about a signature (#433), which is the evidence for 2e;
- and the callback-argument position found two defects the other four positions could not see, one of which is a defect this branch had already reported and `main` had already fixed, in the direction nobody had a cell for (#438).

That is the argument for the receipt rule this work has followed throughout: a state that nothing mechanical established can only be wrong in the flattering direction.

## Exits

Each step states which it is, up front:

- **answer-preserving** — no cell state changes. A refactor of a measuring instrument must not move the measurements.
- **answers move** — the diff **is** the review, and every moved cell is named with its cause, because a moved cell is either a capability change or a regression and the diff alone cannot say which.

A blanket "the report never changes" would be false on contact and would train reviewers to wave the diff through.


milyin:
Closing as superseded. Its deliverable landed; its vehicle will not.

**The findings were the point, and they are done.** This umbrella's own table
lists eleven. Ten are fixed and closed — #410, #411, #412, #413, #414, #429,
#430, #433 and #428 by the merged PRs it names, and #438 separately. The last,
#437, was fixed by #676 replacing the return-type-only prerequisite scan with
the generation plan's own requirement mechanism, and is closed today with a test
recovered from #448 that covers the `vec_scalar` cell it named first. #191 is a
measurement rather than a defect and stays open on its own.

**What does not land is `examples/shape-matrix`** — the 4,848-line harness that
enumerates 412 cells, together with `REPORT.md`, `GUARANTEES.md` and the ratchet.
Its own tests check the harness, not the generators, so there was nothing in it
to recover as a test case for `main`.

The goal it serves, #198, stays open: knowing which shapes cross in which
position for each target is still answerable only by reading the generators. A
successor that wants it should be written against the current tree — `main` is
several hundred commits past this branch's base, and #676 moved crossing
enumeration, site enumeration and artifact ordering into the registry, which is
where a matrix would now hook in. The branch is left in place if the corpus is
worth mining.

— Claude Code (Opus 5)

PR #423, opened 2026-08-18
Spell Cow and MaybeUninit in full in generated Rust (#410)

### LINKED ISSUE #410: jnigen: a Cow in a converter signature is emitted unqualified, so the consumer may not compile it
Found by the shape matrix (`examples/shape-matrix`, #198). Three cells: `cow_str__param__jni`, `cow_str__field__jni`, `cow_str__payload__jni`.

## What happens

A `Cow<'a, T>` crossing the JNI boundary is emitted **unqualified** into the generated file:

```rust
pub(crate) unsafe fn JString_to_Cow_static_str_47272020<'env, 'v>(
    …
) -> ::core::result::Result<Cow<'static, str>, __JniErr> {
```

The generated file is `include!`d into a consumer crate. Unless that crate happens to have `use std::borrow::Cow` in scope at the include site, it does not compile:

```
error[E0412]: cannot find type `Cow` in this scope
```

Every other type in the same file is fully qualified (`::core::result::Result`, `::std::string::String`), so this is an omission rather than a policy.

## Why it was not noticed

`Cow` crosses nowhere in the in-tree examples, and the adapter's unit tests assert on emitted **text**, which contains the right substring either way. The generated file compiles in `zenoh-flat-jni` only if that crate's imports happen to cover it.

## Reproduce

```
cargo run -p shape-matrix
# then read the emitted file the matrix kept:
target/shape-matrix-check/cells/src/cow_str__param__jni.rs
```

## Suggested fix

Spell it `::std::borrow::Cow` where the converter's signature is built. Worth checking the same emission path for other prelude-adjacent names that are qualified by habit rather than by rule.

milyin:
Fixes #410, found by the shape matrix on #402.

## What was wrong

`Normalization::PRELUDE` (`prebindgen-flat/src/flat/spelling.rs`) is the set of constructor paths ingest reduces to a bare name, so that `std::borrow::Cow<'a, str>` and `Cow<'a, str>` index as one type. Its own doc says it is deliberately not identical to Rust's prelude: it adds `MaybeUninit`, which the grammar recognises for out-parameters, and `Cow`, which it treats as transparent.

`TypeRef::spell` — the tokens generated Rust says — returned that reduced spelling unchanged. For `Vec`, `Option`, `Result`, `String` and `Box` the bare name resolves in any crate. For the other two it resolves only where the consumer happens to have imported them, and a generated file is `include!`d into a crate that need not have:

```rust
pub(crate) unsafe fn JString_to_Cow_static_str_47272020<'env, 'v>(
    …
) -> ::core::result::Result<Cow<'static, str>, __JniErr> {
```

```
error[E0412]: cannot find type `Cow` in this scope
```

## The fix

`qualify_for_emission` re-qualifies exactly the two names Rust does not pre-declare, and runs at `TypeRef::spell` and `TypeRef::stripped_syntax` — the two doors every emitted type spelling comes out of, for both the C and the JNI adapter. Every other spelling stays the source crate's own tokens, which is what makes a generated signature name a type the way the source crate does.

Fixing it at the spelling door rather than in the JNI converter builder is what makes it one fix: the same reduction feeds `prebindgen-c`, and `MaybeUninit` reaches emission by the same route `Cow` does.

Two arms in `prebindgen-jni/src/jni/trait_impl.rs` already hand-normalized a `Cow` output converter's parameter type. They are left alone: they also pin the lifetime to `'_`, and removing them would rename converters in `zenoh-flat-jni`'s committed bindings for no behaviour change.

## Verification

- New test `a_spelling_qualifies_what_rusts_prelude_does_not_declare` (`prebindgen-flat/src/flat/tests/roundtrip.rs`): a `Cow`, a `std::borrow::Cow`, a `MaybeUninit` and an `Option<Vec<String>>` parameter, asserting the first three come out qualified and the fourth unchanged.
- `examples/regen-check.sh` reports the in-repo generated output byte-identical, so no committed artifact moves.
- Applied to a checkout of the `shape-coverage` branch and re-ran `cargo run -p shape-matrix`: `cow_str__param__jni`, `cow_str__field__jni` and `cow_str__payload__jni` rise from `bad rust` to `rustc`, and no other cell moves. That report lives on `shape-coverage`, so it is not updated here.

### DIFF
--- a/prebindgen-flat/src/flat/spelling.rs
+++ b/prebindgen-flat/src/flat/spelling.rs
@@ -273,3 +273,47 @@ fn reduce_flat_path(path: &mut syn::Path, against: &Normalization) {
         path.segments = std::iter::once(last).collect();
     }
 }
+
+/// The [`Normalization::PRELUDE`] entries the **language** does not
+/// pre-declare, and the paths that always resolve.
+///
+/// Normalization reduces every prelude constructor to its bare name, which is
+/// what keys agree on. For `Vec`/`Option`/`Result`/`String`/`Box` that name is
+/// in scope in any crate, so a generated signature may spell it. `Cow` and
+/// `MaybeUninit` are in this model's prelude and not in Rust's — see
+/// [`Normalization::PRELUDE`]'s doc, which says so — and generated Rust is
+/// `include!`d into a consumer that need not have imported either.
+const EMISSION_PATHS: &[(&str, &str)] = &[
+    ("Cow", "::std::borrow::Cow"),
+    ("MaybeUninit", "::std::mem::MaybeUninit"),
+];
+
+/// Undo the reduction for [`EMISSION_PATHS`], so a spelling compiles wherever
+/// the generated file is included.
+///
+/// The emission-side inverse of [`normalize_type`], and deliberately only for
+/// the names whose reduced form a consumer crate may not have in scope: every
+/// other spelling stays the source's own, which is what makes a generated
+/// signature name the type the way the source crate does.
+pub(crate) fn qualify_for_emission(ty: &syn::Type) -> syn::Type {
+    use syn::visit_mut::VisitMut;
+    struct Qualifier;
+    impl VisitMut for Qualifier {
+        fn visit_type_path_mut(&mut self, tp: &mut syn::TypePath) {
+            if tp.qself.is_none() && tp.path.segments.len() == 1 {
+                let seg = tp.path.segments.first().expect("len checked");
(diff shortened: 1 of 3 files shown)


milyin (commented):
No blocking findings.

The qualification is correctly applied at the shared TypeRef emission exits, preserves generic arguments and nested types, and does not change registry identity because TypeKey canonicalizes these paths. The regression test covers bare and qualified Cow, MaybeUninit, and unchanged Rust-prelude types.

Verified at 9633c6b8:
- focused regression test passes
- cargo test --workspace --all-features passes
- generation leaves the checkout clean

— Codex (GPT-5)

Issue #413, opened 2026-08-13
cbindgen: returning &[T] emits a wrapper that returns () and calls a placeholder; Vec<&T> builds a map over an unsafe fn


milyin:
Found by the shape matrix (`examples/shape-matrix`, #198). Two cells: `slice_scalar__ret__c`, `vec_ref__ret__c`.

## `&[T]` return — the wrapper drops the value and calls a placeholder

For `pub fn probe() -> &'static [u64]`, the C wrapper is emitted as:

```rust
pub unsafe extern "C" fn probe() -> () {
    let __v = flat::probe();
    let __ret: ();
    __ret = __cbg_outmark_slice_u64(__v);
    __ret
}
```

Two things at once: the exported function **returns `()`** — a C caller receives nothing at all — and `__cbg_outmark_slice_u64` takes no arguments, so it does not compile:

```
error[E0061]: this function takes 0 arguments but 1 argument was supplied
```

The `outmark` symbol reads like a marker left for a converter that was never generated. The compile error is fortunate: without it this would be a wrapper that silently discards its return value.

## `Vec<&T>` return — the element converter is not a closure of the right shape

```
error[E0277]: expected a `FnMut(&flat::Handle)` closure,
              found `unsafe fn(&Handle) -> *const handle {__cbg_out_ref_Handle}`
error[E0599]: the method `collect` exists for `Map<IntoIter<&Handle>, unsafe fn(&…) -> …>`,
              but its trait bounds were not satisfied
```

The per-element converter is `unsafe`, and an `unsafe fn` item is not a `FnMut`, so the `map`/`collect` the emission builds cannot type-check.

## Filed together

Both are **returns of borrowed elements** on the C side, and both look like the same unfinished corner rather than two independent slips — but they are distinct mechanisms, so split this if the fixes diverge.

## Reproduce

```
cargo run -p shape-matrix
target/shape-matrix-check/cells/src/slice_scalar__ret__c.rs
target/shape-matrix-check/cells/src/vec_ref__ret__c.rs
```

PR #424, opened 2026-08-18
Qualify a prelude constructor and its arguments apart (#414)

### LINKED ISSUE #414: cbindgen: a sum payload's drop qualifies std Option into the source module, and leaves the declared type bare
Found by the shape matrix (`examples/shape-matrix`, #198). Cell: `opt_handle__payload__c`.

## What happens

The drop emitted for an `Option<Handle>` payload qualifies **`Option`** — a std type — into the source module:

```rust
drop(::std::boxed::Box::from_raw(*__f0 as *mut flat::Option<Handle>));
```

`flat` is the source crate, which declares no `Option`, so:

```
error[E0412]: cannot find type `Option` in module `flat`
error[E0412]: cannot find type `Handle` in this scope
```

The second error is the same slip from the other side: `Handle` **is** in the source module and is spelled bare.

So one expression gets both halves of the qualification wrong, in opposite directions — which suggests the payload-drop path builds its type from a name plus a fixed prefix rather than from the model's reading, where the distinction between a declared type and a builtin is already recorded.

## Reproduce

```
cargo run -p shape-matrix
target/shape-matrix-check/cells/src/opt_handle__payload__c.rs
```

milyin:
Fixes #414, found by the shape matrix on #402.

## What was wrong

`Cbindgen::src_ty` (`prebindgen-c/src/builder.rs`) resolves a captured type's spelling for generated Rust: it prefixes a bare single-segment path with the declared `.source_module(…)`, so `ZKeyExpr` becomes `zenoh_flat::ZKeyExpr`. Ingest normalizes `std::option::Option<T>` to `Option<T>`, so by the time an emitter spells a type, a prelude constructor and a source-crate item look alike — both are one segment.

`src_ty` treated `Option<Handle>` as the latter and did not descend into the arguments, so the typed drop for an `Option<Handle>` payload came out as:

```rust
drop(::std::boxed::Box::from_raw(*__f0 as *mut flat::Option<Handle>));
```

```
error[E0412]: cannot find type `Option` in module `flat`
error[E0412]: cannot find type `Handle` in this scope
```

One expression, both halves wrong in opposite directions: `Option` was qualified into a module that does not declare it, and `Handle`, which that module does declare, was left bare.

## The fix

`src_ty` now qualifies the arguments first and separately — they resolve against a different root than the constructor does — and then routes the constructor itself. The new `prelude_path` helper (`prebindgen-c/src/lib.rs`) names the seven constructors ingest reduces to a bare name and gives each its full path, so they resolve against the language rather than against the source module. Everything still one segment after that is the source crate's, and is prefixed as before.

The paths are spelled in full (`::core::option::Option`) rather than left bare, for the reason `is_string` already spelled `::std::string::String`: a generated file is `include!`d into a consumer crate, and `MaybeUninit` is not in Rust's prelude.

`is_string`'s early return is left in place — it matches a multi-segment `std::string::String` too, which the single-segment path below it does not reach.

## Verification

- New test `a_payload_drop_qualifies_the_language_and_the_source_crate_apart` (`prebindgen-c/src/tests/tagged_unions.rs`): a `tagged_union` with an `Option<Handle>` payload over an `opaque_ptr` `Handle`, asserting the drop names `::core::option::Option<myflat::Handle>`. It fails on `main` with the old spelling.
- `examples/regen-check.sh` reports the in-repo generated output byte-identical, so no committed artifact moves.
- Applied to a checkout of the `shape-coverage` branch and re-ran `cargo run -p shape-matrix`: `opt_handle__payload__c` rises from `bad rust` to `header`, C's top state, and no other cell moves. That report lives on `shape-coverage`, so it is not updated here.

### DIFF
--- a/prebindgen-c/src/builder.rs
+++ b/prebindgen-c/src/builder.rs
@@ -598,41 +598,60 @@ impl CbindgenBuilder {
 
     // ── Internal helpers ───────────────────────────────────────────────
 
-    /// Fully-qualify a bare single-segment source type against
+    /// Fully-qualify every bare single-segment source type inside `ty` against
     /// [`Self::source_module`] (e.g. `ZKeyExpr` → `zenoh_flat::ZKeyExpr`).
     /// Anything already qualified, or with no `source_module` set, is returned
     /// unchanged.
+    ///
+    /// **Every** path node in the type, not the outermost one: a source type
+    /// can sit under any accepted shape — `Option<&Handle>` in a callback
+    /// argument, `&[Payload]`, `(Rec, Rec)` — and a name left bare there does
+    /// not resolve in the consumer crate the generated file is included into
+    /// (#414). The walk is what makes one rule apply everywhere; the rule per
+    /// node is below.
     pub(super) fn src_ty(&self, ty: &syn::Type) -> syn::Type {
+        use syn::visit_mut::VisitMut;
+        struct Qualifier<'a>(&'a CbindgenBuilder);
+        impl VisitMut for Qualifier<'_> {
+            fn visit_type_path_mut(&mut self, tp: &mut syn::TypePath) {
+                self.0.qualify_path(tp);
+                // The node's own arguments, after it: `Option<Handle>` is the
+                // language's `Option` over the source crate's `Handle`, and the
+                // two halves resolve against different roots.
+                syn::visit_mut::visit_type_path_mut(self, tp);
+            }
+        }
+        // A trait bound is a `syn::Path` and not a `Type::Path`, so overriding
+        // only the type-path visit leaves `impl Fn(..) + Send` alone and still
+        // reaches the argument types written inside it.
+        let mut out = ty.clone();
+        Qualifier(self).visit_type_mut(&mut out);
+        out
+    }
+
+    /// One path node of [`Self::src_ty`]: what root it resolves against.
+    fn qualify_path(&self, tp: &mut syn::TypePath) {
+        if tp.qself.is_some() {
+            return;
+        }
+        let as_ty = syn::Type::Path(tp.clone());
         // Built-in scalar primitives (`f64`, `i32`, …) live in no source module;
         // qualifying them would produce invalid paths like `zenoh_flat::f64` (hit by
         // callback args, e.g. `impl Fn(f64)`). Leave them bare.
-        if is_scalar(ty) {
-            return ty.clone();
+        if is_scalar(&as_ty) {
+            return;
         }
         // Std `String` likewise lives in no source module — qualifying it would
         // produce `zenoh_flat::String`. It can be declared `opaque_ptr` (a boxed
         // pointer the C side holds as `string_t *`), so resolve it to the std path.
-        if is_string(ty) {
-            return syn::parse_quote!(::std::string::String);
+        if is_string(&as_ty) {
+            tp.path = syn::parse_quote!(::std::string::String);
+            return;
         }
-        let syn::Type::Path(tp) = ty else {
-            return ty.clone();
-        };
-        if tp.qself.is_some() || tp.path.leading_colon.is_some() || tp.path.segments.len() != 1 {
-            return ty.clone();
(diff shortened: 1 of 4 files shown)


prebindgen-c/src/builder.rs
```diff
@@ -615,15 +615,41 @@ impl CbindgenBuilder {
...
+        }
+        // The arguments are qualified first, and separately: `Option<Handle>`
+        // is the language's `Option` over the source crate's `Handle`, so the
+        // two halves resolve against different roots and a single prefix over
+        // the whole path gets one of them wrong whichever way it goes (#414).
+        let mut seg = tp.path.segments[0].clone();
+        if let syn::PathArguments::AngleBracketed(args) = &mut seg.arguments {
+            for arg in args.args.iter_mut() {
+                if let syn::GenericArgument::Type(t) = arg {
+                    *t = self.src_ty(t);
```
milyin: The recursive call only qualifies an argument when that argument is itself a path. A supported callback such as impl Fn(Option<&Handle>) reaches src_ty_deep_of as Optional, falls back to src_ty_of, and is emitted as ::core::option::Option<&Handle>; the Handle beneath the reference is never prefixed with myflat::. Both the returned impl Fn type and the closure parameter then fail with E0412 unless the consumer happens to import Handle. The same gap applies to prelude constructors containing other non-path shapes. Please make qualification descend through every accepted type shape, or have src_ty_deep_of recurse through the prelude wrappers, and add the callback case as regression coverage.

— Codex (GPT-5)

milyin: Confirmed and fixed in 11aab392.

I reproduced it exactly as described: `impl Fn(Option<&Handle>)` came out as `impl Fn(::core::option::Option<&Handle>)`, in both the returned `impl Fn(..)` and the closure's own parameter ascription.

Rather than adding wrapper arms to `src_ty_deep_of`, `src_ty` now walks the whole type with a `VisitMut` and applies one rule per path node. That covers references, slices, arrays and tuples in one change instead of a fixed list of prelude constructors, and it keeps the rule (scalar → bare, `String` → std path, prelude constructor → its own path, anything else single-segment → the source module) in one place. Overriding only `visit_type_path_mut` leaves a trait bound alone, since a bound's path is a `syn::Path` and not a `Type::Path`, so `impl Fn(..) + Send + Sync` keeps its bounds and still has its argument types qualified.

Regression coverage is the callback case: `callback_arg_qualifies_a_source_type_under_a_wrapper` in `prebindgen-c/src/tests/callbacks.rs` asserts both spellings name `::core::option::Option<&zenoh_flat::Handle>`.

Separately, while reproducing this I hit a **second** defect in the same shape, which this PR does not address: the trampoline calls `__cbg_outmark_option___Handle(__a0)`, and that marker takes no arguments. It is the callback-argument twin of the `&[T]` return in #413 — a composite marker used where the structural lowering is missing. I'll file it unless you'd rather it ride along here.

prebindgen-c/src/builder.rs
```diff
@@ -615,15 +615,41 @@ impl CbindgenBuilder {
...
+                }
```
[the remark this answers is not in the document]
milyin: Same finding as the thread above, and fixed the same way in 11aab392: `src_ty` now walks the whole type with a `VisitMut` instead of recursing only into path-shaped generic arguments, so a source type under a reference, slice, array or tuple is qualified too. `callback_arg_qualifies_a_source_type_under_a_wrapper` covers the `impl Fn(Option<&Handle>)` case.

milyin (commented):
Re-reviewed at 11aab392. The previous nested-qualification issue is resolved: the visitor now reaches source types beneath references and other wrappers while keeping each prelude constructor rooted separately. The new Option<&Handle> callback regression covers both the returned impl Fn spelling and the closure parameter spelling. I found no remaining blockers.

Validation: cargo test -p prebindgen-c (75 passed), and all required CI checks are green.

— Codex (GPT-5)

PR #427, opened 2026-08-18
Emit a working C return for a slice and for a Vec of borrows (#413)

### LINKED ISSUE #413: cbindgen: returning &[T] emits a wrapper that returns () and calls a placeholder; Vec<&T> builds a map over an unsafe fn

milyin:
Fixes #413, found by the shape matrix on #402. Two returns of borrowed elements, two mechanisms, both in `prebindgen-c`.

## `&[T]` — the wrapper returned nothing and called a marker

A shared slice already had an output entry: a marker function with a `()` destination, which exists so that a slice **callback argument** resolves while `prereq_callback_structs` lowers it structurally as a pointer/length pair. Nothing claimed the shape in **return** position, so the wrapper took the base-value path and inherited that `()`:

```rust
pub unsafe extern "C" fn probe() -> () {
    let __v = flat::probe();
    let __ret: ();
    __ret = __cbg_outmark_slice_u64(__v);
    __ret
}
```

The C caller receives nothing, and the marker takes no arguments (`error[E0061]`). The compile error is the fortunate half: without it this is a wrapper that discards its return value.

`lower_shape` and `encode_value` already lower `Cow<'_, [T]>` to `T_wire* + size_t`, an owned malloc'd copy the C side frees — the same ABI as a `Vec<T>` return. A shared slice now takes that arm, so `&[u64]` returns `u64*` plus a `len` out-parameter. `lower_shape` and `encode_value` are reached only from the function-wrapper return path, so the callback lowering is untouched.

## `Vec<&T>` — an `unsafe fn` item is not a `FnMut`

```
error[E0277]: expected a `FnMut(&flat::Handle)` closure,
              found `unsafe fn(&Handle) -> *const handle {__cbg_out_ref_Handle}`
```

The borrow converter is `unsafe fn`, and an `unsafe fn` item implements no `Fn` trait, so the `map`/`collect` the sequence arms build cannot type-check. The new `map_arg` helper calls such a converter inside a closure — the call site an `unsafe fn` needs, inside a wrapper body that is already an unsafe context.

A **safe** element converter is still passed to `map` by name. That keeps every sequence return that compiled before byte-identical, which is why `examples/regen-check.sh` still reports no drift.

## What this does not change

`__cbg_out_ref_Handle` produces a non-owning `*const`, so a `Vec<&T>` return hands C an array of pointers that live exactly as long as the referents do. That is the policy a single `&T` return already has; this fix makes the emission compile and leaves the policy alone.

A shared-slice return whose element is not a scalar (a value-opaque element, say) still takes the old base-value path. No cell covers it and no declaration in the tree produces one; it is a separate shape, not a half-applied fix.

## Verification

- Two new tests in `prebindgen-c/src/tests/returns.rs`: `slice_returns_scalar_array` pins the exported signature and the copy for a `&'static [u64]` return, and `vec_of_borrows_calls_the_unsafe_element_converter` pins the closure for a `Vec<&'static ZHandle>` return.
- `cargo test --workspace --all-features` passes, including the two existing tests that assert a safe element converter is passed by name.
- `examples/regen-check.sh` reports the in-repo generated output byte-identical.
- Applied to a checkout of the `shape-coverage` branch and re-ran `cargo run -p shape-matrix`: `slice_scalar__ret__c` and `vec_ref__ret__c` rise from `bad rust` to `header`, C's top state. No other cell moves. That report lives on `shape-coverage`, so it is not updated here.

### DIFF
--- a/prebindgen-c/src/emit.rs
+++ b/prebindgen-c/src/emit.rs
@@ -750,9 +750,17 @@ impl CbindgenBuilder {
                 niches: Niches::empty(),
             };
         }
-        // `Cow<'_, [T]>` → `T_wire* + size_t`. The C side receives an owned
-        // malloc'd copy, just like `Vec<T>` outputs.
-        if let Some(elem) = r_cow_slice_elem(ty) {
+        // `Cow<'_, [T]>` and `&[T]` → `T_wire* + size_t`. The C side receives an
+        // owned malloc'd copy, just like `Vec<T>` outputs.
+        //
+        // A shared slice reaches this only as a RETURN: a slice parameter is a
+        // pointer pair the caller supplies, and a slice callback argument is
+        // lowered by `prereq_callback_structs`. Without the second predicate a
+        // slice return fell through to the base-value path and took the `()`
+        // destination of the marker converter that exists for that callback
+        // lowering, so the wrapper returned nothing and called the marker with
+        // an argument it does not take (#413).
+        if let Some(elem) = r_cow_slice_elem(ty).or_else(|| r_scalar_slice_elem(ty)) {
             let entry = registry.output_entry(elem).unwrap_or_else(|| {
                 panic!(
                     "Cbindgen: `Cow` slice element `{}` has no output converter",
@@ -833,6 +841,7 @@ impl CbindgenBuilder {
         if let TypeKind::Vec(elem) = ty.kind() {
             let entry = registry.output_entry(elem).expect("Vec element converter");
             let elem_conv = entry.function.sig.ident.clone();
+            let elem_map = map_arg(&elem_conv, entry.function.sig.unsafety.is_some());
             let elem_wire = entry.destination.clone();
             let t_ptr = &targets[0];
             let t_len = &targets[1];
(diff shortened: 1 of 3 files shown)


milyin (commented):
No blocking findings.

The shared-slice return now follows the array path consistently through shape lowering, encoding, fallibility detection, and allocator prerequisite selection. Wrapping only unsafe element converters in a closure fixes Vec<&T> while preserving byte-identical emission for safe converters.

Verified at 27c8c041:
- cargo test -p prebindgen-c passes all 75 tests
- both new borrowed-element regressions pass
- all current CI checks are green

Minor diagnostic nit only: the lower_shape failure still says Cow slice element when reached through a shared slice.

— Codex (GPT-5)

PR #431, opened 2026-08-18
Mint an optional handle field through the factory (#430)

### LINKED ISSUE #430: jnigen: a data class's fromParts builds an opaque handle through its private constructor
Found by the shape matrix (`examples/shape-matrix`, #198), by compiling the Kotlin it emits — the stage #402's step 2d adds. Cell: `opt_handle__field__jni`.

## What happens

For `pub struct Probe { pub v: Option<Handle> }` with `Handle` declared `ptr_class!` and `Probe` declared `data_class!`, the generated factory calls the handle's constructor:

```kotlin
public fun fromParts(v: Long): Probe = Probe(if (v == 0L) null else Handle(v))
```

```
error: cannot access 'constructor(initialPtr: Long): Handle': it is private in 'Handle'.
```

## Why the constructor is private

Deliberately, and recently: #404 closed the forged-pointer holes in generated Kotlin, and the emitted `NativeHandle` doc states the rule — *"Handle constructors carry no marker at all: they are `private`, reachable only through a synthetic factory."* The factory is `Handle.fromRawPtr`, which the **payload** path already calls for the same `Option<Handle>` in a sealed class's builder.

So this is one path that was not moved to the factory when the constructors were sealed. The Rust half compiles either way, and the two paths disagree only in the Kotlin.

## Not the same as #429

#429 is the payload builder skipping the layers between the wire value and the leaf; this is the field path naming the wrong constructor for the leaf itself. They are worth fixing together — between them the two paths hold the complementary halves of one correct answer, since this one does the null check that #429's builder omits and #429's builder calls the factory this one does not.

## Reproduce

```
cargo run -p shape-matrix
```

The emitted Kotlin is kept under `target/shape-matrix-kotlin/cells/opt_handle__field__jni/`.

milyin:
Fixes #430, found by the shape matrix on #402 — by its new stage, which compiles the Kotlin the JNI adapter emits.

## What was wrong

For `pub struct Bag { pub handle: Option<Handle> }` with `Handle` declared `ptr_class!`, the generated factory called the handle's constructor:

```kotlin
public fun fromParts(handle: Long): Bag = Bag(if (handle == 0L) null else Handle(handle))
```

```
error: cannot access 'constructor(initialPtr: Long): Handle': it is private in 'Handle'.
```

That constructor is private on purpose. #404 sealed it so a Java or Kotlin caller cannot forge a pointer, and `handle_from_raw` — whose doc says *"every generated site that mints a handle goes through here, so the choice of `HANDLE_FACTORY` over a constructor is made once"* — is the replacement. The `Base` arm of `factory_projection_wire_wrap` already used it; the `Optional` arm immediately below it did not.

## The fix

The `Optional{Handle}` arm calls `handle_from_raw` too, so the absent case still rides the `0L` jlong sentinel and the present case goes through the factory.

## Why nothing caught it

The Rust half compiles either way — this is Kotlin the compiler never saw. `prebindgen-jni`'s tests covered a **non-optional** handle field (`data_class_properties_match_their_from_parts_params` asserts `Bag(Handle.fromRawPtr(handle)`), and the nullable arm had no test at all.

## Verification

- New test `an_optional_handle_field_mints_through_the_factory` (`prebindgen-jni/src/jni/tests/values.rs`) asserts the emitted factory reads `if (handle == 0L) null else Handle.fromRawPtr(handle)`. It fails on `main`.
- `examples/regen-check.sh` reports the in-repo generated output byte-identical, so no committed binding changes: no example declares an optional handle field, which is the same gap the test closes.
- On the `shape-coverage` branch this takes `opt_handle__field__jni` from `bad kotlin` to `kotlin`.

### DIFF
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -218,6 +218,12 @@ fn main() {
         // harness is what ties the two decisions together, since an emission
         // test never compiles the inner class.
         .package(package!().class(data_class!(Dossier)))
+        // #430: the same handle field with an `Option` in front of it. The
+        // present arm mints the handle, and it has to do so through the
+        // generated factory — the constructor is private (#404). Only a
+        // compiled run says which one the factory names, so the shape lives
+        // here rather than only in an emission test.
+        .package(package!().class(data_class!(MaybeHolder)))
         // A data class whose FIELDS carry transparent wrappers (#289 + #292):
         // `boxed: Box<Option<i64>>` must cross exactly as `plain: Option<i64>`
         // does — the decoupled `(present, value)` pair — with the `Box` put back
@@ -557,6 +563,10 @@ fn main() {
                 // …and reached one level deeper still, through a nested data
                 // class rather than a sum.
                 .fun(fun!(dossier_new))
+                // #430: the handle field with an `Option` in front of it. The
+                // return is what runs the factory, so both arms — minted and
+                // absent — are compiled and exercised.
+                .fun(fun!(maybe_holder_new))
                 // #213: the output boundary DERIVED from the type's value form
                 // rather than restated. `report_each` delivers the decomposed
                 // `Report` in one crossing; the leaf list comes from
--- a/examples/covertest-kotlin/kotlin/REPORT.md
+++ b/examples/covertest-kotlin/kotlin/REPORT.md
@@ -94,6 +94,7 @@ Base package: `io.prebindgen.covertest`
   - shaped by: return `Lookup` decomposed → [tag, found_v0, failed_v0] (Callback delivery)
 - `marker_of` — `fun markerOf(which: Int, onError: JniErrorHandler<Marker?>): Marker?`
   - shaped by: return `Marker` decomposed → [tag, ranked_v0] (Callback delivery)
+- `maybe_holder_new` — `fun maybeHolderNew(tag: Long, count: Long, total: Double, present: Boolean, onError: JniErrorHandler<MaybeHolder?>): MaybeHolder?`
 - `object_boundary_value` — `fun objectBoundaryValue(value: ObjectBoundary, onError: JniErrorHandler<Long>): Long`
 - `observation_new` — `fun observationNew(which: Int, withFallback: Boolean, onError: JniErrorHandler<Observation?>): Observation?`
 - `observation_which` — `fun observationWhich(o: Observation, onError: JniErrorHandler<Int>): Int`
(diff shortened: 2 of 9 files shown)


milyin (commented):
One required coverage gap remains:

- prebindgen-jni/src/jni/tests/values.rs:1639 — This regression only inspects emitted text. The defect being fixed is a Kotlin compilation failure, so the test does not compile the generated factory or exercise its present/absent values across JNI. The workspace convention requires every JniGen shape to be represented in examples/covertest-kotlin; please add an Option<Handle> data-class field there and cover both Some and None. This is especially important because PR #402 is still open, so its shape-matrix Kotlin compiler is not protection on main.

The implementation change itself is consistent with the existing non-optional handle path.

Validation: cargo test -p prebindgen-jni (261 passed; doctests 2 passed, 4 ignored), cargo fmt --all --check, git diff --check, and examples/regen-check.sh all pass. The unchanged regeneration output also confirms that the covertest does not currently contain this shape.

— Codex (GPT-5)

milyin:
Coverage gap addressed in ab3ce224 — you were right that the emission test proves nothing here.

`examples/perftest-flat` gains `MaybeHolder`, which is `Holder` with an `Option` between the field and the handle, and `maybe_holder_new(tag, count, total, present)` so one function reaches both arms. It is declared `data_class!(MaybeHolder)` in `covertest-kotlin`'s build script and exercised in `Test.kt`:

- **present** — `summary` is non-null, `count()` returns through it, and closing the container closes it (so the cascade is covered as well as the mint);
- **absent** — `summary` is `null` rather than a handle over pointer 0, and closing the container is a no-op.

The harness runs 54 sections and passes.

The load-bearing check is the negative one: with the fix reverted and everything else identical, the Gradle build fails to compile.

```
e: …/kotlin/generated/io/prebindgen/covertest.kt:272:117 Cannot access '<init>': it is private in 'Summary'
```

So the `covertest` CI job now catches this shape on `main`, independently of #402.

Two notes on the diff. The committed generated artifacts grow — `covertest.kt`, `model.kt` and `generated_bindings.rs` — which is the shape appearing in them for the first time; `regen-check.sh` is clean after the commit. And `Holder`'s own factory is visible one line above `MaybeHolder`'s in `covertest.kt`, which makes the required and optional arms reviewable side by side.

milyin (commented):
Re-reviewed at `ab3ce224`. The previous coverage finding is fully resolved: `MaybeHolder` now reaches the generated `fromParts` factory from a real JVM for both `Some` and `None`, verifies the minted handle, and checks cascading close behavior. I found no remaining blockers.

Validation: `cargo test -p prebindgen-jni` (261 unit tests and 2 doctests passed; 4 doctests ignored), `cargo fmt --all --check`, `git diff HEAD^ HEAD --check`, and `./gradlew run --console=plain` (54 JVM sections) all pass. The generated tree remains clean, and all CI checks are green.

— Codex (GPT-5)

Issue #437, opened 2026-08-19
cbindgen: a Vec callback argument builds an array but the array helper is not emitted


milyin:
Found by the shape matrix (`examples/shape-matrix`, #198) after it gained the **callback-argument** position — the axis #428 came through. Six cells: `vec_scalar`, `vec_record`, `vec_handle`, `vec_ref`, `vec_sum` and `opt_vec`, each in `cbarg` position for C.

## What happens

A `Vec<T>` delivered to a C closure lowers to the same `(ptr, len)` pair a `Vec<T>` return does, and the encode calls the shared array builder:

```rust
move |__a0: ::std::vec::Vec<u64>| {
    if let ::core::option::Option::Some(__f) = __call {
        …
        let (__p, __n) = __cbg_alloc_array(__arr);
```

`__cbg_alloc_array` is not emitted for that binding:

```
error[E0425]: cannot find function `__cbg_alloc_array` in this scope
```

The same cell with the `Vec<T>` in **return** position emits the helper and compiles.

## Why

`produces_array` (`prebindgen-c/src/emit.rs`) decides whether to emit the prerequisite by inspecting each declared function's **return type** only:

```rust
self.functions.keys().any(|orig| {
    registry.flat().function(&orig)
        .map(|f| f.ret.walk().iter().any(|t| t.sequence_elem().is_some()))
        .unwrap_or(false)
})
```

A run of values also reaches C as a **callback argument**, and after #435 that path lowers it structurally — so it calls the builder that the gate never asked about. The gate needs to consider a declared function's parameters, and specifically the callback signatures among them, not only its return.

## Reproduce

```
cargo run -p shape-matrix
target/shape-matrix-check/cells/src/vec_scalar__cbarg__c.rs
```

Compare with `vec_scalar__ret__c.rs`, which is the same run of values one position over and does emit the helper.

## Note

This is reachable *because* #435 gave the callback path a structural lowering; before it, the same cell failed earlier and differently (it called the composite marker). The gap in the prerequisite gate is real either way, and the position that shows it is new — which is the point of adding it.

milyin:
Fixed, and now covered by a test.

**The cause is gone.** This issue named `produces_array` in
`prebindgen-c/src/emit.rs`, which decided whether to emit the array helper by
inspecting each declared function's return type only — so a `Vec<T>` reaching a
C closure built an array and then failed to link against a helper nobody had
emitted. That function no longer exists. #676 replaced the ad-hoc prerequisite
[...]

Issue #444, opened 2026-08-19
prebindgen-registry: finish the adapter-neutral transform stack on top of #443


milyin:
## Dependency

This work is a stacked follow-up to #443. The implementation PR should be based on #443 and reference it explicitly. After #443 merges, retarget or rebase the follow-up onto `main` so the two reviews form one continuous stack.

## Goal

Finish the transformation layer introduced by #443 so Cbindgen, JniGen, and future language adapters can reuse one folding/unfolding mechanism.

The registry must own semantic recursion: optional and sequence layers, products, choices, constructor/deconstructor operations, child order, source paths, cycle handling, and dependency discovery. An adapter must lower that semantic tree once into its own resolved boundary plan. It must not implement another recursive `TypeRef`, shape, path, or leaf-group walk.

This does **not** define a common C/JNI wire representation. C and JNI keep different plan types and different ABI policy.

## Current boundary after #443

#443 adds the correct shared kernel: [`TransformNode<D>` and `TransformLowerer<D>`](https://github.com/milyin/prebindgen/blob/1bff059cbeb7de1e359ad2dd0bfe9b46cc81bc84/prebindgen-registry/src/transform.rs#L28-L182). The registry fold emitter already uses that traversal.

The remaining seams prevent the tree from being the adapter boundary:

- [`IntoRust` nodes](https://github.com/milyin/prebindgen/blob/1bff059cbeb7de1e359ad2dd0bfe9b46cc81bc84/prebindgen-registry/src/expand/tree.rs#L45-L139) still contain a physical flat protocol: slot positions and names, an `i32` selector, a `bool` presence flag, and `Option`-wrapped inactive arguments. [`process_expand`](https://github.com/milyin/prebindgen/blob/1bff059cbeb7de1e359ad2dd0bfe9b46cc81bc84/prebindgen-registry/src/expand.rs#L298-L330) makes those leaf types converter roots before an adapter can select its layout.
[...]

Issue #442, opened 2026-08-19
prebindgen-registry: unify FoldPlan and UnfoldPlan over a direction-typed transformation tree


milyin:
Follow-up to #223 under #399. This is the registry-level foundation that #193/#439 and #436 can consume.

## Goal

Unify constructor folding and value unfolding around one retained, direction-typed transformation tree in `prebindgen-registry`:

```text
TypeRef + constructor/deconstructor declarations
                     |
                     v
       prebindgen-registry transformation recipe
       TransformPlan<IntoRust>
       TransformPlan<OutOfRust>
                     |
          converter resolution
                     |
          +----------+----------+
          |                     |
          v                     v
     CValuePlan<D>         JniValuePlan<D>
     C ABI policy          JNI/JVM policy
          |                     |
          v                     v
     C renderers            JNI/Kotlin renderers
```

The shared plan describes **what happens to the Rust value**. Each adapter plan describes **how that operation crosses its ABI**.

This must not recreate #190's global `SemanticShape` tier. `TypeRef` remains the sole recursive description of Rust syntax. `TransformPlan` is a per-use operation plan which refers to `TypeRef`; it does not classify every source type into a second model.

### FoldPlan mixes construction semantics with one flat foreign protocol

`FoldPlan` already retains much of the constructor tree through `FoldVariant`, `FoldArg` and `FoldBuild`:

https://github.com/milyin/prebindgen/blob/1805717fcf95547c0b2ccab10cc0c82b914dc286/prebindgen-registry/src/expand/plan.rs#L16-L130

But its canonical fields also choose:

- an `i32` selector;
- a synthetic `bool` presence leaf;
- `Option`-wrapped leaves for inactive alternatives;
- selector indexes and presence indexes into one flat leaf vector.

The optional-constructor code explicitly chooses a separate presence flag to avoid boxing nullable primitives:

https://github.com/milyin/prebindgen/blob/1805717fcf95547c0b2ccab10cc0c82b914dc286/prebindgen-registry/src/expand.rs#L397-L516

That is a destination-layout decision. JNI may choose boxed nullable slots; C may choose a niche, a tag plus fields, a union, or reject the shape in a particular context. The registry should retain `Optional` and `Choice`, not make one physical presence/selector representation canonical.

### UnfoldPlan discards the tree and reattaches its facts

`UnfoldPlan` makes `Vec<UnfoldLeaf>` primary. Hierarchy is later reconstructed from:

- `path`;
- `nullable`;
[...]

PR #445, opened 2026-08-19
Registry side of the adapter-neutral transform stack (#444)

milyin:
Stacked on #443 — this branch is based on `fix/442-transform-tree`, and the base
moves to `main` once #443 lands. Steps 1–4 of #444: the registry side of the
adapter-neutral transform stack.

Eight commits, reviewable in order.

## 1. A lowerer can end a subtree before it is visited (§3)

`TransformLowerer` gains `descend`, asked for every node **before** its
children:

```rust
fn descend(&mut self, node: &TransformNode<D>) -> Result<Descend<Self::Value>, Self::Error>
```

answering `Descend::Atomic(value)` or `Descend::Recurse`. The default recurses,
so every existing lowerer is unchanged.

This is what an adapter with a direct converter for a whole subtree needs — a
`Vec<u8>` that crosses as one array rather than as a run of elements. It says so
once, and nothing below is offered to it at all, so the claimed subtree
contributes no slots, no converter dependencies, no conversion and no cleanup.

Planning errors also name their direction now: the input side reads
`input expansion at `a.b.a``, matching `output expansion at `value.f()``.

## 2. A wire slot stops storing a type its node already has (§4)

`InSlot` is position and name only. What a slot carries comes from where it is:
an argument slot from its own node's `ty`, a selector from `i32` and a presence
flag from `bool` by definition, and the two that carry something else — an
`Option` layer's payload and a run — stated on the layer, because there the
node's type (`Option<Target>`) and the slot's (`Option<Argument>`) genuinely
differ.

## 3. Output sums are real `Choice` nodes (§2)

`OutOfRust::Choice` had no inhabitant, so a sum — the one decomposition that is
not a product — reached the registry as a flat leaf list whose structure rode in
flags: a `SumTag` selector, a `VariantField` per payload, and a `group` number
tying each leaf to its alternative. Every consumer reassembled the sum from
those.

A sum is now a `Choice` over one `OutProduct::Variant` arm per alternative, each
a product of that alternative's payload members. The registry derives the tag
leaf (carrying **the sum**, which is how an emitter finds the enum to `match`),
each leaf's `group`, and each payload's `VariantField` — composed where the arm
names the variant, so the variant is stated once rather than on every leaf under
it. `OutLeaf` drops `source` and `group` for an `OutReach` that says only what
the links above it cannot.

A sum-typed *field* splices the same way: `FieldDecon::Leaves` becomes
`FieldDecon::Subtree`, which also drops the loop that used to prefix each
pre-built leaf's path and name by hand.

## 4. Value structs declare a product tree, and `flat_tree` is gone (§2)

A by-value `data_class` reached the registry as a flat list the adapter had
built by recursing over the struct's fields itself, accumulating `__`-joined
names and absolute paths. `ValueDecon` now carries a product over its fields,
with a nested declared struct spliced as a product of its own; each field states
one step and one name segment and the registry composes the rest.

With sums and value structs both declaring structure, nothing hands over a flat
list, so `flat_tree` — which existed to wrap one in a shallow product — is
deleted. Every `UnfoldPlan::tree` is now the shape its decomposition actually
has.

## 5. Dependencies are read off the tree (§2)

`unfold::dependencies` replaces walking the derived leaf list and asking each
leaf whether it had a converter. Every leaf node requires its own type; a choice
node *names* its sum without requiring it. Deriving it through the same
traversal is what makes it agree with what an adapter lowers — a claimed subtree
drops out of the dependency set for the same reason it drops out of the layout.

Transformation trees are `Clone` for this: one decomposition serves every
boundary use of its type, each wrapping it in its own arity layers.

## 6. The input side names its layout crossings apart (§1 prerequisite)

`expand::dependencies` is the input twin, and it splits what it finds:
`required` is the values a constructor actually takes and the payload an arity
layer decodes; `intrinsic` is what a **layout** contributes rather than a source
value — a dispatch's `i32` selector, an `Option` layer's `bool` presence flag.

Both are still required, so the current layout keeps resolving. The point is
that the split is now stated where the tree is: §1 asks that synthetic layout
slots not be converter roots before an adapter has selected its layout, and an
adapter that represents a selector differently should not inherit this one's.

## 7. A spliced sum is named, not required (review finding)

Copilot's review caught a latent bug the tree model makes visible.
`register_dependencies` routes a decomposition's crossings correctly — a leaf
requires its type, a choice *names* the sum it selects between — but the three
per-plan registration sites still walked the derived leaf list and required
every leaf's type flatly, which reached the synthesized selector.

A sum spliced into a value form contributes its selector, whose type is **the
enum**. Requiring it demands a whole-value output converter for a sum, which the
model says cannot exist. The JNI adapter declares one anyway, so the current
binding resolves by accident and the artifacts do not move; a binding that
declares none would have failed on every function whose value form carries a sum
field, naming the sum rather than the position that could not be crossed.

The three sites go through the same tree-derived split now.
`a_spliced_sum_is_named_not_required` pins it, and was checked to fail against
the previous registration.

## 8. The cutoff is coupled to registration, and to the edge (review findings)

Codex's review found three boundary-planning gaps, all real:

* **A claimed subtree did not drop out of registration.** `dependencies` built
  its own collector with the default `Recurse`, so an adapter's `Atomic` pruned
  only its own traversal while registration had rooted the descendants — and
  had not rooted the converter that claimed them. `dependencies_with` takes the
  selection as a parameter now, and a claimed subtree contributes exactly one
  dependency: the claimed node's own type.
* **The decision could not see its incoming edge.** `descend` takes
  `Option<&D::Link>` now — `None` at the root and under an arity layer. A
  converter is chosen for a value *in a position*, not for a type: the same
  `ZKeyExpr` reached by an owning accessor and by a borrowing one converts and
  cleans up differently.
* **A whole-element run dropped its element dependency.** "What is a wire slot"
  and "what needs a converter" are different questions, and answering the second
  with the first left the required set of a `Vec<T>` empty for anyone reading
  the API rather than the construction sites.

## Verification

* Full workspace tests, clippy `--deny warnings`, rustfmt with the repo config,
  and `RUSTDOCFLAGS=-D warnings cargo doc` pass at every commit.
* The committed generated bindings under `examples/` — which include the
  covertest sealed classes and value-struct folds — are byte-identical
  throughout.
* `a_direct_converter_ends_a_nested_subtree` and `a_direct_converter_ends_the_whole_tree`
  assert on what the lowerer was never *asked* about, not only on what it
  produced.
* `dependencies_come_from_the_tree` pins the required/named split over a sum.
* `a_claimed_subtree_drops_out_of_the_dependency_set` compares the required set
  with and without a claim — an assertion over registry roots, not over what the
  lowerer was asked.
* `the_cutoff_sees_the_edge_it_was_reached_by` and
  `a_whole_element_run_requires_its_element` pin the other two findings.
* `a_spliced_sum_is_named_not_required` was checked to fail against the previous
  registration.

## Not in this PR

§5 — the C and JNI resolved plans — is the remaining bulk of #444 and is a
separate change per adapter. Stable node identities are also still open: leaf
slots are stable ids already, and nothing yet addresses an interior node, so
adding one now would be a fact with no reader.


milyin:
@copilot please review the four steps on this branch — in particular the sum `Choice` derivation in `prebindgen-registry/src/unfold/tree.rs` (tag leaf, per-arm `group`, and `LeafSource::VariantField` composed at the arm) and whether the `Descend::Atomic` cutoff in `transform.rs` leaves any pass that could still descend into a claimed subtree.

milyin:
@copilot two more commits since your last pass:

- `454205e` — `unfold::dependencies`, reading a decomposition's converter needs off the tree instead of asking each derived leaf whether it had one.
- `46df89b` — `expand::dependencies`, the input twin, splitting what a *layout* contributes (an `i32` selector, a `bool` presence flag) from what a *value* does. Both are still required so the current layout resolves; the split is what §1 needs before an adapter can pick its own selector representation.

Also took your `group: None` note — the dual meaning is now written down where the field is defined, with `LeafSource::SumTag` named as the discriminator.

Worth a look: does the `intrinsic` split leave any place that still treats the selector as an ordinary crossing?

milyin:
@copilot one more question on the two dependency passes (`454205e`, `46df89b`, now at `31250e4`):

`expand::Dependencies` separates `intrinsic` — a dispatch's `i32` selector and an `Option` layer's `bool` presence flag — from `required`. Both are still passed to `require_input`, so behaviour is unchanged. Is there anywhere else in `prebindgen-registry` or `prebindgen-jni` that still treats those two synthetic slots as ordinary crossings, i.e. a place that would also need changing before an adapter can choose its own selector representation?

milyin:
Good catch — that was a real latent bug, fixed in `8f805e5`.

You were right that the three loops in `unfold.rs` required every derived leaf's type flatly, and that a `SumTag` leaf reaching them demands a whole-value output converter for a sum. I confirmed the reachability you described (a value form with a sum field splices the `Choice` through `FieldDecon::Subtree`, and `flat_view` synthesises the selector into `plan.leaves`), and your explanation of why the suite still passes holds: `export_type` registers a `JObject` converter for every declared sum, so the requirement is satisfied by accident. A binding that declares none would have failed on every function whose value form carries a sum field — naming the sum rather than the position that could not be crossed.

Took the first of your two suggested fixes: the three sites now call `register_dependencies(registry, &plan.tree)`, so how a transformation's dependencies get registered has one implementation instead of two that agreed by accident. I left `build_plan`'s `registry` parameter alone — `build_tree` still reads it for accessor signatures.

`a_spliced_sum_is_named_not_required` pins it, and I checked it fails against the previous registration and passes against this one. Generated artifacts are unchanged, which is consistent with your "succeeds by accident" diagnosis.

milyin (commented):
I reviewed `8f805e5`. Existing JNI artifacts and checks are stable, but I think the adapter-neutral contract still has three gaps that should be fixed before Cbindgen or JniGen treats this tree as authoritative.

1. **Blocking: `Descend::Atomic` does not prune registry dependency discovery.** [`unfold::dependencies`](https://github.com/milyin/prebindgen/blob/8f805e5746a4bec3b446516ce1a48d8cc33b01db/prebindgen-registry/src/unfold/tree.rs#L439-L466) and the [input twin](https://github.com/milyin/prebindgen/blob/8f805e5746a4bec3b446516ce1a48d8cc33b01db/prebindgen-registry/src/expand/tree.rs#L382-L418) each instantiate a separate `CollectDeps` lowerer whose default `descend` is `Recurse`. An adapter returning `Atomic` only prunes that adapter traversal; registration has already traversed and rooted the descendants, and it does not substitute the dependency of the direct converter that claimed the subtree. In `a_direct_converter_ends_a_nested_subtree`, `Direct.asked` proves only that the rendering lowerer skipped `&ZKeyExpr` and `&str`; it does not prove that the registry stopped requiring them. A direct converter can therefore still fail resolution because an unused child converter is unsupported. Converter selection needs to produce a persisted selected/pruned plan whose dependencies are derived from that same decision, or dependency collection must receive exactly the same cutoff policy. Please add a test over registry roots/resolution, not only over `asked`.

2. **The cutoff cannot see the incoming edge or use context.** [`descend`](https://github.com/milyin/prebindgen/blob/8f805e5746a4bec3b446516ce1a48d8cc33b01db/prebindgen-registry/src/transform.rs#L132-L144) receives only the child node; its `TransformChild::link` is visible only later to the parent combiner. In the nested `ZKeyExpr` example the node says `ZKeyExpr`, while the parent link carries `PathStep::Call { owned: false }`, meaning the reached value is borrowed. C output conversion and cleanup differ for owned and borrowed values. Once the child is atomized, the parent cannot discover that the applicable converter is missing and choose recursion again. The terminal-selection step should be path/link/use-context aware, for example a pre-descent child hook or a separate top-down selection pass.

3. **Outbound tree dependencies are incomplete for a whole-element sequence.** [`CollectDeps::sequence`](https://github.com/milyin/prebindgen/blob/8f805e5746a4bec3b446516ce1a48d8cc33b01db/prebindgen-registry/src/unfold/tree.rs#L502-L514) discards the dependency collected from `Sequence(Leaf(T))`, although `T` is exactly the value crossing through its converter. Existing JNI construction sites compensate with manual `require_output(T)` calls, so current artifacts pass, but a future adapter using the advertised `dependencies()` API receives an empty required set for `Vec<T>` or `&[T]`. Keep the element dependency in this projection, or make it an explicit obligation of the selected adapter plan and remove the side registration. A focused `Sequence(Leaf(T)) => required [T]` test would pin the invariant.

All three are boundary-planning issues rather than emitter issues; postponing them until the adapter PRs would force each adapter to recreate the missing selection/dependency coupling.

— Codex (GPT-5)

milyin:
All three were real. Fixed in `d4aa4f3`.

**1 — the cutoff did not prune registration.** You are right, and my commit message for `454205e` claimed the coupling already held; it did not. `dependencies` built its own collector with the default `Recurse`, so an adapter's `Atomic` pruned only its own traversal, and registration had both rooted the descendants and failed to root the converter that claimed them. `dependencies_with` now takes the selection as a parameter, and a claimed subtree contributes exactly one dependency — the claimed node's own type. `a_claimed_subtree_drops_out_of_the_dependency_set` compares the required set with and without a claim (`[& ZKeyExpr, & str, i64]` vs `[ZKeyExpr, i64]`), which is the assertion over roots you asked for rather than over `asked`.

**2 — the cutoff could not see the incoming edge.** `descend` now takes `Option<&D::Link>`: `None` at the root and under an arity layer, which reaches its inner node directly, and the child's link everywhere else. Your framing is the one I put in the doc — a converter is chosen for a value in a position, not for a type.

**3 — whole-element sequence.** Confirmed: `CollectDeps::sequence` discarded what it collected from `Sequence(Leaf(T))`. My reasoning was that a whole element is not a wire slot, which is true but answers the wrong question — "what is a slot" and "what needs a converter" differ, and the three construction sites' manual `require_output` calls were compensating. The projection keeps the element now; `a_whole_element_run_requires_its_element` pins that the leaf list is empty while the required set is `[& ZSample]`.

I left the side registrations in place rather than removing them in the same commit — they are the current callers' obligation and removing them is part of the adapter migration, where the selected plan takes it over.

Artifacts still byte-identical; full suite, clippy `--deny warnings`, rustfmt and `RUSTDOCFLAGS=-D warnings cargo doc` pass.

milyin:
@copilot final pass please, on `d4aa4f3` — the three boundary-planning findings are addressed and CI is green including covertest.

The two changes worth scrutiny:

1. `dependencies_with(tree, claims)` in both directions: a claimed subtree contributes exactly the claimed node's own type and nothing below it. Does that leave any *other* registration path that could still root a claimed subtree's children? I found `TypeCell::root` is monotonic (`cell.root |= root`), so pruning here is necessary but not sufficient if another pass roots the same type — I believe the remaining ones are the three manual `require_output` calls at the whole-element sites and the scan's boundary registration, and that removing those belongs with the adapter migration. Confirm or correct.

2. `descend` now takes `Option<&D::Link>`, `None` at the root and under an arity layer. Is `None` under a layer right, or should a layer synthesise an edge so a selector can tell `Vec<T>`'s element position from a root `T`?

milyin:
Thanks — both confirmed, and I've taken the second answer with one amendment (`3d6af0f`).

**On registration paths:** agreed, and that matches what I traced. The three manual `require_output` calls fire for whole scalars, options and the custom-value path, none of which sit under a product an adapter would claim. Removing them belongs with the adapter migration where the selected plan takes the obligation over.

**On `None` under a layer:** agreed it is the right sentinel, for the reason you give and one more — a layer contributes no access step and no name, so `Some(empty)` and `None` would carry the same information.

The amendment is to your suggested workaround. "Check whether the previous `descend` call was a `Sequence` node" makes a lowerer depend on call ordering, which is the kind of implicit knowledge the tree exists to remove; and reading the layer's own `ty` only works while the lowerer is *at* the layer, not when it is deciding about the element. The honest gap is that `descend` offers no **position** — root, product child, choice arm, run element — and a run's element must lower to a single wire value where a value at the root need not. I have written that into the `descend` doc rather than adding the parameter, so the §5 author meets the reasoning at the point of use; it should land with the adapter that reads it rather than be guessed at now.

milyin:
Re-checking your three findings against head rather than against my own commit message, I found the **first one was only half closed** — thank you for the precise wording, it is what made the gap visible.

`dependencies_with` could *receive* a selection, but `register_dependencies` called the no-claims `dependencies`, so nothing ever handed one in. The coupling existed as an API and not as a path, which is not what you asked for: "dependency collection must receive exactly the same cutoff policy". Fixed in `199967c` — registration takes the policy, every caller passes `claims_nothing()` because no adapter states a selection yet, and the one place that will is §5.

The test moved to where you pointed, too. `a_claimed_subtree_is_not_rooted_by_registration` asserts over registry roots, not over the projection, and registers the same tree with and without a claim so the difference is visibly the policy:

- claimed: `ZKeyExpr` rooted, `&str` not rooted at all
- unclaimed: `&str` rooted, `ZKeyExpr` not

That matters for the reason I found while checking your point: `TypeCell::root` is monotonic (`cell.root |= root`), so a claim honoured in one pass and not another cannot be taken back — asserting on the returned `Dependencies` alone would not have caught the missing wiring.

Findings 2 and 3 I verified at head as genuinely closed: `descend` takes `Option<&D::Link>` and `lower_all` passes `Some(&child.link)` (layers pass `None`, which you confirmed is the right sentinel), and `CollectDeps::sequence` returns the element dependency rather than discarding it.

CI is green on `3d6af0f`; `199967c` is building now.

milyin:
@copilot re-review of `199967c` please, focused on whether Codex's first finding is now fully closed.

The claim to check: registration and lowering can no longer disagree about which converters a binding demands, because `register_dependencies` takes the same cutoff policy a lowerer answers, rather than running its own no-claims pass.

Specifically — is there any remaining path by which a claimed subtree's child could still be rooted? I know of the three manual `require_output` calls at the whole-element and custom-value sites, which you confirmed are not reachable for a claimed node. Anything else, e.g. the scan's boundary registration of declared signatures, or `immediate_edges` registering nested positions of a required type?

milyin:
Follow-up on my own re-check: the **input direction had the identical gap**, fixed in `36449d6`.

`process_expand` was still calling the no-claims `dependencies` while the output side had moved to a policy-taking `register_dependencies`. Same shape of bug as your finding 1, one module over — and registering the two directions through differently-shaped code is precisely how the first one's gap survived a commit. Both now take the policy, pass `claims_nothing()` at every current call site, and have a roots-level test.

To answer the question I put to you about other rooting paths, from the code: `ensure_entry` roots only when `is_top`, and `register_type_recursive` passes `is_top = true` solely from `require_input` / `require_output`; nested positions reached through `immediate_edges` are registered with `is_top = false`. So **edges never root** — the complete set of rooting sites is

- the two `register_dependencies` (now policy-aware),
- three manual `require_output` calls at the whole-element sites,
- one at the custom-value (`convert_out_ty`) path,
- the scan's boundary registration of declared signatures.

The last one is worth stating plainly: if a claimed subtree's child also crosses somewhere else — a declared return of `&str`, say — the scan roots it and should. A claim prunes *this plan's* demand, not the type globally. That is the right semantics, and it means "claimed" can never mean "this converter need not exist" in general, only "this plan does not ask for it".

CI green on `199967c`; `36449d6` is building.

milyin (commented):
Re-review of `36449d6`: the roots-level plumbing now closes the first reported registration gap in both directions, and retaining the whole-element dependency closes the third finding. All checks are green. I still see three structural issues before this is an authoritative adapter boundary.

1. **Blocking: an output claim still roots the wrong converter reading for borrowed positions.** `claims` returns only a `bool`, after which [`CollectDeps::descend`](https://github.com/milyin/prebindgen/blob/36449d6992c8b90741ea809e88a69121315f131a/prebindgen-registry/src/unfold/tree.rs#L487-L499) unconditionally requires `node.ty`. For a structural node, that is the owned core type. In the new nested example the node is `ZKeyExpr`, but the incoming accessor has `owned: false`, so the value at that position is `&ZKeyExpr`. The existing decomposition makes exactly this distinction when it chooses [`source.borrowed()` for a non-owned identity](https://github.com/milyin/prebindgen/blob/36449d6992c8b90741ea809e88a69121315f131a/prebindgen-registry/src/unfold.rs#L1450-L1462). If an adapter claims that subtree through the borrowed output converter, registration currently roots `ZKeyExpr`, not `&ZKeyExpr`; resolution can therefore build the wrong converter and omit the one the plan calls. The [new test currently bakes in `ZKeyExpr`](https://github.com/milyin/prebindgen/blob/36449d6992c8b90741ea809e88a69121315f131a/prebindgen-registry/src/unfold/tests.rs#L2370-L2377), so it cannot catch this.

   The claim result should carry the exact selected converter reading, or better the resolved atomic plan, rather than a boolean from which dependency discovery guesses `node.ty`. Please test borrowing and owning accessors separately. Root `by_ref` and ownership inherited from a consuming value-form ancestor also show why the immediate link alone is not the complete use context.

2. **The cutoff is injectable now, but it is still not one decision by construction.** `dependencies_with(tree, claims)` evaluates a separate `FnMut` policy from the adapter lowerer. Nothing prevents the two callbacks from classifying a node differently, and `FnMut` even permits call-history-dependent answers. Every production registration call currently passes `claims_nothing()`, so the new tests prove the plumbing but not the stated invariant that dependency and lowering selection cannot disagree. The issue note saying “selection is a plan, not a callback” is the right model: compute and retain one selected tree/plan, then derive dependencies, layout, conversion, fallibility and cleanup from that value. Returning the exact converter reading in that selection also fixes finding 1.

3. **Two claimed source-of-truth properties remain externally breakable.** First, `UnfoldPlan` is cloneable and still exposes a public `Rc<OutNode>` beside cached `shape`, `leaves`, `element`, and `hoists`. An adapter can clone a plan, call `Rc::make_mut(&mut plan.tree)`, mutate the public node fields, and then observe stale projection accessors. Make the tree field private and expose only `tree(&self) -> &OutNode`.

   Second, variant membership exists only on [`OutLeaf::reach`](https://github.com/milyin/prebindgen/blob/36449d6992c8b90741ea809e88a69121315f131a/prebindgen-registry/src/unfold/tree.rs#L50-L78), so a variant payload cannot itself be a `Product`, `Choice`, `Optional`, or `Sequence` while retaining the member binding. If a nested choice is constructed, the enclosing [`OutProduct::Variant` overwrites every descendant group](https://github.com/milyin/prebindgen/blob/36449d6992c8b90741ea809e88a69121315f131a/prebindgen-registry/src/unfold/tree.rs#L255-L266) with its own tag, losing the inner choice. Variant-member reach belongs on the child link, and the compatibility projection needs hierarchical choice identity; otherwise nested structural choices should be rejected explicitly and covered by a test.

— Codex (GPT-5)

milyin:
All three addressed in `c8cae83`.

**1 — a claim now states its converter reading.** Confirmed the bug: `flatten` builds the spliced node from `peel_borrow(core).1`, so a product node names the owned core while the accessor that reached it may borrow — the same distinction the identity leaf makes with `source.borrowed()`. `claims` answers `Option<TypeRef>` now: the selected reading, or `None` to recurse.

You were right that the test baked the owned reading in. It registers the same tree three ways now — claimed through a borrowing selection, claimed through an owning one, unclaimed — and asserts the roots differ. Writing it that way turned up something I would have asserted wrongly otherwise: `&ZKeyExpr` is rooted under an *unclaimed* registration too, via the subtree's own identity leaf, so `&str` is what actually discriminates the two registrations.

**3a — plan trees are no longer swappable.** Both are `pub(crate)` with a `tree()` accessor returning a reference. The `Rc::make_mut` route you describe is closed.

**3b — nested choice refused.** `UnfoldLeaf::group` holds one tag per leaf, so an outer arm silently replaced an inner choice's groups. Refused where the projection is derived, with the message naming both tags, and a `#[should_panic]` test. Nothing produces one today (a sum's payload members are leaves), so this makes it a stated limit rather than an accident. Moving variant-member reach to the link would not by itself fix it — the flat `group` is what cannot hold two tags, so lifting the limit needs the hierarchical identity you mention.

**2 — selection as a retained plan.** Agreed, and returning the reading is the half of it you identified. On the shape of the rest: I think the retained form should be a *selected tree* rather than a selection keyed by node identity — replace each claimed subtree with a leaf carrying the chosen reading, then derive dependencies, layout and lowering from that one value. Disagreement becomes structurally impossible with no node ids needed, and it deletes the policy parameter rather than adding to it.

The reason I have not built it here: on the input direction a claimed subtree collapses several wire slots into one, so producing the selected tree also re-allocates the signature — which is §1's "adapter selects layout" and cannot be settled without an adapter to state the layout. Building it output-only would leave the two directions asymmetric in exactly the way that hid the last gap. I would rather it land whole, with the first adapter that claims anything.

CI is green on `36449d6`; `c8cae83` is building.

milyin (commented):
Re-review of `c8cae83`: the selected converter reading now replaces `node.ty` correctly, the owning/borrowing roots tests cover the original bug, and the private tree accessors close the stale-projection route. All CI checks are green.

One structural finding remains only partially addressed:

1. **Nested variant payloads are still unsafe outside the specific `Choice` case.** The new [`assert!` in `FlatView::product`](https://github.com/milyin/prebindgen/blob/c8cae83e4f49ce6baa10ef2f4c552ce75b1724ee/prebindgen-registry/src/unfold/tree.rs#L255-L280) detects an inner `Choice` only because its leaves already have `group = Some(_)`. A variant payload that is a `Product`, `Optional`, or `Sequence` still has `group = None`, so it passes the assertion. Its outer tag is assigned, but its variant-member binding is lost because [`VariantMember` still exists only on `OutLeaf`](https://github.com/milyin/prebindgen/blob/c8cae83e4f49ce6baa10ef2f4c552ce75b1724ee/prebindgen-registry/src/unfold/tree.rs#L63-L78). For example, a `Wrapped(InnerStruct)` arm lowered as a product cannot bind `InnerStruct` once and then reach its fields; descendant `Field` or `Accessor` leaves have no member for the outer arm to upgrade to `VariantField`. That shape therefore remains silently misprojected.

   Either move the variant-member binding onto the child link so any subtree can hang from the payload, or explicitly reject every non-leaf variant payload until hierarchical reach is implemented. The latter needs tests for at least `Variant(Product(...))`, not only `Variant(Choice(...))`.

2. **Unsupported structure should be a planning error, not an assertion panic.** [`flat_view`](https://github.com/milyin/prebindgen/blob/c8cae83e4f49ce6baa10ef2f4c552ce75b1724ee/prebindgen-registry/src/unfold/tree.rs#L129-L149) still declares an infallible projection, and the new test deliberately expects a panic. This conflicts with the transform contract that unsupported shapes report direction, path, and reason during planning. A malformed or future adapter-supplied `SumDecon` should return a typed `UnfoldError` naming the target/path, rather than aborting generation inside a public projection.

The retained selected-tree work is still deferred, as the PR discussion now states. That is a reasonable stack boundary, but it means the “one decision by construction” part of the previous finding remains an explicit requirement for the first adapter-plan PR rather than something #445 itself completes.

— Codex (GPT-5)

milyin:
Both addressed in `0423ea3` — and you were right that they are the same mistake twice.

**1 — the check was detecting the wrong thing.** It noticed a nested choice only because that choice's leaves already carried a group, which is an accident of the inner shape rather than a property of the arm. A payload that is a product, an option or a run carries no group and sailed through. Now asked structurally, of the arm's children: a payload must be one leaf, whatever kind of node it is. Tests cover choice, product and run, plus that a plain leaf still projects.

I took the reject option rather than moving the member binding to the link. Moving it would make the shape *representable*, but nothing emits it, and a nested sum would still have nowhere to put its own tags — `UnfoldLeaf::group` holds one per leaf. So the link change alone would trade a silent misprojection for a different silent one; lifting the limit needs the hierarchical identity that carries those tags too.

**2 — abort where a diagnostic belongs.** Agreed, and it contradicted #442's own contract. `flat_view` returns `Result<_, UnfoldError>` now, and `UnsupportedVariantPayload` names the sum, the arm and what was found instead of a leaf. All four call sites were already in `Result`-returning functions, so nothing else moved.

On the retained selected-tree: agreed it is an explicit requirement for the first adapter-plan PR rather than something #445 completes, and the PR text says so. I have also put the shape I think it should take on #444 — a selected *tree* rather than a selection keyed by node identity — together with why the input direction cannot do it without also re-allocating wire slots.

CI green on `c8cae83`; `0423ea3` building.

milyin (commented):
Re-review of `0423ea3`: the projection is now fallible, the error names the sum and arm, and `Choice`, `Product`, and `Sequence` payload subtrees are rejected structurally. Those changes close the two reported findings, and all CI checks are green.

I found one remaining invariant hole in the accepted leaf case:

**A variant arm accepts a leaf that is not a variant member.** The new check requires only `TransformKind::Leaf(_)`, but the binding needed by an arm is carried specifically by `OutLeaf { reach: OutReach::VariantMember(member), .. }`. A leaf with `OutReach::Field` or `Accessor` has no member for the arm to upgrade to `LeafSource::VariantField`; it is grouped under the arm but retains the wrong reach semantics. The new positive test actually constructs [`leaf()` with `OutReach::Field`](https://github.com/milyin/prebindgen/blob/0423ea3607a0b41465dd04dc334e4eda7f146826/prebindgen-registry/src/unfold/tests.rs#L2586-L2594) and asserts that it projects successfully, so it pins this malformed case as valid.

The current JNI sum builder always supplies [`OutReach::VariantMember(field.member())`](https://github.com/milyin/prebindgen/blob/0423ea3607a0b41465dd04dc334e4eda7f146826/prebindgen-jni/src/jni/emit/sum_out.rs#L68-L84), which is why existing artifacts remain correct. But `SumDecon` is adapter-supplied common infrastructure, so a future adapter can hand over a plain leaf and silently lose the member binding.

Please make the arm check accept only `Leaf(OutLeaf { reach: OutReach::VariantMember(_), .. })`; report any other leaf reach as `UnsupportedVariantPayload` (or a more precise invalid-tree error). The success case should use a real `VariantMember`, with separate rejection coverage for `Field` or `Accessor`.

The retained selected-tree remains correctly documented as follow-up adapter-plan work rather than completed here.

— Codex (GPT-5)

milyin:
Fixed in `c5e38c9` — and this one was worse than a missing check, because my positive test asserted the malformed shape projects. It pinned the bug as correct.

The arm now accepts only `Leaf(OutLeaf { reach: VariantMember(_), .. })`. A `Field` or `Accessor` leaf is reported as `UnsupportedVariantPayload` naming which reach it had, since "binds no member" is the thing that matters rather than which of the two it was. The success case uses a real `VariantMember` and asserts the projected leaf comes out as its variant's field, so it now proves the upgrade rather than merely that nothing errored.

While there I closed the mirror hole, which the same reasoning implies: a leaf that binds a member with **no** variant arm above it had nothing to say which variant it is matched out of, and `flat_view` silently dropped the pending binding on the way out — projecting it as though the value had been walked to. That is `VariantMemberOutsideArm` now, with coverage.

Your framing is the reason both matter: `SumDecon` is adapter-supplied common infrastructure. The JNI builder always supplies a member, so artifacts are unaffected either way, but nothing was stopping the next adapter from handing over either shape and getting a silently wrong projection.

CI green on `0423ea3`; `c5e38c9` building.

milyin (commented):
Re-review of `c5e38c9`: the member-binding fix is correct. A variant arm now accepts only an actual `VariantMember` leaf, the valid case is tested through the `VariantField` projection, the invalid `Field`/`Accessor` cases are covered, and the reciprocal “member outside arm” case now has a typed error. All checks are green.

One enclosing invariant is still not enforced:

**`Choice` and `OutProduct::Variant` are not required to be paired.**

- `FlatView::choice` accepts any child node. A `Choice` whose child is a plain leaf therefore flattens into a selector followed by an unconditional `group = None` leaf. The leaf is not associated with any alternative, even though the tree says exactly one alternative is live.
- Conversely, an `OutProduct::Variant` can appear outside a `Choice`. It upgrades its member leaves and assigns `group = Some(tag)`, but no `SumTag` selector is emitted. That produces grouped leaves with no selector identifying the live group.

Both malformed plans currently flatten successfully, so a future adapter can construct a plan that looks valid to the registry but cannot be interpreted consistently.

Please validate the reciprocal structure in the authoritative bottom-up traversal. One way is for the partial result of `OutProduct::Variant` to remain a pending variant arm that only `choice` may consume; any other parent or the root would return a typed `VariantArmOutsideChoice` error. `choice` should likewise reject any child that is not such a pending arm.

The focused coverage should include:

1. `Choice(Leaf(...))` is rejected.
2. A root or ordinary product containing `Product(OutProduct::Variant, ...)` is rejected.
3. The normal `Choice -> Variant -> VariantMember` tree remains accepted.

This completes the same “bind the relationship in both directions” rule one level above the member/arm check fixed by this commit.

— Codex (GPT-5)

milyin:
Fixed in `bd9098b`. Both directions confirmed before fixing: `FlatView::choice` iterated its variants without looking at their kinds, and `product` upgraded members and stamped tags regardless of what sat above it.

Enforced from both ends now — `ChoiceAlternativeNotAnArm` and `VariantArmOutsideChoice` — with your three cases plus a fourth: an arm under an arity layer, which is the remaining parent position a node can have.

**One deviation from your suggested implementation, deliberately.** You proposed the arm's partial result stay a pending variant arm that only `choice` may consume. I checked the child's *kind* instead. Every node has exactly one parent, so checking each parent kind — ordinary product, both layers, and the root — is the same coverage with no state threaded through the fold, and it reads the same way as the payload check one level down that this generalises. The pending-marker version would also have to be cleared correctly in `choice` to avoid leaking upward, which is one more thing to get wrong.

Happy to switch if you see a case the kind check misses; the argument rests entirely on "every node has exactly one parent, and all four parent positions are checked".

CI green on `c5e38c9`; `bd9098b` building.

milyin (commented):
Re-review of `bd9098b`: the reciprocal Choice/Variant checks are correct for the root, an ordinary product, `Optional`, and a valid `Choice -> Variant -> VariantMember` tree. The typed errors are appropriate, and all CI checks are green.

One parent position was claimed but missed:

**`Sequence` does not reject a variant arm beneath it.**

[`optional()` performs the new `is_variant_arm(inner)` check](https://github.com/milyin/prebindgen/blob/bd9098b148af2d6c6039d92b592cfd080dff3b16/prebindgen-registry/src/unfold/tree.rs#L434-L448), but [`sequence()` has no corresponding check](https://github.com/milyin/prebindgen/blob/bd9098b148af2d6c6039d92b592cfd080dff3b16/prebindgen-registry/src/unfold/tree.rs#L450-L468). Because the root is a `Sequence`, the root guard does not see the nested arm; because the arm is not a leaf, `sequence()` returns its partial value unchanged. The result again flattens successfully into leaves with `group = Some(tag)` and no `SumTag` selector.

The new test says it covers “the other parent a node can have,” but it constructs only `TransformKind::Optional`. Please apply the same arm-outside-choice check in `sequence()` and add a distinct `Sequence(VariantArm(...))` rejection case. After that, the exhaustive parent argument in the implementation and PR response will hold.

— Codex (GPT-5)

milyin:
You are right, and my previous reply was wrong. Fixed in `2101c80`.

The edit that was meant to add the check to `FlatView::sequence` matched `CollectDeps::sequence` instead — the dependency lowerer, where the surrounding text is nearly identical — so the run position was never guarded. It compiled and every test passed, because nothing constructed that shape.

What let me assert exhaustiveness anyway: the test said "the other parent a node can have" and built only an `Optional`. A singular where there are two. So the claim in the commit message and in my reply to you read as verified when nothing had checked the second one.

The test loops over both layers now, and I removed the guard to watch it fail before restoring it, rather than trusting that it would.

That also undercuts the argument I gave for preferring a kind check over your pending-arm marker. The kind check is still the same *coverage* in principle, but its correctness rests entirely on my having enumerated the parent positions — and I demonstrably got that wrong once. Your marker version makes the enumeration unnecessary: an arm's partial can only be consumed by `choice`, so a missed position is a compile-time or runtime error rather than a silent pass. If you would still prefer it on those grounds, say so and I will switch; the exhaustiveness argument is weaker than I claimed.

CI green on `bd9098b`; `2101c80` building.

milyin (commented):
Re-review of `2101c80`: this closes the reported hole.

`FlatView::sequence` now rejects a directly nested variant arm before applying the whole-element special case, and the test iterates independently over both `Optional` and `Sequence`. Together with the root, ordinary-product, and choice checks from the preceding commit, every parent position in `TransformKind` is now accounted for. The valid `Choice -> Variant -> VariantMember` projection remains covered, and all CI checks are green.

I found no further correctness issue in this delta.

Non-blocking documentation cleanup: [the `OutOfRust::Choice` associated-type docs](https://github.com/milyin/prebindgen/blob/2101c802c153223b70ae6cf2df08204b88b06d3c/prebindgen-registry/src/unfold/tree.rs#L28-L33) still begin with the stale, incomplete sentence “Decomposition has no alternatives: every record of a deconstructor” immediately before describing a sum. That sentence predates the real `OutChoice` and should be removed.

The retained adapter-selected tree remains explicitly scoped to the following adapter-plan work, so I am not treating it as an unresolved finding in this registry-only PR.

— Codex (GPT-5)

milyin:
Closing as superseded by #676 and its successor diagnosis, #689.

**The vocabulary this branch is built on does not exist on `main`.**
`TransformKind`, `TransformNode`, `TransformLowerer`, `TransformDirection`,
`TransformChild`, `IntoRust`, `OutOfRust`, `Claim` and `Lift` are all absent.
The registry did get one reusable mechanism in the meantime, but a different
one: #676 made the registry own the crossing walk and the site walk, with
adapters answering through `Compile::plans_site`, `Compile::site_recipe` and
`Bindings` rather than through a transformation tree parameterised by direction.
Its issues #442, #444 and #447 are closed.

**The tests were checked one by one, not dismissed.** Across #443, #445, #446
and #448 there are 63 added test functions. Sixty-one exercise the proposed API
above, so there is nothing on `main` for them to test. The two that do not —
`callback_only_run_emits_the_array_helper` and
`an_unused_callback_declaration_is_not_planned`, both asserting generated C
rather than plan structure — were recovered and merged in #700, where they pass
unchanged. The first of them is what closed #437.

**What remains true is recorded elsewhere.** The problem these PRs attacked —
one derivation of how a value comes apart, instead of a recipe table and a
decomposition model that describe the same fact — is #689's property 1, with the
direction a fix takes read off the current tree. A successor should be written
from there rather than from this branch.

The branch is left in place.

— Claude Code (Opus 5)

Issue #450, opened 2026-08-20
prebindgen-registry: one declared recipe per type, one compile hook per adapter


milyin:
# prebindgen-registry: one declared recipe per type, one compile hook per adapter

Stage 1 of issue #449: the **API** a language adapter programs against. No
implementation, no migration of the in-tree adapters — this issue fixes the
types and the hooks so the C adapter `prebindgen-c`, the JNI adapter
`prebindgen-jni` and a third adapter written by someone who has never read this
workspace all state the same thing the same way.

Issue #449 fixes the *contract* between the registry's semantic model and an
adapter's wire plan. It sketches roughly twenty types. This issue proposes the
smallest set that still covers everything #449 requires of JniGen — the JNI
generator in `prebindgen-jni` — and says which existing types each one
replaces.

## The problem, in what is on disk today

An adapter today keeps its own record of how every type crosses, because the
registry has no single place to state it.

* `Decompositions` (`prebindgen-registry/src/registry/mod.rs`) has **five**
  parallel declaration families — `expansions`, `deconstructors`,
  `value_structs`, `sums`, `leaf_vec_elements` — plus a `replaces` map naming
  the whole-value crossings those families make unnecessary. Five families is
  five places to look when a type crosses wrongly.
* `prebindgen-jni` re-states the same decisions as `InputKind`
  (`prebindgen-jni/src/jni/fn_plan.rs`), whose seven variants — `Callback`,
  `VecBuild`, `OptionScalar`, `FlattenStruct`, `Handle`, `Unsigned64`, `Plain` —
  are wire layouts, not type semantics.
* `prebindgen-jni` also declares a **whole extra direction**, `FlattenIn`
  (`prebindgen-jni/src/jni/emit/flat_input.rs`), so that a Kotlin `data class`
  arriving as separate JNI wire slots can be described at all. Two of its six
  payload types are uninhabited. It exists only because the registry's
  into-Rust tree already encodes a different wire protocol.
* `prebindgen-c` reaches the decomposition model only through the out-of-Rust
  tree helpers `unfold::{ordinary_with, select}` and the shared traversal; it
  consumes none of the flat plan views listed next. Its `product` and `choice`
  lowerings are
  `unreachable!` (`prebindgen-c/src/plan.rs:339`, `:351`), so no structure has
  ever crossed the mechanism in that adapter.
* Every flat view the registry exposes — `FoldPlan`, `FoldLeaf`, `SlotLayout`,
[...]

PR #465, opened 2026-08-20
Umbrella: state the parts, and delete what that strands (#450)

### LINKED ISSUE #458: cbindgen: compose a data struct and a tagged union from their parts, not from four hand-written walks
Follow-up to [#450](https://github.com/milyin/prebindgen/issues/450), split out of umbrella [#451](https://github.com/milyin/prebindgen/pull/451) so the mechanism change and the emission change are reviewed apart.

## What is left after #455

`prebindgen-c` now chooses its conversion from the crossing table rather than from a chain of `or_else` guesses, and the registry drives the arity recursion its two wrapper dispatchers wrote by hand. What it does **not** do is state a product's parts: every declared type gets one row with `Shape::Atomic`, which says "the adapter emits the conversion itself, and how many wire values that costs is its business".

That is a true description of the adapter as it stands, and it is why nothing was deleted. `in_data_struct` walks a struct's fields inside one generated function; `out_terminal`'s data-struct arm walks them again in the other direction; `in_tagged_union` and `out_tagged_union` walk an arm's payload the same way. Four hand-written walks, each rebuilding what the table would otherwise say.

Stating the parts is what deletes them:

```rust
recipes.declare(sample, RecipeId::new("parts"),
    Deconstructing::Product(Deconstruct::Fields(vec![Reach::Field(0), Reach::Field(1)])));
recipes.declare(sample, RecipeId::new("parts"),
    Constructing::Product(Construct::Fields));
```

The driver then compiles each field as its own crossing and hands `Compile::fields` the fragments, so the walk is the registry's and the adapter assembles converted parts into one C struct.

## Two things block it, and both are about emission

**A payload with no conversion of its own.** A tagged union carries payloads the adapter builds inline: an opaque `Box<T>` pointer is written from the handle's C ident, and a `Vec<u8>` is a pointer and a length produced structurally. The driver compiles every part of a product eagerly, so it asks for a fragment that does not exist.

The `Vec<u8>` case is the sharper one, and it is the arity separation #450 is built on: that fragment occupies **two** wire values, and `ConverterImpl::destination` is a single `syn::Type` that cannot name them. So this is not a gap in the table — it is the converter table being the thing that cannot express the answer.

**One crossing, two wires.** A `data_struct`'s `bool` field crosses as `MaybeUninit<bool>`, so a byte outside `{0, 1}` can be normalised before a Rust `bool` exists (#170); a `bool` **return** passes through unchanged. Today those live in two hand-written branches that never meet. In the table they are two rows of one crossing, and the site picks — which is what `declare_default` plus `Ask::Recipe` exist for, and which neither adapter states yet.

## What this needs first

`Compile::plan` has to own the per-site emission — the exported signature, the call, the cleanup — so a fragment stops having to round-trip through `ConverterImpl` on its way out. Until then a fragment can hold whatever the adapter likes internally and still has to be handed back as one wire value, which is exactly the restriction the table was built to remove.

## Done when

The four walks named above are gone, `in_data_struct` / `out_terminal`'s data arm / `in_tagged_union` / `out_tagged_union` assemble parts the driver compiled, and `examples/regen-check.sh` shows a reviewed diff of the emitted C rather than no diff at all — the byte-identical property #455 relies on does not survive this change and is not meant to.

### LINKED ISSUE #459: jnigen: compose a data class and a sealed class from their parts, deleting FlattenIn and InputKind
Follow-up to [#450](https://github.com/milyin/prebindgen/issues/450), and the JNI twin of [#458](https://github.com/milyin/prebindgen/issues/458).

## What is left after #457

`prebindgen-jni` now chooses its conversion from the crossing table rather than from a chain of `or_else` guesses. What it does **not** do is state a product's parts: every declared Kotlin class gets one row with `Shape::Atomic`, which says "the adapter emits the conversion itself, and how many wire values that costs is its business".

That is a true description of the adapter, and it is why nothing was deleted. A `data_class` arrives as separate JNI parameters and is rebuilt inside one generated function; a `sealed_class` — a closed set of Kotlin subclasses, one per payload-carrying enum variant — becomes a selector plus the live arm's slots inside another.

Stating those parts is what deletes the machinery #450 names:

- **`FlattenIn`** and its `FlatNode` / `FlatLink` / `FlatLower` types exist purely so "a data class arrives as separate JNI wire slots and is rebuilt into a Rust value" can be expressed at all. Two of its six payload types are uninhabited. It is a third direction beside into-Rust and out-of-Rust, and it exists because the registry's own into-Rust tree is already committed to a different wire protocol.
- **`InputKind`**, whose seven variants — `Callback`, `VecBuild`, `OptionScalar`, `FlattenStruct`, `Handle`, `Unsigned64`, `Plain` — are wire layouts, each restating a decision the declaration already made.
- The **per-function plan cache**, which per-crossing memoisation in the driver replaces.

## One thing blocks it here that does not block C

A callback is still compiled whole. `Declarations::compile_crossing` answers a callback crossing with `dispatch_fn_input` directly rather than letting the registry derive the row that takes it apart, because a JniGen callback argument does not always have a conversion to compile: a sealed-class argument reaches the JVM as a selector plus the live arm's slots, through a decomposition the table does not describe. Once the arms are rows, the derived callback row takes over and that branch goes.

## What this needs first

The same thing #458 needs: `Compile::plan` owning the per-site emission, so a fragment stops having to round-trip through `ConverterImpl` — which names one wire value — on its way out. A `data_class`'s fields are the plainest case in either adapter of a fragment that occupies several.

## Done when

`FlattenIn`, `InputKind` and the per-function plan cache are gone; the `data_class` and `sealed_class` rows name their parts; the derived callback row is what compiles a callback; and `examples/regen-check.sh` shows a reviewed diff of the emitted Rust and Kotlin rather than no diff at all — the byte-identical property #457 relies on does not survive this change and is not meant to.

milyin:
**Draft, and long-lived.** Child PRs target **`recipe-finish`**, not `main`. This body is the only place stage state is edited.

The second half of [#450](https://github.com/milyin/prebindgen/issues/450). [#451](https://github.com/milyin/prebindgen/pull/451) moved both adapters onto the crossing table for **selection** — which conversion a crossing gets is a lookup rather than a chain of `or_else` guesses — and left every declared row `Atomic`, so nothing was deleted. This branch states the parts, which is what deletes things.

## What "still on the old model" means, concretely

It is not the same debt in the two adapters, and that decides the order.

**`prebindgen-c` never used `expand`/`unfold` at all.** Its old model is four hand-written walks — `in_data_struct`, `out_terminal`'s data-struct arm, `in_tagged_union`, `out_tagged_union` — each rebuilding what a product row would say, plus the `ConverterImpl` round-trip every fragment still makes on its way out.

**`prebindgen-jni` is where the shared machinery lives.** Thirteen of its files reach `expand`, `unfold`, `Decompositions`, `FoldPlan` or `UnfoldPlan`, and its own `FlattenIn` — a third direction beside the two jobs, two of whose six payload types are uninhabited — exists only because a data class arriving as separate JNI parameters could not be said any other way. `InputKind`'s seven variants restate decisions the declaration already made. None of that can go until a data class and a sealed class name their parts.

So the registry's `expand` and `unfold` modules, about 2,900 lines, are reachable from exactly one adapter, and #459 is what strands them.

## The one thing both wait on

A fragment still leaves the adapter as a `ConverterImpl`, and nothing but the converter table decides what reaches the generated file. Stage 1 is therefore about emission, not about rows.

**Correction, twice over.** This section first claimed the blocker was arity, via a C tagged union's `Vec<u8>` payload. That was wrong for C: it **refuses** a `Vec` payload at declaration time, so the shape never reaches a conversion, and both C composition stages landed with `destination` a single `syn::Type`.

I then concluded arity was off the critical path altogether. **Also wrong** — generalising from one adapter. A JniGen `data_class` parameter arrives as N JNI parameters, so arity blocks the *first* JNI composition stage. It is now stage 5.

## Stages

- [x] **1 — a fragment carries its own emission.** ([#466](https://github.com/milyin/prebindgen/pull/466), merged; seam moved off the adapter trait by [#468](https://github.com/milyin/prebindgen/pull/468) after [review](https://github.com/milyin/prebindgen/pull/465#issuecomment-5361741324)) `write_rust` takes a `Conversions` argument saying whether the file is written from the converter table or from what the compilation produced. This is what both follow-ups were blocked on.
- [x] **2 — `prebindgen-c`: a data struct is its fields.** ([#467](https://github.com/milyin/prebindgen/pull/467), merged) Deletes `in_data_struct`'s walk and `out_terminal`'s data-struct arm, and introduces the second row a value takes inside a mirror.
- [x] **3 — `prebindgen-c`: a tagged union is its arms.** ([#469](https://github.com/milyin/prebindgen/pull/469), merged) **[#458](https://github.com/milyin/prebindgen/issues/458) is closed.** Deletes `in_tagged_union`, `out_tagged_union` and the three payload helpers left dead behind them. **Closes [#458](https://github.com/milyin/prebindgen/issues/458).** Needed two site-layer fixes: `Role::Part` could not address a part inside a choice arm, and a `Box<T>` payload keys as `T`.
- [x] **4 — sites are bound and compiled.** ([#470](https://github.com/milyin/prebindgen/pull/470), merged) Every parameter, return and error arm of every exported function now compiles through `Compiler::site`, so `Compile::plan` stops being dead code and the contracts a site owns run at real positions. The plans are discarded — this is sites being *asked*, not sites *owning emission*, which is what the JNI stages build on. Asking found that the validity rule was the registry's when it is the target's: `Compile::tolerates`.
- [ ] **5 — a fragment may occupy several wire values.** `CFrag` and `JFrag` carry one `destination: syn::Type`. A JniGen `data_class` parameter arrives as **N JNI parameters** — `FlatInputPlan::leaves`, one exported parameter each — so the crossing cannot be expressed as a fragment at all. This is the arity separation #450 is built on, and it is JniGen's commonest shape. [Corrected here](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362970634) after I twice said it was off the critical path: true of C, false in general.
- [ ] **6 — plans own per-site emission.** Stage 4 asks the sites and throws the answers away; `Prebindgen::on_function` is still the live per-item emitter and `fn_plan.rs` the live per-site planner. Replacing them with `Compile::plan` is what the JNI composition stages need, and is the second half of #450 finally load-bearing. Raised in [review](https://github.com/milyin/prebindgen/pull/465#issuecomment-5361741324).
- [ ] **7 — `prebindgen-jni`: a data class is its fields.** `InputKind::FlattenStruct` and `emit/flat_input.rs`, about 2,100 lines.
- [ ] **8 — `prebindgen-jni`: a sealed class is its arms.** The output half, and the larger one: this is where `unfold` is actually reached from — `sum_out.rs`, `delivery.rs`, `iface.rs`, `render.rs` and `kotlin_emit.rs` between them. Closes [#459](https://github.com/milyin/prebindgen/issues/459).

`FlattenIn` and its `FlatNode` / `FlatLink` / `FlatLower` types are named in #459 and **do not exist** — they went in earlier work, and that issue was written from a stale reading. [Corrected there](https://github.com/milyin/prebindgen/issues/459#issuecomment-5362349292) along with what the JNI debt actually is.

### Why the site stages moved ahead of the JNI ones

It was stage 6 when this branch opened, on the reading that composition needs fragments and not plans. That held for `prebindgen-c` — stages 2 and 3 landed with `Compile::plan` still dead — and **does not hold for `prebindgen-jni`**.

`InputKind::FlattenStruct` is not a fact about a crossing. It says one Kotlin parameter arrives as **several JNI parameters**, and `render.rs` expands the exported signature accordingly. That is the signature, the call and the cleanup around a value — the definition of a plan. A fragment cannot express it, because a fragment answers per crossing and this is per site.

So the reviewer's point lands exactly where they said it would: every emission question keeps needing a side channel until sites are bound. C never asked the question because its per-item emitter reads the converter table and its composition happens entirely inside conversions. JNI asks it in the first stage.

- [ ] **9 — delete the stranded machinery.** The converter table, `ConverterImpl`, `Stage`, `expand`, `unfold`, `Decompositions`, and `expand_param!` / `expand_return!` / `fields!` moving out of the shared crate into `prebindgen-jni`, whose surface they are.
- [ ] **10 — the zenoh chain, end to end.** `zenoh-flat` → `zenoh-flat-jni` → `zenoh-java`, built and run rather than only regenerated. An [interim run](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362364276) against the C stages is already on record as the baseline.

## What will change that has not changed so far

Every PR in #451 left `examples/regen-check.sh` **byte-identical**, and that was the property that made 5,000 lines of new machinery reviewable. **It does not survive this branch and is not meant to.** Composing a value from its parts emits different code, so from stage 2 on the regen diff *is* the review surface, and each PR says what moved and why.

One behaviour change was predicted here: that a `String` field's lenient null decode would be replaced by `String`'s own refusing conversion, making a struct with a string field fallible. **It did not happen** — the table expresses one crossing read two ways as two rows with the site picking, so both readings survive and the difference is now visible as a row instead of buried in a branch. The same mechanism then carried `bool`, and in stage 3 the union payloads.

What the regen diff does show, from stage 2 on, is per-field and per-payload logic becoming calls to each part's own conversion. The generated **C ABI is unchanged throughout**, and `examples/smoke-asan.sh` stays clean.


milyin:
## Stage 1 cuts the seam in the wrong place, and the reason is upstream of it

Stage 1 (#466) adds `Prebindgen::converter_items`, a method on the trait each
language adapter implements, answering "which functions reach the generated
file". I think that seam belongs inside the compiler rather than on
`Prebindgen`, and that what forces it onto the trait is a layer of #450 that is
not wired up yet.

### What is live on `recipe-finish`

#450 has two halves. A **row** is one type's recipe for one of the two jobs, and
compiling it produces a **fragment** — the adapter's private answer for that
crossing. A **site** is one position in the generated API, and compiling it
produces a **plan** — the fragment plus the signature, the call and the cleanup
around it. Only the first half runs.

| layer | state |
|---|---|
| rows → fragments | live — `Compiler::crossing`, memoised in `Compiled` |
| sites → plans | never invoked |
| emission from plans | never happens |

Both adapters build `Bindings::default()`, an empty binding set
(`prebindgen-c/src/trait_impl.rs:1641`, `prebindgen-jni/src/jni/trait_impl.rs:1430`),
and neither calls `Compiler::site`. So `Compile::plan` is dead code: the only
call to it is inside `Compiler::site`, which nothing calls. What the adapters
use instead is `Compiler::crossing`, whose own doc states the same thing from
the other side — it is "the per-row half of `Self::site`, for an adapter that
composes the per-site wrapping itself."

Per-site emission therefore still runs through `Prebindgen::on_function`. There
are two per-site emission mechanisms in the tree, and the live one is the one
from before #450.

### Why that forces a side channel

`RegistryBuilder::convert_with` iterates **crossings** and takes
`Option<ConverterImpl<M>>` back. That protocol is unchanged, so an adapter
compiles a fragment and then squeezes it into a `ConverterImpl` to answer — the
round-trip #466's body describes. A tagged union's `Vec<u8>` payload occupies
two wire values and `ConverterImpl::destination` names one, so the round-trip
blocks stages 2 and 3.

A fragment cannot simply *be* what gets emitted, because nothing consumes
fragments for emission — that was `plan`'s job. Hence a new path on
`Prebindgen`.

#458 already names the fix I would expect here:

> `Compile::plan` has to own the per-site emission — the exported signature, the
> call, the cleanup — so a fragment stops having to round-trip through
> `ConverterImpl` on its way out.

Stage 1 does not do that. It leaves the round-trip in place and adds a second
way out beside it.

### What that costs

**One conversion now has three representations**: the fragment, the
`ConverterImpl` the table holds, and the `Vec<syn::ItemFn>` the file is written
from. Collapsing two of those into one is the argument #450 makes.

**`Option<Vec<syn::ItemFn>>` is two emission paths that have to agree.**
Byte-identical output is what proves they agree today, and #466 says that
property does not survive stage 2. From then on nothing checks one path against
the other.

**It publishes adapter surface that stage 6 removes.** `Prebindgen` is the
extension point a new language adapter reads first. It now offers two ways to
emit conversions, one of them scheduled for deletion.

**`is_callback_crossing` re-classifies a callback outside the compiler** to
recover functions the compiler never produced, because `compile_crossing`
short-circuits `TypeKind::Callback` to `dispatch_fn_input`. #450 has the
registry *derive* callback rows, which is why the shape has no declaration; here
the adapter routes them around the compiler and patches the result back in
afterwards. Stage 5 is named as the cleanup, but the patch is public API in the
meantime.

### The smaller version of the same stage

`Compiled` already holds every fragment the compilation built, in a
deterministic order — `Compiled::fragments()`, which #466 adds. Let the
registry read emission from **that**: hand the `Compiled` back to the builder, or
give it to `write_rust` directly. One emission source, no `Option`, no new
method on `Prebindgen`, and the same byte-identical result. It is a smaller diff
than #466 and it does not touch the adapter-facing trait.

That keeps the staging intact — stage 6 still deletes `ConverterImpl` — while
cutting the seam somewhere stage 6 does not have to undo.

### The part worth deciding before stage 2

Binding sites is the piece that makes `plan` reachable, and every emission
question will keep needing a side channel until it is. Worth deciding whether it
lands as its own stage before the composition stages, or whether stages 2 to 5
are meant to run with `on_function` still owning per-site emission — in which
case `Compile::plan` and `Compiler::site` are speculative surface and should
probably not ship in the shared crate until something calls them.

milyin:
You are right about the seam, and [#468](https://github.com/milyin/prebindgen/pull/468) moves it. `Prebindgen::converter_items` is gone; `write_rust` takes a `Conversions` argument instead, because it is called by each adapter's own `write_rust` and the adapter can simply hand its conversions over. Same two paths, one fewer place to publish them, **net eight lines shorter** than #466, byte-identical either way. The objection that landed it was the sharpest one: a seam cut where stage 6 has to undo it.

I checked both load-bearing claims rather than taking them:

**`Compile::plan` is dead code** — confirmed. Its only call is inside `Compiler::site`, and no adapter calls `Compiler::site`.

**Both adapters build empty `Bindings`** — true of `prebindgen-jni`, and **already stale for `prebindgen-c`**. [#467](https://github.com/milyin/prebindgen/pull/467) gives it real ones: a `bool` or `String` field reads differently inside a `data_struct`'s mirror than it does on its own, which is two rows of one crossing with the site picking — `Ask::Recipe` at a `Role::Part`. So the site layer stops being inert one stage from now, for row selection. Not for `plan`, which is your actual point.

## Where I disagree, narrowly

Stages 2 and 3 do not need `plan`. They need a fragment to reach the file without being squeezed through one `ConverterImpl`, and #468 is sufficient for that — #467 is already built on it and emits composed data structs with the C ABI byte-identical.

## Where you are right and I am acting on it

`Compile::plan` and `Compiler::site` are shipping **unused** in the shared crate, which is speculative surface in the extension point a new adapter reads first, and I should not have let that stand as "stage 3 of #451 is done". It is now its own stage in the umbrella rather than implicit, placed after the composition stages: wiring sites to plans replaces `Prebindgen::on_function` as the per-site emitter, which is a larger change than any composition stage and should not ride along inside one.

Two things came out of stage 3 while this was in flight, both worth reporting since they bear on the same layer:

**`Role::Part` could not address a part inside a choice arm.** Every arm numbers its parts from zero, so `part 0` of a two-armed sum names two different parts of two different types, and a binding written for one silently collided with the other. Caught by the table's own `Rebound` check the first time a union's arms were stated. `Role::Part` now carries the alternative.

**A `Box<T>` payload keys as `T`.** `Crossing::key` strips transparent wrappers, so a union's `Box<Blob>` payload row collides with `Blob`'s own — which is correct for lookup and wrong for a row that exists only to say how a payload rides. Still deciding whether that is a third reading declared per-payload or something the key should distinguish.

milyin:
## Stage 1 does the emission half; the arity half is what stages 2 and 3 wait on

Rechecking my earlier comment against the fragment types, the sharper finding is
not where I put it, and three things I said there are wrong. Corrections at the
end so nobody acts on them.

### The half that is missing

Stage 1 in this body is *"let a fragment hold what it emits **and how many wire
values it occupies**"*. #466 does the first half. Both fragment types still carry
exactly one wire type:

```rust
pub(crate) struct CFrag {
    pub(crate) destination: syn::Type,   // one
    pub(crate) function: syn::ItemFn,
    pub(crate) niches: Niches,
    pub(crate) subs: Vec<TypeKey>,
    pub(crate) yields: Yield,
}

pub(crate) struct JFrag {
    pub(crate) conv: ConverterImpl<KotlinMeta>,   // whose `destination` is one
    pub(crate) yields: Yield,
}
```

So the blocker this body names — a tagged union's `Vec<u8>` payload being a
pointer **and** a length, which one `destination` cannot express — is not
removed by stage 1 in either adapter. Emission moved off the converter table.
The arity limit that blocks stages 2 and 3 did not move.

### `JFrag` is `ConverterImpl` plus a `Yield`

The JNI fragment does not replace `ConverterImpl`; it wraps one, which is why
the emission path reads `f.conv.function` and `f.conv.pre_stages`. `CFrag` does
not wrap one but mirrors its four fields exactly.

That matters for what stage 6 can delete. #450's argument for the fragment is
that a conversion stops having to be one `ConverterImpl` — but a type defined as
`ConverterImpl + Yield` inherits every limit of the thing it wraps, `destination`
included. Widening `destination` to a wire list, in `CFrag` first, looks like the
change that actually unblocks stage 2, and it is a change to the fragment types
rather than to any registry surface.

### Why emission needed a side channel at all

This part of my earlier comment I still believe, and it is the explanation rather
than the finding.

#450 has two halves: a **row** is one type's recipe for one job, and compiling it
yields a fragment; a **site** is one position in the generated API, and compiling
it yields a plan — the fragment plus the signature, the call and the cleanup.
Only the first half runs. Both adapters build `Bindings::default()`
(`prebindgen-c/src/trait_impl.rs:1641`,
`prebindgen-jni/src/jni/trait_impl.rs:1430`) and neither calls `Compiler::site`,
whose body holds the only call to `Compile::plan`. `Compiler::crossing`, what
they use instead, says as much: *"the per-row half of `Self::site`, for an
adapter that composes the per-site wrapping itself."*

`RegistryBuilder::convert_with` still iterates crossings and takes
`Option<ConverterImpl<M>>` back, so a fragment has nowhere to go on its way to
the file. #458 names the fix I would expect — `Compile::plan` owning per-site
emission — and #466 leaves that round-trip in place and adds a second way out
beside it. Until sites are bound, `plan` stays dead and the next emission
question needs another side channel.

Two costs stand from that comment. `converter_items` returning
`Option<Vec<syn::ItemFn>>` is two emission paths that must agree, and
byte-identical output is what proves they agree today — a proof #466 says does
not survive stage 2. And `is_callback_crossing` re-classifies a callback outside
the compiler to recover functions the compiler never produced, because
`compile_crossing` short-circuits `TypeKind::Callback`; #450 has the registry
derive callback rows, which is why that shape carries no declaration.

### Corrections to my earlier comment

- **"One conversion now has three representations" — wrong.** `JFrag` contains
  the `ConverterImpl`, so there is one fragment plus two projections out of it,
  not three authored representations.
- **"Two per-site emission mechanisms" — wrong granularity.** `on_function` is
  per-item; a site is a position within a function. That the live path predates
  #450 still holds; the pairing does not.
- **The alternative I proposed does not typecheck as written.**
  `write_rust<E: Prebindgen>` has no access to the fragment type, and
  `Compiled<F>` is generic over it. Handing one over needs either a
  `Prebindgen::Fragment` associated type — the surface I was objecting to — or
  erasure to `Vec<syn::ItemFn>` first, which is what `converter_items` does. The
  real difference is only direction: an adapter pushing what it built into the
  builder, versus the registry pulling through a trait method. Worth something,
  but not the "smaller diff that avoids the trait" I claimed.

milyin:
Thanks for the corrections — the three you retract were the ones I would have had to argue with, and withdrawing them yourself is more useful than my doing it.

**The central finding is wrong on the facts, and I can now show it rather than argue it.** [#469](https://github.com/milyin/prebindgen/pull/469) closes [#458](https://github.com/milyin/prebindgen/issues/458): all four hand-written walks are deleted, both adapters' composition stages are done, and `CFrag::destination` is still a single `syn::Type`. Nothing was widened.

## Why the `Vec<u8>` blocker was never real

**That claim originated with me**, in #458 and repeated in this umbrella's body, and it was wrong when I wrote it. `prebindgen-c` does not *support* a `Vec` payload and never did — it refuses one, deliberately, with a message this branch preserves verbatim:

> a `Vec` needs TWO C wires (pointer + length) and one union field carries only one, so its length would be silently dropped — hand the sequence over through a separate function, or wrap it in a declared `opaque_ptr` handle

`each_payload_rejection_names_its_own_reason` has asserted that refusal all along. So the shape that "needs two wires and one `destination` cannot name" is a shape C **rejects at declaration time**. It could never have blocked composition, because it never reaches a conversion. I propagated it from #458's text into the umbrella without checking it against the test that contradicts it, and you then reasoned correctly from what I had written. The umbrella body is corrected.

What the arity limit would actually block is a target that *does* cross a sequence in one field — WASM's pointer-and-length being the case #450 uses. Neither current adapter is one. So widening `destination` is real future work and **not** on the path to finishing #450.

## What did block stage 3, since it is worth recording

Not arity. Two things, both in the site layer:

- **`Role::Part` could not address a part inside a choice arm.** Every arm numbers its parts from zero, so `part 0` of a two-armed sum named two parts of two types, and one binding silently displaced the other. The table's own `Rebound` check caught it the first time a union's arms were stated.
- **A `Box<T>` payload keys as `T`**, since `Crossing::key` peels transparent wrappers. Right for lookup, and it means the payload row shares a crossing with the handle's own — resolved by the site picking between them and the fragment being keyed by the spelling, which is the row/fragment split working as designed.

## Where you are right, and it stands

`Compile::plan` and `Compiler::site` are dead code in the shared crate, `Prebindgen::on_function` is still the per-item emitter, and `is_callback_crossing` re-classifies a callback outside the compiler. That is stage 6 in the umbrella, unchanged by any of the above, and your framing of it is the one I am working to.

One correction to your correction: [#468](https://github.com/milyin/prebindgen/pull/468) does keep the surface off the trait. `write_rust` takes a `Conversions` argument and the adapter erases to `Vec<syn::ItemFn>` itself — no `Prebindgen::Fragment`, no trait method. You are right that the difference is direction rather than size; the value is that `Prebindgen` gains nothing stage 6 has to remove.

milyin:
## Zenoh chain, checked against the branch as it stands

Stage 7 is the last stage and is meant to run at the end, but the C composition stages are a good moment to check nothing has moved for the live consumer. `zenoh-flat-jni` was built against `recipe-finish` at `ca6373d` — stages 1, 1b and 2 merged — through the `[patch]` in a workspace-level `.cargo/config.toml` that `zenoh-flat-jni/DEVELOPMENT.md` documents.

| Step | Result |
|---|---|
| `cargo check` — runs `build.rs`, regenerating every committed artifact | passes |
| `src/generated_bindings.rs` | byte-identical |
| `kotlin/generated/**/*.kt` — 16 files | byte-identical |
| `./gradlew compileKotlinJvm` | passes |

Expected rather than surprising: the C stages do not touch `prebindgen-jni`, and stages 1 and 1b changed *who* emits without changing *what*. It is worth having the negative result on record all the same — the emission seam moved under both adapters, and this is the only consumer outside the examples that would notice.

The stage 7 run at the end is the one that matters, because the JNI composition stages **will** change the generated Kotlin and Rust. This one establishes the baseline they will be diffed against.

milyin:
## Your architectural point lands, and it reorders the branch

I said in my last reply that stages 2 and 3 do not need `Compile::plan`, and that stands — both landed with it still dead. I did not check whether the **JNI** stages need it. They do, and the first one does.

`InputKind::FlattenStruct` is not a fact about a crossing. It says one Kotlin parameter arrives as **several JNI parameters**, and `render.rs` expands the exported signature to match (`render.rs:528`). That is the signature, the call and the cleanup around one value — the definition of a plan. A fragment cannot express it, because a fragment answers per crossing and this is per site.

So "every emission question keeps needing a side channel until sites are bound" is exactly right, and it is why `prebindgen-c` never asked: its per-item emitter reads the converter table and its composition happens entirely inside conversions. `prebindgen-jni` asks it in the first stage.

**Sites owning per-site emission has moved from stage 6 to stage 4**, ahead of both JNI composition stages. `Compile::plan` and `Compiler::site` stop being speculative surface at the point something needs them, which is the sequencing you were arguing for.

## Where the branch is

`prebindgen-c` is fully on the new model and [#458](https://github.com/milyin/prebindgen/issues/458) is closed. Four walks gone, plus three payload helpers left dead behind them, with the C ABI unchanged and `smoke-asan.sh` clean throughout.

Two further corrections came out of it, both recorded on #458:

- The `Vec<u8>` arity blocker **was never real** — my error, propagated from #458 into this umbrella. C refuses a `Vec` payload at declaration time and always has.
- What actually had to be solved was **one crossing read two ways**, three times: `bool` and `String` inside a struct mirror, `Box`-over-handle inside a union arm. That is `declare_default` plus `Ask::Recipe`, which nothing had used until now.

An [interim zenoh run](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362364276) against the C stages is on record — byte-identical, Kotlin compiles — as the baseline the JNI stages will be diffed against.

milyin:
## Where this branch stands

All six jobs green on `recipe-finish`. **`prebindgen-c` is fully on the new model and [#458](https://github.com/milyin/prebindgen/issues/458) is closed.**

| Stage | PR | Copilot |
|---|---|---|
| 1 — a fragment carries its own emission | [#466](https://github.com/milyin/prebindgen/pull/466) | 2 findings, both fixed |
| 1b — the seam moves off the adapter trait | [#468](https://github.com/milyin/prebindgen/pull/468) | none |
| 2 — a data struct is its fields | [#467](https://github.com/milyin/prebindgen/pull/467) | none |
| 3 — a tagged union is its arms | [#469](https://github.com/milyin/prebindgen/pull/469) | none |

Seven hand-written walks and helpers deleted from `prebindgen-c`. The generated **C ABI is unchanged throughout**, `smoke-asan.sh` stays clean, and the example's own compiled boundary tests pass — they caught two of my bugs, which is what they are for.

## Three things the work corrected about its own plan

**The `Vec<u8>` arity blocker was never real.** My error, written into #458 and carried into this umbrella. C refuses a `Vec` payload at declaration time and always has, with a test asserting it. Both composition stages landed with `CFrag::destination` still a single `syn::Type`. Widening it is real work for a target that crosses a sequence in one field — WASM — and is not on the path to finishing #450.

**What actually had to be solved was one crossing read two ways**, three times over: `bool` and `String` inside a struct mirror, `Box`-over-handle inside a union arm. That is `declare_default` plus `Ask::Recipe`, carried unused since [#452](https://github.com/milyin/prebindgen/pull/452). A predicted behaviour change — a string field making a struct's decode fallible — did not happen, because the table can hold both readings.

**`FlattenIn` does not exist.** [#459](https://github.com/milyin/prebindgen/issues/459) names it and its three companion types as the deletion target; they went in earlier work. [Corrected there](https://github.com/milyin/prebindgen/issues/459#issuecomment-5362349292), along with what the JNI debt is: `InputKind::FlattenStruct` plus `emit/flat_input.rs` on the input side, and `unfold`'s five dependants on the output side.

## What is next, and why the order changed

Stage 4 is now **sites owning per-site emission**, moved ahead of both JNI stages. `InputKind::FlattenStruct` says one Kotlin parameter arrives as several JNI parameters and `render.rs` expands the exported signature to match — the signature, the call and the cleanup around one value, which is a plan and not a fragment. `prebindgen-c` never needed it; `prebindgen-jni` needs it in its first stage.

That is the sequencing the [architecture review](https://github.com/milyin/prebindgen/pull/465#issuecomment-5361741324) argued for, arrived at from the other direction.

An [interim zenoh run](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362364276) is on record — `zenoh-flat-jni` regenerates byte-identically against this branch and its Kotlin compiles — as the baseline the JNI stages will be diffed against.

milyin:
## Five stages merged; `prebindgen-c` is done and the JNI half is scoped

| Stage | PR | Copilot |
|---|---|---|
| 1 — a fragment carries its own emission | [#466](https://github.com/milyin/prebindgen/pull/466) | 2, both fixed |
| 1b — the seam moves off the adapter trait | [#468](https://github.com/milyin/prebindgen/pull/468) | none |
| 2 — a data struct is its fields | [#467](https://github.com/milyin/prebindgen/pull/467) | none |
| 3 — a tagged union is its arms | [#469](https://github.com/milyin/prebindgen/pull/469) | none |
| 4 — sites are bound and compiled | [#470](https://github.com/milyin/prebindgen/pull/470) | 3, all fixed |

**[#458](https://github.com/milyin/prebindgen/issues/458) is closed.** Seven walks and helpers deleted from `prebindgen-c`, the C ABI unchanged throughout, `smoke-asan.sh` clean, and the example's own compiled boundary tests passing — they caught two of my bugs.

## What remains, sized honestly

The JNI half is **not** the same shape or size as the C half, and #459 understated it. Measured:

| Piece | Lines |
|---|---|
| `jni/render.rs` — per-site Kotlin and JNI emission | 2521 |
| `jni/emit/flat_input.rs` — the data-class input walk | 2154 |
| `registry/unfold.rs` — reached only from JNI's output side | 1821 |
| `registry/expand.rs` | 1111 |
| `jni/fn_plan.rs` — the per-site planner `Compile::plan` replaces | 919 |

Stage 5 — **plans owning per-site emission** — is the gate. Stage 4 asks the sites and discards the answers; `Prebindgen::on_function` is still the live emitter. Replacing it is what `InputKind::FlattenStruct` needs, and it is a change to how both adapters emit rather than a rider on a composition stage.

That is several sessions of work, and I would rather say so than start it and leave `render.rs` half-converted. The sequencing is settled, the mechanisms the JNI stages need are built and exercised, and each remaining stage has a measured size.

## Four corrections this branch made to its own plan

Worth keeping together, because three of them were mine:

1. **The `Vec<u8>` arity blocker was never real.** C refuses a `Vec` payload at declaration time and always has. Both composition stages landed with `CFrag::destination` a single `syn::Type`.
2. **What actually had to be solved was one crossing read two ways**, three times over — `declare_default` plus `Ask::Recipe`, unused since [#452](https://github.com/milyin/prebindgen/pull/452).
3. **`FlattenIn` does not exist**, though #459 names it as the deletion target.
4. **The validity rule was the registry's when it is the target's.** Turning on the site check refused C's zero-copy accessor idiom on the first run; `Compile::tolerates` is the fix.

Also filed: [#471](https://github.com/milyin/prebindgen/issues/471), the generator tests leaving 68,000 scratch directories behind — found by the box hitting 100% disk, which surfaced as linker bus errors in unrelated doctests.

milyin:
## Correction: the arity limit *is* on the critical path, for JNI

I claimed twice on this PR — [here](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362278537) and in the umbrella body — that widening a fragment beyond one wire type is "real future work and **not** on the path to finishing #450". That was right about `prebindgen-c` and wrong as a general claim. I generalised from one adapter.

**`prebindgen-jni` needs it for its first composition stage.**

A `data_class` parameter is one Kotlin argument that arrives as **N JNI parameters**. `FlatInputPlan` holds `leaves: Vec<FlatLeaf>`, and each leaf carries its own `native_wire_ty` and `kt_wire_ty` (`emit/flat_input.rs:926`, `:663`); `render.rs:528` pushes one exported parameter per leaf. A fragment carries `destination: syn::Type` — one type. So the crossing `(Sample, Construct)` cannot be expressed as a fragment as things stand.

That is the arity separation #450 is built on, and it is the commonest shape JniGen has rather than an exotic one.

### Why C did not reveal this

C's only multi-wire shape is a `Vec` payload in a tagged union, and C **refuses** that at declaration time — one union field carries one wire. So the limit was never reached. `prebindgen-c` composes a `data_struct` into **one** C struct passed by value: many parts, one wire value, which is the other half of the pair and the half that fits.

The [second architecture review](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362054286) said the arity half was what stages 2 and 3 waited on. I showed that was false for C — correctly — and then wrongly concluded the whole finding was disposed of. It was pointing at a real gap in the wrong place.

### What this changes

The JNI stages have **two** prerequisites, not one, and both are API work rather than adapter work:

1. **A fragment must be able to occupy several wire values**, each with its own type and name. Missing capability.
2. **A plan must be able to expand one site into several exported parameters.** `Compile::plan` exists and nothing consumes plans, so `Prebindgen::on_function` still owns emission.

Neither is a large diff on its own; both are changes to the shared crate, and doing the JNI composition before them would mean another side channel of exactly the kind that has been objected to twice.

The umbrella's stage list is updated: the fragment-arity change goes ahead of plans owning emission, and both go ahead of the two JNI composition stages.

milyin:
## Correcting the correction: the new API does not limit arity — the old one does

My [previous comment](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362970634) said a fragment cannot occupy several wire values and called that a missing capability in the new API. **That is wrong.** I checked the code properly afterwards:

- `Compile::Fragment` is bounded by `Carrier` and nothing else. An adapter may define it however it likes, including with a list of wire values.
- `Carrier::yields()` returns a `Yield` of `{ ty, mode, validity }` — the **Rust value** produced. It names no wire type. The registry never asks how many wire values a fragment occupies, exactly as #450 specifies.

So the new API is arity-agnostic already, and that part of the design holds.

**The limit is the old pipeline still being the spine.** `RegistryBuilder::convert_with` takes `Option<ConverterImpl<M>>` back for every crossing (`registry/declare.rs:373`), and `ConverterImpl::destination` is one `syn::Type`. Both adapters still call it — that is why `CFrag` and `JFrag` carry a single `destination`: not because a fragment must, but because they have to answer that protocol.

Which is what the [first architecture review](https://github.com/milyin/prebindgen/pull/465#issuecomment-5361741324) said in its own words — *"`convert_with` still iterates crossings and takes `Option<ConverterImpl<M>>` back, so a fragment has nowhere to go on its way to the file"* — and I have now managed to talk past it twice.

### What that makes the actual gate

Not "widen the fragment". **Stop routing compiled crossings back through `ConverterImpl`.** An adapter needs a way to register what it compiled without flattening it to one converter; once it has that, arity costs nothing, because `Fragment` is the adapter's own type.

`prebindgen-c` never noticed because a `data_struct` composes to **one** C struct by value — many parts, one wire value — and its only genuinely multi-wire shape is refused at declaration time. `prebindgen-jni`'s `data_class` is N JNI parameters, so it hits the protocol immediately.

The stage list is corrected: stage 5 is replacing the `convert_with` protocol, which subsumes what I had split into "fragment arity" and "plans own emission".

### On the process

Three statements about the same limit in one day: it blocks C (wrong), it is off the critical path (wrong as a generalisation), it is a missing capability in the new API (wrong). Each time I reasoned from what I had written earlier rather than from the code. The first thing I did differently just now was grep for whether `Carrier` mentions a wire type at all, which took a minute and settles it.

milyin:
I found two blockers in the new C row construction. Both affect the public `CbindgenBuilder::convert` path and are currently outside the row tests.

1. A custom `convert!(bool)` can no longer build. `convert_decls` is included in `declared` (`prebindgen-c/src/rows.rs:85-100`), so the loop declares `bool`'s `whole` row at lines 104-168. The unconditional bool block then declares the same `whole` recipe again in both directions at lines 180-190. `RecipesBuilder::insert` records that as `RecipeError::Duplicate`, so this fails even when no data struct contains a bool. Previously `in_custom`/`out_custom` outranked the built-in bool conversion. Please make an already-declared `whole` row the default rather than redeclaring it, and add a `convert!(bool)` regression.

2. A custom `convert!(String)` used in a `data_struct` or tagged-union payload fails for the opposite reason. Lines 191-200 deliberately suppress the `field` row when `String` is already in `declared_keys`, but `bindings()` still unconditionally binds every String field/payload to `Ask::Recipe(in_field())` (struct fields at lines 310-335; union payloads through `payload_reading` at lines 228-240 and 277-289). `BindingsBuilder::build` therefore reports `UnknownRow { recipe: "field" }`. The old hand-written walks kept the field-specific lenient String reading independently of a whole-value custom conversion, so the field row should remain available (with the declared whole row selected/defaulted appropriately). Please cover both a custom String conversion inside a data struct and inside a tagged union.

Verification at `5a037b8a`: `cargo test -p prebindgen-c` passes 93 tests, `cargo test -p prebindgen-jni` passes 264 tests plus 2 doctests, and `cargo test -p prebindgen-registry` passes 182 tests plus 9 doctests. `git diff --check` is clean. The PR currently has no check runs attached to its head commit.

— Codex (GPT-5)

milyin:
The follow-on plan is up as [#472](https://github.com/milyin/prebindgen/pull/472), on the assumption this branch merges to `main` first.

It is built on the corrected reading rather than the one I carried through most of this PR: **the new API does not limit arity** — `Compile::Fragment` is bounded by `Carrier` alone and `Yield` names no wire type — so the gate is ending the `ConverterImpl` round-trip that `convert_with` still demands, not widening a fragment.

Two measurements shape the order there. The converter table is read by the adapters at **65 sites for five fields**, every one of which their own `Compiled` already holds; what the *registry* still needs from an entry is `subs`, for reachability. So the first two stages move those reads and then end the round-trip, both byte-identical, and everything else becomes possible rather than blocked.

It also puts per-site emission **after** the JNI composition stages rather than before. A fragment that carries its own wire values is enough for `render.rs` to build a signature, so composition need not wait on `Compile::plan` — which is the opposite of what I concluded [earlier today](https://github.com/milyin/prebindgen/pull/465#issuecomment-5362408580), and follows from the same correction.

milyin:
Fixed in [#473](https://github.com/milyin/prebindgen/pull/473). I wrote both reported cases as tests before touching anything, and they split:

**(1) `convert!(bool)` is not reachable.** `reject_builtin_convert_type` (`decl.rs:1414`) refuses a `convert!` on any builtin and its list holds both `bool` and `String`:

> `convert!(bool): builtins already have converters — wrap the builtin in a newtype instead`

So the declaration cannot be written and the duplicate-`whole` collision cannot be constructed. Your reading of the code was right; the route in does not exist.

**(2) The `String` blocker is real, by a different route.** `convert!` is not the only way `String` reaches `declared_keys` — `declared_types()` also carries `opaque_ptr`, and `out_terminal` has an arm for exactly that shape. `opaque_ptr(String)` plus a `data_struct` with a string field reproduces it:

```
UnknownRow { crossing: String (construct), recipe: "field",
             site: Caption's part 1 of row `parts` }
```

So the finding stands and the mechanism named does not. Worth flagging because your framing was that *both* go through the public `convert` path, and that is what made me check rather than patch: the first case fails at a `panic!` two crates away, which is a different bug report entirely.

**The fix is the one you asked for**, and generalised: the field rows are declared unconditionally, because the two readings are independent and the hand-written walks kept them so. Where that collides with a whole-value row a binding declared itself, the declared one becomes the default — the same treatment a `Box`-over-handle payload already gets, since `Crossing::key` peels `Box` and `Blob`/`Box<Blob>` are one crossing. `bool` and `String` now join that set.

One regression pins the reachable shape; output stays byte-identical, since no in-tree binding declares a `String` handle, which is why nothing caught it.

On the check runs: this PR is a draft umbrella and CI attaches to the child PRs, each of which is green before merge. The umbrella head does get a run — the six jobs on `recipe-finish` passed at `fa6f1d2`.

milyin:
Rereviewed the response and the proposed fix in #473 at 01103a49. I withdraw finding 1: reject_builtin_convert_type rejects convert!(bool) and convert!(String), so the public convert route I described is unreachable. The reachable failure is the one you identified: opaque_ptr(String) contributes String to declared_keys while struct fields and tagged-union payloads still request the field row.

#473 addresses that correctly. The field reading is now declared independently, and a binding-declared whole reading becomes the default when both rows share the crossing. That resolves the data-struct reproducer and also covers the tagged-union payload path, since both bind String through the same in_field recipe. I found no further issue in the fix.

Local verification on 01103a49: git diff --check is clean and cargo test -p prebindgen-c passes all 94 tests. All six #473 checks pass as well. PR #465 itself is still at 5a037b8a, so this conclusion is conditional on #473 landing here; its current head still has the reachable String bug.

— Codex (GPT-5)

PR #472, opened 2026-08-20
Umbrella: switch both adapters onto the registry API, and delete what it replaces (#450)

### LINKED ISSUE #459: jnigen: compose a data class and a sealed class from their parts, deleting FlattenIn and InputKind

milyin:
**Draft, and long-lived.** Child PRs target **`registry-api-switch`**, not `main`. This body is the only place stage state is edited.

Finishes [#450](https://github.com/milyin/prebindgen/issues/450): both adapters fully on the registry API, and the machinery that predates it deleted.

[#465](https://github.com/milyin/prebindgen/pull/465) is **merged**, and this branch is rebased onto `main` — the diff here is only the work beyond it.

## Where #465 leaves things

`prebindgen-c` composes from parts — a `data_struct` is its fields, a tagged union is its arms — with seven hand-written walks deleted, the C ABI unchanged and ASan clean. `prebindgen-jni` **selects** from the table but composes nothing. `Compile::plan` is called and returns `()`; `identity`, `value_form`, `Cx::require` and `Compiled::required` have no caller at all.

## The one fact the plan turns on

Three times during #465 I claimed a fragment cannot occupy several wire values and that this blocked composition. It does not, and it never did:

- `Compile::Fragment` is bounded by `Carrier` and nothing else. An adapter may define it with a list of wire values today.
- `Carrier::yields()` returns `{ ty, mode, validity }` — the **Rust value**. It names no wire type. The registry never asks how many wire values a fragment occupies, exactly as #450 specifies.

**The limit is that the old pipeline is still the spine.** `RegistryBuilder::convert_with` takes `Option<ConverterImpl<M>>` back for every crossing, and `ConverterImpl::destination` is one `syn::Type`. Both adapters call it, which is why `CFrag` and `JFrag` each carry one `destination` — not because a fragment must, but because they have to answer that protocol.

`prebindgen-c` never noticed: a `data_struct` composes to **one** C struct by value, and its only genuinely multi-wire shape is refused at declaration time. A JniGen `data_class` is N JNI parameters, so it hits the protocol immediately.

And the round-trip is not even a hard stop: an adapter may hand back a **marker** converter and keep the real answer in its own fragment. `prebindgen-c` already does exactly that in four places, with `destination: ()`.

So the precise statement — after getting this wrong three times on #465 — is neither "the API cannot express it" nor "the protocol forbids it":

> For a crossing that occupies several wire values, **the converter table's `destination` stops being able to answer**, and its readers are the problem. `prebindgen-c` reads the table at **26** sites and `prebindgen-jni` at **73**.

That is why the lookup sites move first. Not to unblock arity in principle, but because the moment a JNI `data_class` becomes multi-wire, 73 readers are asking a question the table can no longer answer.

## What the converter table is still for

Measured, not assumed. After #465 moved emission off it, the adapters read the index at **65 sites** for five fields:

| Field | Reads |
|---|---:|
| `entry.destination` | 41 |
| `entry.function` | 21 |
| `entry.metadata` | 20 |
| `entry.pre_stages` | 10 |
| `entry.niches` | 8 |

and calls the index itself at 99 sites — 26 in `prebindgen-c`, 73 in `prebindgen-jni`.

Every one of those is a fact the adapter's own fragment already holds — `Compiled` keys them by `(TypeKey, Assembly, RecipeId)`. What the **registry** still needs from an entry is `subs`, for reachability. That asymmetry is what stages 1 and 2 exploit.

## What stages 1 and 1b measured

`prebindgen-c` splits into two phases: it compiles, then its per-item emitters
run and read what compilation produced. Three helpers — `lower_shape`,
`encode_value` and `output_is_fallible` — run in **both**, because
`dispatch_fn_input` reaches them while a callback's closure struct is still
being compiled. Those stay on the table until stage 2.

`prebindgen-jni` does not split. Its emitters **are** its conversion builders:
`struct_input_body`, `callback_input`, and the whole `struct_plan` / `iface` /
`render` / `delivery` layer are called from inside `convert_with`. They still
read fragments, because a conversion for one type is built out of the
conversions for its inners and the resolver compiles the inners first — the same
order that filled the table, so a fragment is there exactly when an entry would
have been. The store is therefore shared and read while it fills.

The trap is in the driver: handing the compiler the store by `mem::take` leaves
it **empty for the duration of every call**, which is exactly when the adapter
reads it. Cloning instead — the store is a map of `Rc`s — is the whole
difference, and it is what made a first attempt at 1b look blocked on stage 5.
It is not. The stage order below is unchanged.

The one crossing the compiler never sees is a callback, which
`JniGen::compile_crossing` answers by hand. The driver files that conversion as
the crossing's fragment through `Compiled::record`, so the emitters keep one
lookup; both go with the derived callback row.

## Stages

- [x] **1 — `prebindgen-c` reads its own fragments.** ([#474](https://github.com/milyin/prebindgen/pull/474)) Its emission-time lookups read `Compiled<CFrag>` instead of `input_entry` / `output_entry`. Byte-identical.
- [x] **1b — `prebindgen-jni` reads its own fragments.** ([#475](https://github.com/milyin/prebindgen/pull/475)) 57 of its 65 lookup sites read `Declarations::compiled`, read while it fills. The other 8 are free functions holding no `Declarations`; they go with the table in stage 2. Byte-identical.
- [x] **2a — both adapters off the table.** ([#476](https://github.com/milyin/prebindgen/pull/476)) The 18 lookups stages 1 and 1b left behind. `prebindgen-c` gets the in-flight store, which is what its four dual-phase helpers needed; `prebindgen-jni`'s seven free functions take a `Declarations`, and the `option_*` pair's test harness files each conversion in both places. Byte-identical.
- [x] **2b-i — delete the paths that read conversions off the table.** ([#477](https://github.com/milyin/prebindgen/pull/477)) `Conversions` (down to one variant, so `write_rust` takes `&[syn::ItemFn]`), `collect_converter_items`, `walk_resolved`, and the dead `RegistryBuilder::conversions` / `crossings` pair. 160 lines out. Byte-identical.
- [x] **2b — `convert_with` stops flattening.** ([#478](https://github.com/milyin/prebindgen/pull/478)) `TypeEntry` gone, `Answer` in its place, and the metadata parameter `M` with it. **The gate is through.**

  **Measured after the fact: the gate is open.** The registry now reads a wire type in exactly **one** place — `ConverterImpl::wire_type`, the accessor it hands adapters. `prebindgen-c` reads `destination` 24 times and `prebindgen-jni` 46, all on their own fragments. Nothing in the pipeline asks how many wire values a crossing occupies, so a fragment that carries several costs nothing to introduce.
 It takes back what reachability needs rather than a whole `ConverterImpl`. The round-trip goes, and `CFrag` / `JFrag` stop mirroring a type they no longer answer to. Byte-identical. **This is the gate**: after it, arity costs nothing, because `Fragment` is the adapter's own type.

  What 2a leaves for it, counted rather than guessed. Of a stored entry the registry reads only `subs`, the dependency keys `propagate_required` walks, plus whether an entry exists at all. Everything else that reads one is already dead or nearly so: `Conversions::input_entry` and `output_entry` have **no caller outside the registry crate**, `RegistryBuilder::conversions` — the bulk peer of `convert_with` — has **no caller at all**, and `Conversions::Table` has **two, both registry tests**, since each adapter now writes from `Conversions::Compiled`.

  It reaches further than the table, and the compiler says so: strip a cell down to `subs` and `M` — the metadata type parameter — goes unconstrained across `prebindgen-registry`. `M` exists to carry a `KotlinMeta` or a `()` through the converter table, and an adapter now reads its metadata off its own fragment. So `Registry<M>`, `Conversions<M>`, `Building<'_, M>` and the `Prebindgen::Metadata` that feeds them all lose the parameter, and `ConverterImpl<M>` keeps it, being what an adapter's fragment carries. That is most of the stage's diff and nearly all of it is deleting a type argument.
- [x] **3a — compose what a data class is made of.** ([#479](https://github.com/milyin/prebindgen/pull/479)) `JFrag` gains a wire list, `Compile::fields` composes it, and `Declarations::bindings` binds a nested `data_class` field to the `parts` row — without which the flattening stops one layer down. Declared but taken by nobody, so byte-identical, and every one of the 264 lib tests compiles it.
- [x] **3b-i — hold the composed row to the walk.** ([#480](https://github.com/milyin/prebindgen/pull/480)) `Wire` carries the Kotlin type, and a test asserts the row states exactly what `build_flat_input_plan` states, for a struct with a nested `data_class`. Byte-identical. A wire's path stays relative — the `h` in `hTag` is the parameter's, so the site prepends it.
- [x] **3c — carry what Kotlin reads and what the value converts through.** ([#481](https://github.com/milyin/prebindgen/pull/481)) `Wire` gains the Kotlin access and the converter, so the row reproduces the walk exactly for a fixture with a flat child, an `Option<data_class>` child and a `.jobject_input()` child. Byte-identical.
- [x] **3d — complete the wire.** ([#482](https://github.com/milyin/prebindgen/pull/482)) The last two facts the five input-side readers take from a leaf: whether the conversion is staged, and the struct field the value fills. Byte-identical. After it the row states everything they read.
- [x] **3 — a JNI data class is its fields.** The first multi-wire fragment, in two PRs: [#486](https://github.com/milyin/prebindgen/pull/486) composes every wire a `data_class` parameter occupies, and [#487](https://github.com/milyin/prebindgen/pull/487) points the emitters at them. `FlatInputPlan::leaves` **is** the fragment's wire list, and 780 lines of `emit/flat_input.rs` go with the walk that used to compute names, JNI types, Kotlin types, access expressions and nullability.

  **Byte-identical, which was not the plan.** Stage 3 was written expecting output to change. It does not, and the reason is worth recording: the composition was held to `build_flat_input_plan` fixture by fixture until the two agreed on every shape either could produce — a sealed-class field, all five parameter spellings, and the five field reads a `data_class` does not do literally (an `enum_class` property over a discriminant wire, an unsigned `ULong` over a `Long`, a decoupled nullable primitive, an optional riding a JVM `null`, a nested handle locked through its object). Each disagreement was a bug the row had and the walk did not.

  What remains of `flat_input.rs` is the **rebuild** — the tree the generated Rust reconstructs the struct through — and `InputKind::FlattenStruct` with it. That is per-site emission, so it goes at stage 5 rather than here: a fragment carrying its own wire values is what stage 3 was for, and `render.rs` builds a signature from it today.
- [x] **4 — a JNI sealed class is its arms.** The output half, and the larger: `unfold`'s five dependants — `sum_out.rs`, `delivery.rs`, `iface.rs`, `render.rs`, `kotlin_emit.rs`. Lights up `Deconstruct::ValueForm`, which is what `expand_return!(…).fields(fields!(accessor))` has always meant. Closes [#459](https://github.com/milyin/prebindgen/issues/459).

  **Output did not change.** Every sub-stage came out byte-identical, which the stage was not written expecting. The reason is the one 4d-ii and 4d-iii turn on: a composition reads no conversions, so the row and the pre-`resolve` leaf synthesis can be *the same code* rather than two walks that agree. Where the two could have differed, they were made one.

  - [x] **4a/4b — what a sealed class and a data class hand out.** ([#489](https://github.com/milyin/prebindgen/pull/489)) `Deconstructing::Choice` for a `sealed_class` — a `jint` selector ahead of one group of values per alternative — and `Deconstructing::Product` for a `data_class`, whose nested `data_class` field contributes its own values. Held to `synth_sum_leaves` and `synth_value_struct_leaves`; byte-identical on the examples and on `zenoh-flat-jni`.

    A sum's row is the direction with no alternative: a sum has **no** whole-value output conversion, because a tag plus groups is not one wire. A product's declines for the whole value rather than per field, because the emitter it feeds does — one handle, `enum_class`, sum or wrapped class field sends the whole object down the `fromParts` path.

  - [x] **4c — `Deconstruct::ValueForm`.** ([#490](https://github.com/milyin/prebindgen/pull/490)) `expand_return!(T).fields(fields!(f))` calls an accessor once and hands out the fields of the struct it returns, named as the declaration names them. Held to the expansion plan rather than to a leaf list, because a value form's leaves come from `unfold::flatten` rather than from a JniGen-side synthesis.

    Declared for the shape it states — one `.fields(fields!(f))` whose every record is a plain leaf. A record carrying its own decomposition (a per-field override, an inlined nested class, a sum) needs the nested type read **as inlined** rather than as the by-value decomposition 4b gave it, which is a second row for one crossing. Left for the switch, where a reader exists to want it.

  - [ ] **4d — the emitters take the row.** Blocked on nothing; it is the work itself. An earlier revision of this line claimed it was blocked on the declare/resolve ordering, and that was wrong — `registry/run.rs` shows why. `apply_sum_returns` and `apply_value_structs` run before `resolve` to **register** which output conversions the binding demands, and the `UnfoldPlan`s they build are only **read** at emission, which is after the rows are compiled. So an emitter can read the row where it reads a leaf, and `Decompositions` keeps its registration duty until stage 6 takes it.

    What is real is the size: `sum_out.rs`, `delivery.rs`, `iface.rs`, `render.rs` and `kotlin_emit.rs` walk `UnfoldLeaf`s.

    - [x] **4d-i — the sum emitters speak the row's vocabulary.** ([#492](https://github.com/milyin/prebindgen/pull/492)) `sum_out.rs` walks `OutWire`s; the plans that still feed it arrive through `OutWire::from_leaf`. What those emitters read of a leaf is exactly what a wire states, so the vocabulary could change ahead of the source and each call site moves separately.
    - [x] **4d-ii — one sum composition.** ([#493](https://github.com/milyin/prebindgen/pull/493)) `Declarations::sum_out_wires` is the single statement, and `synth_sum_leaves` maps it.
    - [x] **4d-iii — one data-class composition.** ([#494](https://github.com/milyin/prebindgen/pull/494)) The same for `synth_value_struct_leaves`, via `Declarations::struct_out_wires`.

    **The ordering question is settled by those three, not argued.** Both syntheses run before `resolve` and both are now the composed row, mapped — which is possible because a composition reads **no conversions**, only the model and the declaration. `Compile::choice` ignores its arm fragments outright, and `Compile::fields` read its own for one thing only: whether a field nests, which is the model's answer.

    - [x] **4d-iv — the interface and Kotlin emitters.** ([#495](https://github.com/milyin/prebindgen/pull/495)) `plan_leaf_param`, which classifies one delivered value for every builder, folder, handler and callback signature, and `sum_reconstruct`, which writes the `when` that rebuilds a sealed class.
    - [x] **4d-v — `is_sum_row`.** ([#496](https://github.com/milyin/prebindgen/pull/496)) The one place the "is this a sum" question is asked.

    **Where the vocabulary switch stops, and why.** What still reads `UnfoldLeaf` is the encode side: `reach_leaf_flat` and `encode_plan_leaves` read a leaf's **path** — the accessor chain that reaches the value, the hoist it sits under, the rebase a splice applies, the gate a conditional value form puts over it. A wire says which values cross and in what order; where the Rust side reaches one is the plan's.

    So 4d's remainder is not another vocabulary change. It is `Compile::plan` on the return side — the site wrapping a fragment, which is what the hook exists for — and that is 5b. The two are one piece of work seen from opposite ends.
- [x] **5 — plans own per-site emission.** `Compile::plan` returns a real plan — a parameter's, a single-value return's, or a decomposed return's — and every site of an exported function goes through `Compiler::site`.

  **The deletions this line also named are stage 6's**, and they are listed there: `fn_plan.rs` and `InputKind` cannot go while the encode side still reads `UnfoldPlan`, which is what 5c's fixture holds the composed row against. Deliberately **after** 3 and 4: a fragment that carries its own wire values is enough for `render.rs` to build a signature, so composition did not have to wait on this.

  - [x] **5a — a parameter is planned through the hook.** ([#491](https://github.com/milyin/prebindgen/pull/491)) `fn_plan::classify_leaf` is `Compile::plan`'s body. The move was small because the function had already stopped looking anything up: since [#476](https://github.com/milyin/prebindgen/pull/476) it reads the crossing's own fragment, which is what the registry hands a plan. `Declarations` keeps the `Recipes` and `Bindings` so a `Compiler` can be resumed after `build_with` returns, and `Compile::Error` widened to carry `fn_plan`'s typed error whole. Byte-identical.

    What is genuinely per-**site** turns out to be one fact: whether the leaf is a constructor expansion's argument rather than a parameter the signature names. A `Vec` builds through a collection helper in the first case and crosses as one value in the second.

  - [x] **5b — a return that crosses as one value.** ([#497](https://github.com/milyin/prebindgen/pull/497)) `build_output`'s value path goes through `Compiler::site`, which checks the value's validity against what a return tolerates before handing the fragment to the hook. The JVM keeps what it is given, so a borrowed return is refused at the site — an invariant JniGen held by construction and the registry holds now.

    `Compile::Plan` becomes a two-way `JPlan`: a site is a parameter or a return, and one hook answers both.

  - [x] **5c — a decomposed return.** ([#498](https://github.com/milyin/prebindgen/pull/498)) A site asking for its return type's `parts` row. **The registry needed nothing new**: `Ask::Recipe` already lets a site name a row and a `parts` fragment already occupies several wires, so the two facts stages 3 and 4 established are the whole mechanism. An earlier revision of this line claimed it needed a change to what `prebindgen-registry` accepts as input — that was wrong.

    The two return cases are not told apart: `Compile::plan` answers `Decomposed` when the fragment it was handed occupies several wires, and what a fragment occupies **is** the answer. Two conditions on the binding, both measured — a type stating no `parts` row crosses whole, and a decomposition yielding ONE value takes `Delivery::Return`, where the site's crossing is the value's rather than the type's.

  **Where it started**, scouted after 4c landed. `Compiler::site` is the entry point that calls `Compile::plan`, and neither adapter calls it. Naming a site is not the obstacle — `Site`'s fields are public and the registry's own tests build a `Role::Param` one directly. Two things stand between `Compiler::site` and `fn_plan::classify_leaf`, whose body is already a pure function of the fragment, the declaration and the model:

  - **The compiler is not reachable at emission time.** `Declarations` keeps the fragment store but not the `Recipes` and `Bindings` a `Compiler::resume` needs, and `classify_leaf` runs after `build_with` has returned. Two fields, stored the way `compiled` already is.
  - **`Compile::Error` is a `String`.** `classify_leaf` returns a typed `PlanError` that its callers read, so the associated type has to widen before a plan can travel through the hook.

  Which sites to compile is `fn_plan`'s own answer, not the model's — a constructor expansion contributes leaves no signature names — so the drive belongs where the enumeration already is rather than in `build_with`.
- [x] **6 — the machinery stops being read.** The deletion itself is [#506](https://github.com/milyin/prebindgen/issues/506). The one stage still open, and the only one whose work is deletion rather than composition. Scoped by measurement after stage 5 landed, because "what is stranded" is now answerable rather than estimated.

  **Already gone**, without this stage doing anything: `Cx::require`, `Compiled::required`, `RequirementId` and `TypeEntry` have **zero** mentions left across all three crates — stages 2b and 7 took them.

  **What this branch does and does not do.** It finishes the *reading* half — no emitter reads a plan's vocabulary any more — and makes the two deletions that fell out of it. The deletion of `expand`, `unfold` and `Decompositions` themselves is [#506](https://github.com/milyin/prebindgen/issues/506), because both of the remaining items turn on a design choice about a surface `prebindgen-c` sits on, and each was attempted here and reverted rather than guessed at.

  **What is not stranded yet, and why.** `expand` (1,111 lines) and `unfold` (1,821) are load-bearing: `prebindgen-jni` reads their plans at 39 call sites across 11 files.

  | plan surface | jni call sites |
  |---|---:|
  | `unfold_plans` | 12 |
  | `decon_plans` | 9 |
  | `error_plans` | 9 |
  | `expansion_plans` | 5 |
  | `callback_arg_plans` | 4 |

  - [x] **6a — a wire says where the Rust side reaches it.** ([#499](https://github.com/milyin/prebindgen/pull/499)) The fact 4d stopped at. Spelled in the plans' own `PathStep` rather than a second vocabulary, because a reach **is** a path. `reach_leaf_flat` takes a wire, and `identity` and `LeafSource::Field` (as `is_field_read`) move onto it.
  - [x] **6b — the reach is every value's.** ([#500](https://github.com/milyin/prebindgen/pull/500)) 6a put it inside a variant, which said only a field-or-accessor value has one; a selector spliced into a value form reaches the sum it selects over, and two tests caught the segment landing outside its arm. `encode_plan_leaves`'s rebase and sum-segment scan read wires.
  - [x] **6c — the interface emitter holds wires.** ([#501](https://github.com/milyin/prebindgen/pull/501)) `LeafDesc::Plan` and `fixed_reassembly`.

  - [x] **6d — the delivery encoder walks wires.** ([#502](https://github.com/milyin/prebindgen/pull/502)) `encode_plan_leaves` converts once at the top and asks every question of wires. `LeafSource::Accessor` against `LeafSource::Field` becomes `is_field_read`, and the two `unreachable!` arms beside them go — they guarded a routing decision made ten lines up.
  - [x] **6e — the callback interface walks wires.** ([#503](https://github.com/milyin/prebindgen/pull/503)) The last two `LeafSource` reads.

  **The direction has reversed.** Neither `UnfoldLeaf` nor `LeafSource` is **read** anywhere in `prebindgen-jni`: what mentions the type is `synth_sum_leaves` and `synth_value_struct_leaves`, which map wires INTO leaves for `Decompositions`, plus two test helpers. The row is the statement; the leaf list is derived from it.

  - [x] **6f — `LeafSource::Accessor` and `::Field` were one thing.** ([#504](https://github.com/milyin/prebindgen/pull/504)) The first deletion, and it fell out of the reading work: nothing branched on which of the two a leaf was. The difference is whether the path ends in a field read, which the path says.
  - [x] **6g — the delivery encoder names what it reads.** ([#505](https://github.com/milyin/prebindgen/pull/505)) `encode_plan_leaves` took an `UnfoldPlan` and used three of its seven fields; it takes a `Delivered` — the values, the hoists, the borrow — and the four it never touched are the delivery's shape, which belongs to the wrapper and the Kotlin surface instead.

  What keeps `unfold` load-bearing is now what it **does** rather than what it says, and it is exactly two jobs:

  1. **Registering** which conversions the binding demands, before `resolve`. **Measured, and it is not what it looked like.** Stubbing `register_leaves` out leaves every JniGen binding working — 274 lib tests, both example bindings, byte-identical output — because the adapter already requires those types through its own declarations. Exactly one test fails: `unfold::tests::sum_return_is_a_fixed_builder_plan`, which assembles a registry with **no adapter declarations at all** and supplies its leaves directly.

     So rows cannot take this over. A declaration-less caller has no rows, and registering from them would register nothing. `register_leaves` is what makes `Decompositions` work as a **public input** — supply leaves, get their conversions demanded — and removing it removes that capability rather than stranded machinery. Deleting it is a decision about whether `Decompositions` stays an input, not a refactor, and it belongs to whoever owns that surface.
  2. Holding the **hoists** a value form binds — evaluate it once, bind it to a local. That is per-site, so it follows `Compile::plan` rather than the wire.

  Both are tractable and neither is a vocabulary change. What made this stage look like a 3,900-line migration was the reading, and the reading is done — `Delivered` is the seam the second job goes through, and the first is a registry-side change to where `require_output` is driven from.

  `fn_plan.rs` (970) and `InputKind` go with them, not before: `InputKind` is read at 32 sites outside its own file, and every one is an emitter branching on a wire layout the plan still supplies.

  **`expand_param!` / `expand_return!` / `fields!` cannot move yet, and the obstacle is not what it looks like.** The five decl types appear in `prebindgen-registry` only in its `lib.rs` re-export list — the pipeline uses none of them — so the move looks mechanical. Attempted, and it is not: `FunctionDecl` carries `param_expands` and `return_expand`, and it is **shared**, because `convert!(T).input(fun!(f))` puts `fun!` in `prebindgen-c`'s path too.

  The registry never interprets those two fields — it stores them and hands them back — so what blocks the move is the chaining API rather than any pipeline dependency: `fun!(f).expand_param(…)` needs `FunctionDecl` to hold the payload, and `FunctionDecl` has to stay where `fun!` is. The move therefore needs `fun!` to produce an adapter-specific decl, or the shared `FunctionDecl` to be generic over what an adapter hangs on it. Both are real designs; neither is a rename.

- [x] **7 — the surface nothing uses.** ([#483](https://github.com/milyin/prebindgen/pull/483)) `Cx::require` / `Compiled::required` / `RequirementId` removed — both adapters emit their runtime helpers through `prerequisites`, so the mechanism duplicated a working route. `Construct::Identity` and `Compile::identity` removed together: nothing built the shape, yet the hook was required, so both adapters carried a refusal for a shape they could never be handed. 102 lines out, byte-identical.
- [x] **8 — the zenoh chain, end to end.** ([comment](https://github.com/milyin/prebindgen/pull/472#issuecomment-5372817161)) `zenoh-flat` → `zenoh-flat-jni` → `zenoh-java`, built and run: `./gradlew :zenoh-java:jvmTest` is **BUILD SUCCESSFUL** with **112 tests, 0 failures**. `zenoh-flat-jni` regenerates byte-identically against this branch and `./gradlew jvmJar` builds it — 425 compiled classes plus the dylib.

  **One thing does not work, and it is `main`'s.** Compiling `zenoh-java` against a **regenerated** `zenoh-flat-jni` fails with 28 Kotlin errors, all "inferred type is `T?` but `T` was expected". Measured against three generators: the committed files compile, prebindgen `main` gives 28 errors, this branch gives the same 28. Fallible getters return a nullable type on `main` and `zenoh-java`'s committed source predates that — a `zenoh-java` change, owed to whichever PR made returns nullable there.

## How each stage is checked

Stages 1, 2 and 5 must leave `examples/regen-check.sh` **byte-identical** — they move where an answer comes from, not what it is. That property is what made #465's 5,000 lines reviewable and it is deliberately kept for the mechanical stages.

Stages 3 and 4 **change generated Rust and Kotlin**, and there the regen diff is the review surface. Each will say what moved and why, and the covertest JVM harness plus the zenoh chain are what say it still works.

## What #465 established that this branch assumes

- **One crossing read two ways is two rows**, with the site picking. C needed it three times; JniGen will need it wherever a value crosses differently as a class field than as a parameter.
- **A holding form belongs to the container**, not the value: C's mirror wraps a nested enum in `MaybeUninit` and the value's own conversion is untouched. The JVM equivalent is a slot's boxing.
- **`Role::Part` carries the alternative**, because every arm numbers its parts from zero. A sealed class is exactly that shape.
- **What a role tolerates is the target's answer**, through `Compile::tolerates` — C hands out non-owning pointers, the JVM keeps what it is given.

### DIFF
--- a/prebindgen-c/src/compile.rs
+++ b/prebindgen-c/src/compile.rs
@@ -306,10 +306,6 @@ impl<R: Conversions> Compile for CCompile<'_, R> {
         self.wrap(at, "no C representation for this run", conv)
     }
 
-    fn identity(&mut self, _cx: &mut Cx<'_>, at: At<'_>, _inner: &CFrag) -> Frag<Self> {
-        Err(refuse(at, "Cbindgen declares no identity rows"))
-    }
-
     fn construct(
         &mut self,
         _cx: &mut Cx<'_>,
--- a/prebindgen-c/src/compile.rs
+++ b/prebindgen-c/src/compile.rs
@@ -66,7 +66,7 @@ impl Carrier for CFrag {
 
 impl CFrag {
     /// One of the adapter's existing converter builders, as a fragment.
-    fn from_converter(at: At<'_>, conv: ConverterImpl<()>) -> Self {
+    fn from_converter(at: At<'_>, conv: ConverterImpl) -> Self {
         let validity = validity_of(&conv, at.crossing.assembly());
         Self {
             destination: conv.destination,
@@ -83,7 +83,7 @@ impl CFrag {
     }
 
     /// What the registry that still emits converters expects back.
-    pub(crate) fn into_converter(self) -> ConverterImpl<()> {
+    pub(crate) fn into_converter(self) -> ConverterImpl {
         ConverterImpl {
             destination: self.destination,
             function: self.function,
@@ -101,7 +101,7 @@ impl CFrag {
 /// Reading the spelling is wrong in both directions: a conversion may clone or
 /// allocate out of a borrow, and — the case C actually has — a conversion may
 /// hand C a pointer *into* a Rust value from a crossing that looks owned.
-fn validity_of(conv: &ConverterImpl<()>, assembly: Assembly) -> Validity {
+fn validity_of(conv: &ConverterImpl, assembly: Assembly) -> Validity {
     match assembly {
         // Rust to C. A `*const T` is the zero-copy borrow: `out_borrow_or_result`
         // casts the Rust value's own address, and `repr_c_struct`'s reinterpret
@@ -194,14 +194,14 @@ fn refuse(at: At<'_>, why: &str) -> String {
     format!("Cbindgen: {} ({why})", at.crossing.key())
 }
 
-impl<R: Conversions<()>> CCompile<'_, R> {
-    fn wrap(&self, at: At<'_>, why: &str, conv: Option<ConverterImpl<()>>) -> Frag<Self> {
+impl<R: Conversions> CCompile<'_, R> {
+    fn wrap(&self, at: At<'_>, why: &str, conv: Option<ConverterImpl>) -> Frag<Self> {
         conv.map(|c| CFrag::from_converter(at, c))
             .ok_or_else(|| refuse(at, why))
     }
 }
 
-impl<R: Conversions<()>> Compile for CCompile<'_, R> {
+impl<R: Conversions> Compile for CCompile<'_, R> {
     type Fragment = CFrag;
     /// C keeps its own per-site emission for now: the exported signature, the
     /// call and the cleanup are built in `emit.rs` from the resolved registry.
--- a/prebindgen-c/src/compile.rs
+++ b/prebindgen-c/src/compile.rs
@@ -397,7 +397,7 @@ impl<R: Conversions<()>> Compile for CCompile<'_, R> {
                     // to the value's own conversion.
                     match (
                         at.crossing.assembly(),
-                        self.gen.payload_field_wire(&part.ty, self.registry),
+                        self.gen.payload_field_wire(&part.ty),
                     ) {
                         (Assembly::Deconstruct, Ok(w)) if held_uninit(&w, &frag.destination) => {
                             quote!(::core::mem::MaybeUninit::new(#call))
--- a/prebindgen-c/src/convert.rs
+++ b/prebindgen-c/src/convert.rs
@@ -3,7 +3,7 @@ use prebindgen_registry::Conversions;
 use super::*;
 
 impl CbindgenBuilder {
-    pub(crate) fn prereq_domain_constants(&self, registry: &Registry<()>) -> Vec<syn::Item> {
+    pub(crate) fn prereq_domain_constants(&self, registry: &Registry) -> Vec<syn::Item> {
         let mut items = Vec::new();
         for decl in &self.convert_decls {
             let Some(domain) = decl.domain() else {
@@ -55,9 +55,9 @@ impl CbindgenBuilder {
     pub(crate) fn in_custom(
         &self,
         ty: &TypeRef,
-        registry: &impl Conversions<()>,
+        registry: &impl Conversions,
         emit: &prebindgen_registry::Emit,
-    ) -> Option<ConverterImpl<()>> {
+    ) -> Option<ConverterImpl> {
         let key = ty.key();
         let decl = self.convert_decls.iter().find(|d| *d.key() == key)?;
         let spec = decl.input_spec().as_ref()?;
@@ -123,9 +123,9 @@ impl CbindgenBuilder {
     pub(crate) fn out_custom(
         &self,
         ty: &TypeRef,
-        registry: &impl Conversions<()>,
+        registry: &impl Conversions,
         emit: &prebindgen_registry::Emit,
-    ) -> Option<ConverterImpl<()>> {
+    ) -> Option<ConverterImpl> {
         let key = ty.key();
         let decl = self.convert_decls.iter().find(|d| *d.key() == key)?;
         let spec = decl.output_spec().as_ref()?;
@@ -193,7 +193,7 @@ impl CbindgenBuilder {
         &self,
         decl: &ConvertDecl,
         spec: &ConvertSpec,
-        registry: &impl Conversions<()>,
+        registry: &impl Conversions,
         emit: &prebindgen_registry::Emit,
     ) -> (syn::Type, syn::Expr, bool) {
         let target = self.src_ty_of(&decl.rust_type().key());
@@ -240,7 +240,7 @@ impl CbindgenBuilder {
         &self,
         decl: &ConvertDecl,
         spec: &ConvertSpec,
-        registry: &impl Conversions<()>,
+        registry: &impl Conversions,
         emit: &prebindgen_registry::Emit,
     ) -> (syn::Type, syn::Expr, bool) {
         let target = self.src_ty_of(&decl.rust_type().key());
@@ -282,7 +282,7 @@ impl CbindgenBuilder {
     fn c_domain_niches(
         &self,
         decl: &ConvertDecl,
-        registry: &impl Conversions<()>,
+        registry: &impl Conversions,
         direction: Direction,
     ) -> Niches {
         let Some(domain) = decl.domain() else {
@@ -329,7 +329,7 @@ impl CbindgenBuilder {
         )
     }
 
-    fn conversion_fn_path(&self, registry: &impl Conversions<()>, ident: &syn::Ident) -> syn::Path {
+    fn conversion_fn_path(&self, registry: &impl Conversions, ident: &syn::Ident) -> syn::Path {
         let Some(mut module) = registry.origin_module(ident) else {
             return self.src_fn(ident);
         };
--- a/prebindgen-c/src/emit.rs
+++ b/prebindgen-c/src/emit.rs
@@ -493,8 +493,7 @@ impl CbindgenBuilder {
         // the `ReturnType::Default` arm this used to write.
 
         let has_fallible_input = f.params.iter().any(|p| {
-            registry
-                .input_entry(&p.ty)
+            self.in_frag(&p.ty)
                 .map(|e| returns_result(&e.function.sig.output))
                 .unwrap_or(false)
         });
@@ -508,7 +507,7 @@ impl CbindgenBuilder {
             None => (&f.ret, None),
         };
         let err_ty: Option<syn::Type> = err_reading.map(|t| spelled(t, emit));
-        let has_fallible_output = Self::output_is_fallible(value_ty, registry);
+        let has_fallible_output = self.output_is_fallible(value_ty);
 
         // Error wiring: the error type must be declared via `.error()`.
         let err_bits = err_ty.as_ref().map(|err_ty| {
@@ -716,14 +715,14 @@ impl CbindgenBuilder {
         // own; decomposing it anyway would describe a different ABI from the one
         // the converter table hands out, and the two must agree (#428 review).
         // The base case below already does this for a type with no shape arm.
-        if !r_has_own_wire(ty, registry) {
+        if !self.has_own_wire(ty) {
             // `Vec<T>` → `T_wire* + size_t`. The element must lower to a single C
             // value (one converter); a composite element is unsupported.
             // `TypeKind::Vec` and not `sequence_elem`: that reading peels the
             // erased wrappers first, so a `Cow<'_, [u8]>` would answer here and
             // take the `Vec` lowering instead of its own arm below.
             if let TypeKind::Vec(elem) = ty.kind() {
-                let entry = registry.output_entry(elem).unwrap_or_else(|| {
+                let entry = self.out_frag(elem).unwrap_or_else(|| {
                     panic!(
                         "Cbindgen: `Vec` element `{}` has no output converter",
                         elem.key()
@@ -767,7 +766,7 @@ impl CbindgenBuilder {
             // lowering, so the wrapper returned nothing and called the marker with
             // an argument it does not take (#413).
             if let Some(elem) = r_cow_slice_elem(ty).or_else(|| r_scalar_slice_elem(ty)) {
-                let entry = registry.output_entry(elem).unwrap_or_else(|| {
+                let entry = self.out_frag(elem).unwrap_or_else(|| {
                     panic!(
                         "Cbindgen: `Cow` slice element `{}` has no output converter",
                         elem.key()
@@ -813,7 +812,7 @@ impl CbindgenBuilder {
         // Base value: one wire component from its rank-0/1 converter. Custom
         // conversions may declare scalar niches; otherwise a pointer wire
         // (String, opaque handle, `&'static`) carries a free NULL niche.
-        let entry = registry.output_entry(ty).unwrap_or_else(|| {
+        let entry = self.out_frag(ty).unwrap_or_else(|| {
             panic!(
                 "Cbindgen::on_function: type `{}` has no output converter",
                 ty.key()
@@ -848,9 +847,9 @@ impl CbindgenBuilder {
         // The peer of `lower_shape`'s guard: a node with a wire of its own is
         // encoded by its own converter, whatever its shape. The two walk the
         // same value and must stop at the same places (#428 review).
-        if !r_has_own_wire(ty, registry) {
+        if !self.has_own_wire(ty) {
             if let TypeKind::Vec(elem) = ty.kind() {
-                let entry = registry.output_entry(elem).expect("Vec element converter");
+                let entry = self.out_frag(elem).expect("Vec element converter");
                 let elem_conv = entry.function.sig.ident.clone();
                 let elem_map = map_arg(&elem_conv, entry.function.sig.unsafety.is_some());
                 let elem_wire = entry.destination.clone();
@@ -878,9 +877,7 @@ impl CbindgenBuilder {
                 }
             }
             if let Some(elem) = r_cow_slice_elem(ty).or_else(|| r_scalar_slice_elem(ty)) {
-                let entry = registry
-                    .output_entry(elem)
-                    .expect("slice element converter");
+                let entry = self.out_frag(elem).expect("slice element converter");
                 let elem_conv = entry.function.sig.ident.clone();
                 let elem_map = map_arg(&elem_conv, entry.function.sig.unsafety.is_some());
                 let elem_wire = entry.destination.clone();
@@ -935,7 +932,7 @@ impl CbindgenBuilder {
             }
         }
         // Base value: run its output converter into the single target.
-        let entry = registry.output_entry(ty).expect("base value converter");
+        let entry = self.out_frag(ty).expect("base value converter");
         let conv = entry.function.sig.ident.clone();
         let t0 = &targets[0];
         if returns_result(&entry.function.sig.output) {
@@ -946,7 +943,7 @@ impl CbindgenBuilder {
         }
     }
 
-    fn output_is_fallible(ty: &TypeRef, registry: &impl Conversions<()>) -> bool {
+    fn output_is_fallible(&self, ty: &TypeRef) -> bool {
         // The third walk over the same value, and it stops where the other two
         // do: a node with a wire of its own is encoded by its own converter, so
         // whether the encode can fail is THAT converter's answer. Peeling past
@@ -954,7 +951,7 @@ impl CbindgenBuilder {
         // binding needs `.panic()`, so the two disagreeing is a wrapper that
         // aborts where nothing opted in, or an opt-in demanded for a conversion
         // that cannot fail (#428 review).
-        if !r_has_own_wire(ty, registry) {
+        if !self.has_own_wire(ty) {
             let vec_elem = match ty.kind() {
                 TypeKind::Vec(e) => Some(&**e),
                 _ => None,
@@ -965,11 +962,10 @@ impl CbindgenBuilder {
                 .or_else(|| r_cow_slice_elem(ty))
                 .or_else(|| r_scalar_slice_elem(ty))
             {
-                return Self::output_is_fallible(inner, registry);
+                return self.output_is_fallible(inner);
             }
         }
-        registry
-            .output_entry(ty)
+        self.out_frag(ty)
             .is_some_and(|entry| returns_result(&entry.function.sig.output))
     }
 
--- a/prebindgen-c/src/emit.rs
+++ b/prebindgen-c/src/emit.rs
@@ -6,7 +6,7 @@ impl CbindgenBuilder {
     /// Whether the generated layer hands `char*` data memory to C — a `String`
     /// return value, or a declared data struct that is produced as output and has
     /// a `String` field. When true, a `free_memory_function` must be declared.
-    pub(super) fn needs_free(&self, registry: &Registry<()>) -> bool {
+    pub(super) fn needs_free(&self, registry: &Registry) -> bool {
         let string_ty: syn::Type = syn::parse_quote!(String);
         // A `String` return hands out a `char*` — unless `String` is declared
         // `opaque_ptr` (then it crosses as `string_t *`, freed by `string_drop`).
@@ -65,7 +65,7 @@ impl CbindgenBuilder {
     /// path segment, and fetched the `syn::ItemEnum` to reach the same list.
     pub(super) fn enum_alternatives<'r>(
         &self,
-        registry: &'r Registry<()>,
+        registry: &'r Registry,
         key: &TypeKey,
     ) -> Option<&'r [prebindgen_registry::flat::Alternative]> {
         match registry.flat().declared_type(&key.ident()?)? {
@@ -284,7 +284,7 @@ impl CbindgenBuilder {
         &self,
         fty: &TypeRef,
         wire: &syn::Type,
-        registry: &Registry<()>,
+        registry: &Registry,
     ) -> bool {
         if matches!(wire, syn::Type::Ptr(_)) {
             return true;
@@ -297,7 +297,7 @@ impl CbindgenBuilder {
     pub(super) fn owning_data_struct_fields<'r>(
         &self,
         fty: &TypeRef,
-        registry: &'r Registry<()>,
+        registry: &'r Registry,
     ) -> Vec<(syn::Ident, &'r TypeRef)> {
         if !self.data.contains_key(&fty.key()) {
             return Vec::new();
@@ -313,7 +313,7 @@ impl CbindgenBuilder {
     /// is a pointer (`String` → `char *`), or it is a declared
     /// [`CbindgenBuilder::tagged_union`] with an owning arm — which crosses by value,
     /// so the pointer it owns is one level further down.
-    fn data_field_owns(&self, fty: &TypeRef, registry: &Registry<()>) -> bool {
+    fn data_field_owns(&self, fty: &TypeRef, registry: &Registry) -> bool {
         if matches!(self.data_field_wire(fty), Some(syn::Type::Ptr(_))) {
             return true;
         }
@@ -328,7 +328,7 @@ impl CbindgenBuilder {
     /// containing struct has to call it, so a union nested inside a payload
     /// cannot be freed through a symbol that was never emitted. `false` for
     /// anything that is not a declared tagged union.
-    pub(super) fn tagged_union_has_drop(&self, fty: &TypeRef, registry: &Registry<()>) -> bool {
+    pub(super) fn tagged_union_has_drop(&self, fty: &TypeRef, registry: &Registry) -> bool {
         if !self.tagged_unions.contains_key(&fty.key()) || self.out_frag(fty).is_none() {
             return false;
         }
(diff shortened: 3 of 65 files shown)


milyin:
## Stage 8, rehearsed: the zenoh chain regenerates byte-identically

Built `zenoh-flat-jni` against this branch rather than prebindgen's `main`. **The generated Rust and Kotlin are byte-identical** — `git status` after a full regeneration shows only `Cargo.lock`, moved by the patch itself.

That is the strongest evidence so far that the ten merged stages are the refactor they claim to be. `examples/regen-check.sh` covers prebindgen's own examples; this covers a real binding of ~40 Kotlin files that nothing in this repo generates.

Not yet stage 8 as written — that asks for the chain **built and run**, and this is regeneration plus `cargo check`. The Gradle build and the zenoh-java run are still to come.

### One thing worth recording for whoever does the full run

The chain needs **two** patches, not one:

```toml
[patch."https://github.com/milyin/prebindgen.git"]
prebindgen = { path = "prebindgen/prebindgen" }
# … and the rest of the crates

[patch."https://github.com/eclipse-zenoh/zenoh-flat.git"]
zenoh-flat = { path = "zenoh-flat" }

[patch.crates-io]
prebindgen = { path = "prebindgen/prebindgen" }
prebindgen-proc-macro = { path = "prebindgen/prebindgen-proc-macro" }
```

Patching only prebindgen fails, and prebindgen's own error says why:

> The directory … was written by prebindgen <= 0.5.0, which this prebindgen cannot read: the capture layout changed and carries a format number now. **Every prebindgen that touches one capture directory has to agree on the format — they are separate dependencies, so Cargo does not check it.**

`zenoh-flat-jni` takes `zenoh-flat` from git, so the **source** crate was capturing with the published proc-macro while the local generator read the capture. The source crate has to be patched to the local checkout as well, and its proc-macro comes from crates.io rather than the git source. Worth putting in `DEVELOPMENT.md` if stage 8 lands this way.

milyin:
## Stage 8, second step: `zenoh-flat-jni` builds and jars against this branch

Beyond the byte-identical regeneration reported above, the JVM artifact now builds from it:

- `./gradlew jvmJar` — **BUILD SUCCESSFUL**, with the Rust release library compiled from this branch's generated bindings.
- The jar carries the freshly built `libzenoh_flat_jni.so` and **425 compiled classes** from `kotlin/generated`.

So the generated Kotlin does not merely regenerate identically — it still **compiles**, which the regen diff alone cannot say.

### Environment note for CI or anyone reproducing

`./gradlew build` fails before reaching the JVM target:

> SDK location not found. Define a valid SDK location with an ANDROID_HOME environment variable…

That is the `com.android.library` plugin, unrelated to prebindgen — the multiplatform project publishes a `-jvm` and an `-android` flavour, and the Android one needs an SDK this machine has none of. `jvmJar` is the target that exercises the generated code without it.

milyin:
## Stage 3 verified on the zenoh binding, and stage 8's baseline restated

Regenerated `zenoh-flat-jni` three times against the same checkout — prebindgen `main`, this branch's head before stage 3, and this branch with [#486](https://github.com/milyin/prebindgen/pull/486) and [#487](https://github.com/milyin/prebindgen/pull/487) — and diffed the three generated trees against each other.

**All three are identical.** So stage 3 changes nothing in a binding of ~40 generated Kotlin files that this repo does not produce, which is the check `examples/regen-check.sh` cannot make. Worth stating plainly because stage 3 was written expecting output to change.

### What that run did turn up

The **committed** generated files in the `zenoh-flat-jni` checkout are stale against prebindgen `main`, not against this branch. Regenerating changes 15 files (+749/−362), and the changes are recognisably from work already merged to `main`:

- `@file:OptIn(io.zenoh.jni.UnsafeNativeApi::class)` on every generated file,
- typed handles taking a `private constructor` and minting through `fromRawPtr` (#404, which sealed the constructor so no JVM caller can forge a pointer),
- fallible getters returning a nullable type, with the error handler typed to match.

So the earlier note on this PR — that regeneration is byte-identical — held for the checkout as it stood then and does not now. The diff belongs to `main`, and stage 8 has to review it there rather than attribute it to this branch. Nothing here regresses: the branch and `main` agree wire for wire.

### Reproducing

`zenoh-flat-jni` resolves prebindgen from crates.io, so the chain needs both patches from the [earlier comment](https://github.com/milyin/prebindgen/pull/472#issuecomment-5367583173) — the six prebindgen crates through `[patch.crates-io]`, and `zenoh-flat` through its git URL, so the crate that captures and the crate that reads the capture are the same build.

milyin:
## Stage 8, second step re-run against the branch as it now stands

Built `zenoh-flat-jni` from this branch — stages 3, 4's composition and 5a all in — rather than from the head the [earlier note](https://github.com/milyin/prebindgen/pull/472#issuecomment-5368078049) used.

- Regeneration is **identical to what prebindgen `main` produces**, so nothing this branch adds reaches the generated files.
- `./gradlew jvmJar` — **BUILD SUCCESSFUL**. The jar carries the freshly built `libzenoh_flat_jni.so` and **425 compiled classes** from `kotlin/generated`, the same count the earlier run reported.

So the generated Kotlin does not merely regenerate identically — it still compiles, which the regen diff alone cannot say. What stage 8 still asks for beyond this is the zenoh-java run.

`./gradlew build` still fails before reaching the JVM target on a machine with no Android SDK (`SDK location not found`), which is the `com.android.library` plugin and unrelated to prebindgen. `jvmJar` is the target that exercises the generated code without one.

milyin:
## Stage 8 re-verified after ten merged stages

Rebuilt `zenoh-flat-jni` against this branch with everything through [#497](https://github.com/milyin/prebindgen/pull/497) merged — stages 3, 4 (composition and most of the switch), 5a and 5b.

- Regeneration is **identical to what prebindgen `main` produces**: the ~40 generated Kotlin files and the generated Rust, diffed tree against tree.
- `./gradlew jvmJar` — **BUILD SUCCESSFUL**, 425 compiled classes from `kotlin/generated` plus the freshly built `libzenoh_flat_jni.so`.

So ten stages of moving where an answer comes from have changed no answer, on a binding this repo does not produce and whose ~40 files `examples/regen-check.sh` never sees.

milyin (commented):
Found one correctness issue:

- **[P2] Make `Compiler::row_of` reject an absent `RecipeId`** (`prebindgen-registry/src/recipe/compile.rs:542`). `row_of` promises to compile a particular named row, but it delegates to `row`, where a failed `Recipes::get` silently selects the derived/default row. This happens today for empty `data_class` / sealed declarations: `rows::recipes` omits `parts` when there are no fields/arms, while `declared_decompositions` still calls `row_of(..., parts())`. The result is that `Compiled::row_fragment(..., parts)` returns the whole atomic fragment even though no `parts` row exists. That bypasses the row-table validation and leaves a semantically mislabeled fragment for emitters. Please separate explicit named-row lookup from the default fallback (or guard `row_of`) and add a missing-row regression test.

Verification completed: `cargo test --workspace --all-targets`, clippy, fmt, rustdoc with warnings denied, `examples/regen-check.sh`, and the JVM covertest (57 sections) all pass.

— Codex (GPT-5)

milyin (commented):
Rereviewed `4f31ebb0`. The prior P2 is resolved: `Compiler::row_of` now rejects absent named rows with `RecipeError::NoSuchRow`, JniGen guards conditional `parts` rows before compiling them, and empty aggregates no longer file mislabeled fragments. The added registry and JniGen regressions cover both sides of the fix.

No new findings.

Verified locally:
- `cargo test -p prebindgen-registry -p prebindgen-jni` (459 unit tests plus doc tests)
- `examples/regen-check.sh` (byte-identical generated output)

All current GitHub checks are green.

— Codex (GPT-5)

PR #475, opened 2026-08-21
prebindgen-jni's emitters read their own fragments

milyin:
Stage 1b of #472. Peer of [#474](https://github.com/milyin/prebindgen/pull/474), which did the same for `prebindgen-c`.

## What moves

`prebindgen-jni` asked the converter table — the registry's index of one
resolved conversion per crossing — what a type crosses as, at **65** sites. A
table entry carries one `destination`, the single wire type that conversion
produces, so it can answer only while a crossing occupies one wire value.

**57 of the 65** now read `Declarations::compiled`, the fragments this binding
built, through `Declarations::{in_frag, out_frag}`.

The other 8 are free functions with no `Declarations` in scope. Four are the
`Option<_>` conversion builders in `emit/convert.rs`, which have nine test
callers that construct none; the rest are in `sum_out.rs`, `delivery.rs` and
`fn_plan.rs`. They go with the table in stage 2 rather than growing a parameter
here.

## Why this is not the same change as #474

`prebindgen-c` splits into two phases: it compiles, then its per-item emitters
run and read what compilation produced. `prebindgen-jni` does not split. Its
emitters **are** its conversion builders — `struct_input_body`,
`callback_input`, `struct_plan`, `iface`, `render`, `delivery` are all called
from inside `convert_with` — so they ask while the store is still filling.

That is fine, and for the reason the table worked: a conversion for one type is
built out of the conversions for its inners, and the resolver compiles the
inners first. A fragment is there exactly when a table entry would have been. So
`Declarations::compiled` is an `Rc<RefCell<…>>`, read during compilation rather
than after it.

The trap is in the driver. Handing the compiler the store with `mem::take` and
putting the result back afterwards leaves the store **empty for the duration of
every call**, which is precisely when the adapter reads it. `Compiled` is a map
of `Rc`s, so the driver clones it instead — 46 of the 264 tests turn on this
alone.

## The one crossing the compiler never sees

`JniGen::compile_crossing` answers a callback with `dispatch_fn_input` rather
than through a row, so no row builds a callback crossing and nothing would file
a fragment for it. That is pre-existing and already noted in the driver.

`Compiled::record` is the escape hatch: the driver files the conversion it built
by hand as that crossing's fragment, so every emitter keeps **one** lookup
rather than a fall-back for the one crossing kind the compiler skips.
`JFrag::by_hand` supplies the `Yield` no row was there to derive — owned and
self-sufficient, which a callback delivered as a JVM object is, and which
nothing reads because nothing composes a callback as an inner. Both go with the
derived callback row.

## Registry support

`ConverterImpl` gains the four reads `TypeEntry` already offered
(`converter_ident`, `wire_type`, `input_stage_order`, `output_stage_order`). The
two types carry the same six fields — an entry is what the resolver stored, a
conversion is what the adapter handed it — so every read an emitter makes works
on either.

`TypeEntry::as_converter` bridges the compile-time callers that still hold an
entry. Both it and those reads go with the table in stage 2.

Fourteen parameters became dead and are removed, including one `carry_layers`
argument clippy flagged as used only in its own recursion.

## Checks

- `cargo test --all` — green (264 in `prebindgen-jni`)

### DIFF
--- a/prebindgen-jni/src/jni/builder.rs
+++ b/prebindgen-jni/src/jni/builder.rs
@@ -148,6 +148,7 @@ impl Default for Declarations {
     fn default() -> Self {
         Self {
             compiled_fns: Vec::new(),
+            compiled: Default::default(),
             package: String::new(),
             fun_name_mangle: None,
             ptr_class_name_mangle: None,
@@ -1939,9 +1940,7 @@ impl Declarations {
         let inner = if is_self {
             None
         } else {
-            registry
-                .reading_of(&ty)
-                .and_then(|tr| registry.input_entry(&tr))
+            registry.reading_of(&ty).and_then(|tr| self.in_frag(&tr))
         };
         match inner {
             None if is_self || is_wire_type(&ty) => {
@@ -2085,9 +2084,7 @@ impl Declarations {
         let inner = if is_self {
             None
         } else {
-            registry
-                .reading_of(&ty)
-                .and_then(|tr| registry.output_entry(&tr))
+            registry.reading_of(&ty).and_then(|tr| self.out_frag(&tr))
         };
         match inner {
             None if is_self || is_wire_type(&ty) => {
@@ -2095,7 +2092,7 @@ impl Declarations {
                 let (kotlin_name, value_rust_type) = if let Some(a0) = arg0 {
                     registry
                         .reading_of(a0)
-                        .and_then(|tr| registry.output_entry(&tr))
+                        .and_then(|tr| self.out_frag(&tr))
                         .map(|e| {
                             (
                                 e.metadata.kotlin_name.clone(),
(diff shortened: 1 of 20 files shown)


milyin:
Both addressed in 9c44978.

**`as_converter` dropping `subs`** — correct, and it corrects the reason I gave for the emptying. `ConverterImpl::subs` is `Vec<TypeKey>`, the same as `TypeEntry::subs`; the PR body's claim that one keys its inners and the other spells them is wrong. There was nothing to convert, so the bridge now clones the field. Nothing reads `subs` off a bridged value today — every use in `prebindgen-jni` is a write — but a public bridge that silently drops reachability edges is a trap for the next caller.

**The doc link** — fixed to `ConverterImpl::function`, which is what that helper takes now.

Worth recording for stage 6 of #472: with `subs` compared field by field, `TypeEntry` and `ConverterImpl` are structurally identical. That is a stronger case for deleting one of them than the umbrella currently makes.

PR #476, opened 2026-08-21
Take both adapters off the converter table

milyin:
First half of stage 2 in #472.

## Where this starts

Stages [1](https://github.com/milyin/prebindgen/pull/474) and
[1b](https://github.com/milyin/prebindgen/pull/475) moved the bulk of both
adapters' lookups from the converter table — the registry's index of one
resolved conversion per crossing — to the fragments each adapter compiled.
They left **18**: 10 in `prebindgen-c` and 8 in `prebindgen-jni`.

Both counts are now **zero**. That is what stage 2b needs: while an adapter
still reads the table, `convert_with` has to keep handing back a whole
`ConverterImpl` for the table to hold.

## `prebindgen-c`: the same in-flight store `prebindgen-jni` has

Four `prebindgen-c` helpers run during compilation as well as after it —
`lower_shape`, `encode_value`, `output_is_fallible`, and the
`has_own_wire` / `shape_is_lowerable` / `is_lowered_composite` cluster.
`dispatch_fn_input` builds a callback's closure struct out of them while
compilation is still in progress, which is why #474 left them behind.

`CbindgenBuilder::compiled` is now an `Rc<RefCell<…>>` fed after every
crossing, exactly as `Declarations::compiled` is. A conversion for one type
is built out of the conversions for its inners, which the resolver compiles
first, so a fragment is there when the helper asks. The same `mem::take`
trap applies and the driver clones instead.

The three lowering predicates were free functions taking a registry. They
are methods on `CbindgenBuilder` now — "does this binding give this type a
wire" was always its question, not the registry's.

## `prebindgen-jni`: seven parameters and one test harness

Seven of the eight were free functions with no `Declarations` in scope —
`leaf_is_prim`, `leaf_ty_is_prim`, `leaf_slot`, `encode_group_leaf`,
`nullable_kind_for`, `nullable_kind_for_output`, `jvm_parameter_slots`.
Each takes one now; every caller already had it.

The eighth pair, `option_input` / `option_output`, is why #475 left this
group alone: nine tests in `jni/tests/niches.rs` call them against a bare
`Registry` built by `install_input` / `install_output`, with no
`Declarations` anywhere. Those installers now file each conversion in
**both** places — the registry the test builds and the `Declarations` as
the fragment a compiled binding would have recorded. A test still installs
one entry per type; the helper under test reads the adapter's own answer.

## Left over

Nine more parameters became dead and are removed.

What still reads a table entry is the registry itself, and only for `subs`,
the dependency keys reachability walks. Narrowing `convert_with` to exactly
that is stage 2b.

## Checks

- `cargo test --all` — green

### DIFF
--- a/prebindgen-c/src/emit.rs
+++ b/prebindgen-c/src/emit.rs
@@ -493,8 +493,7 @@ impl CbindgenBuilder {
         // the `ReturnType::Default` arm this used to write.
 
         let has_fallible_input = f.params.iter().any(|p| {
-            registry
-                .input_entry(&p.ty)
+            self.in_frag(&p.ty)
                 .map(|e| returns_result(&e.function.sig.output))
                 .unwrap_or(false)
         });
@@ -508,7 +507,7 @@ impl CbindgenBuilder {
             None => (&f.ret, None),
         };
         let err_ty: Option<syn::Type> = err_reading.map(|t| spelled(t, emit));
-        let has_fallible_output = Self::output_is_fallible(value_ty, registry);
+        let has_fallible_output = self.output_is_fallible(value_ty);
 
         // Error wiring: the error type must be declared via `.error()`.
         let err_bits = err_ty.as_ref().map(|err_ty| {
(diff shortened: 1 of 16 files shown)


milyin:
Addressed in 36459b4.

`Compiled` already stored each fragment behind an `Rc`, so the lookups now hand that back instead of copying: `fragment` and `row_fragment` return `Option<Rc<F>>`. `prebindgen-c` passes it straight through, which is where the recursion you named lives — `shape_is_lowerable` asks once per layer as it walks a type down.

A borrow guard would not work here: both adapters read the store **while compilation is still writing to it**, so a `Ref` cannot be held across the next write. Sharing the `Rc` costs a refcount and has no such lifetime.

`prebindgen-jni`'s lookups answer with a conversion rather than a fragment, so a straight `Rc` would have moved 57 call sites onto `.conv.field`. `Conv` holds the `Rc` and derefs to the conversion instead; the call sites read what they always did. Four needed an explicit `&*`, where `&expr?` gives the coercion nothing to fire on.

PR #477, opened 2026-08-21
Delete the paths that read conversions off the table

milyin:
Opens stage 2b of #472. Pure deletion: 160 lines out, 53 in, generated output byte-identical.

## What this removes

[#476](https://github.com/milyin/prebindgen/pull/476) took the last adapter
lookup off the converter table — the registry's index of one resolved
conversion per crossing. Nothing now asks the registry to hand a conversion
back, which strands four things:

- **`Conversions`**, the enum telling `write_rust` where generated
  conversions come from, had two variants and one live one. `write_rust`
  takes the adapter's compiled functions as `&[syn::ItemFn]` directly, and
  the enum is gone. `Conversions::Table` had two users, both registry tests,
  and neither asserted anything about a converter — they check const
  ordering and feature guards.
- **`collect_converter_items`** and **`walk_resolved`** existed to serve
  that variant.
- **`RegistryBuilder::conversions`**, documented as the bulk peer of
  `convert_with` for a generator that walks the crossings itself, had no
  caller at all.
- **`RegistryBuilder::crossings`** was the other half of that pair. It hands
  over the crossing list for a caller driving the loop itself — which, with
  `conversions` gone, has no way to hand a result back. `convert_with` is
  the one way to supply conversions now, and its doc comment says so instead
  of pointing at a second mechanism.

## The test

`dedup_and_sort` covered `collect_converter_items`, asserting that one
function per name reaches the file in name order. That property now lives in
`dedup_by_name`, so the test exercises it there — with the same two
conversions plus a repeat, which is what gives the de-duplication something
to do. It no longer needs to build a registry and stuff two `TypeEntry`s
into it to make the point.

## What is left for the rest of the stage

`convert_with` still takes back a whole `ConverterImpl`. What the registry
does with one is read `subs` — the dependency keys `propagate_required`
walks — and note that the crossing was answered. Narrowing it to that is the
next PR, and it reaches further than the table: strip a cell down to `subs`
and `M`, the metadata type parameter, goes unconstrained across the crate.

## Checks

- `cargo test --all` — green

### DIFF
--- a/prebindgen-c/src/lib.rs
+++ b/prebindgen-c/src/lib.rs
@@ -462,12 +462,12 @@ pub struct CbindgenBuilder {
     >,
     /// Every conversion this binding compiled.
     ///
-    /// Filled once by [`Self::build_with`] and handed to `write_rust` as
-    /// `Conversions::Compiled`. It is what reaches the generated file,
-    /// so a fragment no longer has to be expressible as one `ConverterImpl` to
-    /// be emitted — only to be looked up. The writer sorts and de-duplicates by
-    /// function name, so the order here decides which of two same-named
-    /// functions wins and not where any of them lands.
+    /// Filled once by [`Self::build_with`] and handed to `write_rust` directly. It is what
+    /// reaches the generated file, so a fragment no longer has to be
+    /// expressible as one `ConverterImpl` to be emitted — only to be looked up.
+    /// The writer sorts and de-duplicates by function name, so the order here
+    /// decides which of two same-named functions wins and not where any of them
+    /// lands.
     pub(crate) compiled_fns: Vec<syn::ItemFn>,
 }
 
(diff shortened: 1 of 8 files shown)


milyin:
All four addressed in 6ad414b — and the general point stands: rustdoc's `-D warnings` catches a broken link, not a paragraph that is merely wrong, so fixing the links this deletion broke was not the same as fixing what it made untrue.

- **`write.rs` module docs** — said `write_rust` collects every resolved converter off the table. Rewritten to the slice the adapter hands over, with the reason it matters: conversions arriving from the adapter is what lets one crossing contribute more than one function, or one occupying more than a single wire value.
- **`compiled_fns` in both adapters** — named the deleted variant; they now say `write_rust` takes them directly.
- **`prebindgen.rs`** — two further places, which the review's "other module docs" pointed at. Both described a generator answering the demand `RegistryBuilder::crossings` hands it. That call is gone and there is no replacement for it: a conversion is built inside `convert_with`.

Also narrowed the `order` field comment, which listed `crossings` among the callers needing the scan to have run.

PR #478, opened 2026-08-21
Take back reachability from convert_with, not the conversion

milyin:
Completes stage 2 of #472 — **the gate**. Stacked on [#477](https://github.com/milyin/prebindgen/pull/477); merge that first.

849 lines out, 591 in, generated output byte-identical.

## What the registry was holding, and why it stopped

The registry stored a whole `ConverterImpl` per crossing — a `TypeEntry` —
so an adapter could look one up. After
[#476](https://github.com/milyin/prebindgen/pull/476) neither adapter does.

What the registry itself reads of a stored entry is `subs`: the crossings
this one delegates to, which `propagate_required` walks to decide what a
binding actually has to be able to make. Plus whether the crossing was
answered at all.

So `convert_with` takes back an `Answer` carrying exactly that. The
conversion stays in the adapter, which emits from it and looks it up.

`TypeEntry` is gone. It held six fields, all but `subs` for readers that no
longer exist, and — as Copilot's review of #475 established — it was
structurally identical to the `ConverterImpl` it was built from.

## Why the diff is this big

Strip a cell down to `subs` and the compiler says where the rest of it
reached: **`M` is unconstrained across `prebindgen-registry`**. The metadata
parameter existed to carry a `KotlinMeta` or a `()` through the converter
table, and an adapter reads its metadata off its own fragment now.

So `Registry<M>`, `RegistryBuilder<M>`, `Conversions<M>`, `Building<'_, M>`
and the `Prebindgen::Metadata` that fed them all lose the parameter.
`ConverterImpl<M>` keeps it — it is what an adapter's fragment carries, and
where `KotlinMeta` still lives.

Most of the diff is deleting a type argument. The load-bearing part is
`Answer`, `convert_with`'s signature, and the deletions below.

## Also gone

- **`Conversions::conversion`**, with `input_entry` and `output_entry` on
  top of it, and the inherent pair on `Registry` that shadowed them.
- **`Building::built`** — `conversion` was its only reader. `Building` no
  longer lends out conversions built so far; one `ConverterImpl` per
  crossing was both less than an adapter already holds and less than a
  multi-wire crossing can be described by.

## Tests

Two changed rather than moved:

- `type_entry_helpers_expose_converter_chain_contract` asserted
  `converter_ident` / `wire_type` / `input_stage_order` /
  `output_stage_order` / `dependency_keys`. Those reads moved to
  `ConverterImpl` in #475, so the test follows them there.
- The fixtures that built a `TypeEntry` purely to put one in a registry
  were saying "this crossing is answered, and delegates to these". They say
  it directly now — `Answer::over(vec![…])` — which is both shorter and the
  thing those tests were ever about.

`jni/tests/niches.rs`'s `install` helper files a `ConverterImpl` as the
adapter's fragment and an `Answer` in the registry, which is the split this
PR makes everywhere.

## Checks

- `cargo test --all` — green
- `cargo doc` with `-D warnings` — clean, including the prose that described
  the deleted type rather than only the links that broke

### DIFF
--- a/prebindgen-c/src/compile.rs
+++ b/prebindgen-c/src/compile.rs
@@ -66,7 +66,7 @@ impl Carrier for CFrag {
 
 impl CFrag {
     /// One of the adapter's existing converter builders, as a fragment.
-    fn from_converter(at: At<'_>, conv: ConverterImpl<()>) -> Self {
+    fn from_converter(at: At<'_>, conv: ConverterImpl) -> Self {
         let validity = validity_of(&conv, at.crossing.assembly());
         Self {
             destination: conv.destination,
@@ -83,7 +83,7 @@ impl CFrag {
     }
 
     /// What the registry that still emits converters expects back.
-    pub(crate) fn into_converter(self) -> ConverterImpl<()> {
+    pub(crate) fn into_converter(self) -> ConverterImpl {
         ConverterImpl {
             destination: self.destination,
             function: self.function,
@@ -101,7 +101,7 @@ impl CFrag {
 /// Reading the spelling is wrong in both directions: a conversion may clone or
 /// allocate out of a borrow, and — the case C actually has — a conversion may
 /// hand C a pointer *into* a Rust value from a crossing that looks owned.
-fn validity_of(conv: &ConverterImpl<()>, assembly: Assembly) -> Validity {
+fn validity_of(conv: &ConverterImpl, assembly: Assembly) -> Validity {
     match assembly {
         // Rust to C. A `*const T` is the zero-copy borrow: `out_borrow_or_result`
         // casts the Rust value's own address, and `repr_c_struct`'s reinterpret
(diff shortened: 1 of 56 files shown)


milyin:
The exact line you flagged — `RegistryBuilder`'s `order` field comment — was already fixed in the parent PR [#477](https://github.com/milyin/prebindgen/pull/477), which is where `crossings` was deleted; this branch is rebased on it and reads `[`Self::convert_with`] and [`Self::build`] each need it to have run`.

The sweep it prompted did find three more, fixed in 942c8da:

- `resolve.rs`'s module docs — "the generator fills the cells itself (`RegistryBuilder::crossings` → `convert_with`)".
- `order.rs`'s module docs — "`RegistryBuilder::build` takes every conversion back at once". It takes an `Answer` per crossing now, and grades what came back.
- `RegistryBuilder::built`, which called itself "conversions handed over so far" — the thing this PR stops it from holding.

Worth stating because it is the second review in a row to catch this: `cargo doc` with `-D warnings` fails on a **broken link** and says nothing about a paragraph that is merely wrong. Deleting a public item leaves both, and only one of them is caught.

PR #479, opened 2026-08-21
Compose what a JNI data class is made of

milyin:
First half of stage 3 in #472. Byte-identical: nothing takes the new row yet.

## The thing being moved

A JniGen `data_class` parameter arrives as **several** JNI parameters. In the
covertest goldens, `holderTagOr` takes an `Option<Holder>` and its extern is

```kotlin
external fun holderTagOr(
    hPresent: Boolean, hTag: Long, hSummary: Long, fallback: Long, errorSink: Any,
): Long
```

Today `render.rs` builds that by walking `FlatInputPlan::leaves` — a
hand-written descent in `emit/flat_input.rs`, 2,148 lines. This PR states the
same thing as a row and composes it from the parts.

## What composes

`JFrag` gains a **wire list**, for a crossing that occupies more than the one
`conv.destination` can name. `Compile::fields` fills it by concatenating its
parts' own wires, so a field that is itself a `data_class` contributes its
several rather than one. That is the recursion, stated once instead of walked.

## The site has to speak, not just the row

The first attempt composed `Holder { tag: i64, summary: Summary }` into **two**
wires, not three. A part is compiled under its **own default row**, and a
nested `data_class`'s default is `whole` — so the flattening stopped one layer
down.

`Declarations::bindings` binds each data-class field of a `data_class` to the
`parts` row. That is the "one crossing read two ways is two rows, with the site
picking" mechanism `prebindgen-c` already uses for union payloads, and it is the
one thing a site can say that a row cannot.

## Why this is still byte-identical

- `whole` stays the default row every site takes; `parts` is declared but
  chosen by nobody.
- The `parts` fragment carries a **marker** rather than a conversion. The
  function that reads several values and rebuilds the struct is what the
  emitter switch brings; inventing one here would be inventing the half this
  PR is not doing. `prebindgen-c` does the same for a union arm.
- A fragment holding only wires emits nothing, so the marker does not reach
  the generated file.

## How the composition is checked before anything depends on it

`JniGen::compile_crossing` compiles the `parts` row for **every** declared
`data_class`, so all 264 lib tests exercise it rather than one fixture. A
refusal panics naming the type: every part of a `data_class` is a crossing that
already resolved on its own, so nothing there can legitimately be missing, and
returning `None` would report an unresolved crossing and blame the declaration.

One test asserts the shape that makes the recursion visible — `Holder` is a
scalar plus a nested `Summary`, so it crosses as `tag`, `summary.count`,
`summary.total` and not as two values.

`Compiler::row_of` is what compiles a named row; the read side,
`Compiled::row_fragment`, was already public.

## What is left for the second half

Point `render.rs` and the emitters at the fragment, give the row a real
conversion, and delete `emit/flat_input.rs`. That one changes generated output.

## Checks

- `cargo test --all` — green

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -24,6 +24,29 @@ use super::{
 pub(crate) struct JFrag {
     pub(crate) conv: ConverterImpl<KotlinMeta>,
     pub(crate) yields: Yield,
+    /// The wire values this crossing occupies, when it occupies more than the
+    /// one `conv.destination` names.
+    ///
+    /// A JniGen `data_class` parameter arrives as **several** JNI parameters —
+    /// `Option<Holder>` is `(hPresent, hTag, hSummary)` — so a single
+    /// destination cannot say what it costs. `None` is the ordinary case: one
+    /// wire, and `conv.destination` is it.
+    ///
+    /// Composed by [`Compile::fields`] from the parts' own wires, which is what
+    /// makes a nested `data_class` field contribute its own several rather than
+    /// one.
+    pub(crate) wires: Option<Vec<Wire>>,
+}
+
+/// One wire value of a crossing that occupies several.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct Wire {
+    /// The JNI type this value crosses as.
+    pub(crate) ty: syn::Type,
+    /// The path through the value that reached it — `tag`, `summary.count` —
+    /// which is what a JNI parameter name and a Kotlin accessor are both
+    /// mangled from.
+    pub(crate) path: String,
 }
 
 impl Carrier for JFrag {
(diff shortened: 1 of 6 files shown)


milyin:
Right, and it names an option I had not weighed. I chose between returning `None` and panicking; the builder already has a third that is better than both.

`None` is wrong for the reason the comment gave — it reports an unresolved crossing and blames the declaration for what is a bug in the composition. But panicking unwinds a build over a row **no site takes**, and it hands the caller nothing through the `Result` every other refusal arrives on.

Fixed in e85da9e: the reasons are collected during the walk and surfaced together as `ScanError::AdapterInvariant`. Forcing `Compile::fields` to refuse now yields

```
Scan(AdapterInvariant { message: "`Summary` crosses as its fields, but composing them failed: …; `Holder` crosses as its fields, but composing them failed: …" })
```

— both the nested class and its container in one message, where the panic aborted on the first. That is the "beside other diagnostics" half of your point, and it is worth more than I expected: a composition bug in a leaf type shows up together with every container it breaks.

PR #480, opened 2026-08-21
Hold the composed row to what the walk it replaces says

milyin:
Continues stage 3 of #472, after [#479](https://github.com/milyin/prebindgen/pull/479). Byte-identical.

## What this establishes

#479 composed a `data_class` into a wire list but only checked its shape
against a hand-written expectation. The row is meant to state **exactly** what
`build_flat_input_plan` — the 2,148-line walk in `emit/flat_input.rs` — states,
and asserting that is what turns the emitter switch from a rewrite into a move.

So `Wire` now carries the Kotlin type beside the JNI one, and a test puts the
row and the walk side by side for a real binding:

```
[("hTag", "Long"), ("hSummaryCount", "Long"), ("hSummaryTotal", "Double")]
```

That is `Holder { tag: i64, summary: Summary }` — a scalar and a nested
`data_class` — composed and walked, equal. The nested class contributing
`hSummaryCount` and `hSummaryTotal` rather than one value is the recursion the
switch depends on.

## A wire's path is relative, on purpose

The `h` in `hTag` belongs to the **parameter**, not to `Holder`: the same
struct is `hTag` in one signature and `otherTag` in the next. A fragment
answers for a crossing, so it carries `tag` and `summary.count`, and the site
prepends. That keeps naming out of the fragment and leaves it where
`Compile::plan` will want it, without this PR needing plans.

## What is deliberately not here

An `Option<data_class>` adds a presence flag — the `hPresent` in
`holderTagOr(hPresent, hTag, hSummary, …)`. Composing it belongs to the
optional's row, and reaching it needs the optional's **inner** bound to `parts`
too. I tried that: it makes the optional's fragment carry both a wire list and
a real conversion, at which point "a fragment holding only wires emits nothing"
— the rule that keeps #479's marker out of the generated file — stops holding,
and two externs lost their Kotlin declarations.

That rule is the emitter switch's to restate, since it is what decides which
fragments produce functions at all. So the presence flag goes with the switch
rather than ahead of it.

## Checks

- `cargo test --all` — green

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -43,9 +43,14 @@ pub(crate) struct JFrag {
 pub(crate) struct Wire {
     /// The JNI type this value crosses as.
     pub(crate) ty: syn::Type,
-    /// The path through the value that reached it — `tag`, `summary.count` —
-    /// which is what a JNI parameter name and a Kotlin accessor are both
-    /// mangled from.
+    /// The Kotlin type of the same value, as an `external fun` writes it.
+    pub(crate) kt_ty: String,
+    /// The path through the value that reached it — `tag`, `summary.count`.
+    ///
+    /// **Relative**, because a fragment answers for a crossing and a name
+    /// belongs to a site: the same `Holder` is `hTag` in one signature and
+    /// `otherTag` in the next, so the parameter it hangs off is the caller's to
+    /// prepend.
     pub(crate) path: String,
 }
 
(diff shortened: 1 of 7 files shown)


milyin:
Confirmed and fixed in 81e9bbf — and it was two bugs, not one. Putting the row and the walk side by side for the `Hybrid` fixture you pointed at:

```
composed  hFlatId  hMaybe                  hObjectName
walked    hFlatId  hMaybePresent hMaybeId  hObject
```

**`hObjectName` — your finding.** `ObjectChild` is declared `.jobject_input()` and goes over as one `JObject`; I bound its parts anyway and flattened a boundary the declaration drew. `field_crosses_as_its_fields` is the predicate now and excludes them.

**`hMaybe` — the one next to it.** An `Option<data_class>` field is a presence flag plus the inner's wires, since no one of `(tag, summary)` can carry the absence. I had deferred this as "the emitter switch's problem", and the wider fixture shows it was a live divergence rather than a deferral. The optional keeps the row the registry derived from its shape — it has no `parts` row to take — and it is the inner, one layer down, that takes `parts`. Binding the wrong one got me an exact error from the bindings validator: `asks for row \`parts\`, which Option < FlatChild > (construct) does not have`.

Composing that presence needs a fragment carrying **both** a wire list and a conversion of its own, which is what the emission filter refused — it skipped anything with wires, and that is why my first attempt at this dropped two externs. `composed_only` is the precise fact and only the `parts` row is it.

The equivalence assertion has moved to this fixture. `Holder` — a nested class and plain scalars — is what let both of these through, which is the real lesson: the check was right and its coverage was not.

PR #481, opened 2026-08-21
Carry what Kotlin reads and what the value converts through

milyin:
Continues stage 3 of #472, after [#480](https://github.com/milyin/prebindgen/pull/480). Byte-identical; still nothing switched.

## What a wire was missing

A `Wire` said what it is (JNI type, Kotlin type) and where it sits (path). To
point an emitter at it, two more facts have to come from the same place: the
Kotlin expression that **fills** it, and the conversion it **crosses through**.
Both are now composed the way the type and path already were.

The access is relative, for the same reason the path is: a wire reads
`.flat.id`, and the site turns that into `h.flat.id` in one call and
`this.flat.id` in the next.

- `fields` prefixes the field a part reads, then whatever the part's own wire
  reaches on from there.
- `optional` chains through its gate and supplies the literal a non-nullable
  slot holds while the value is absent — the presence flag is what tells the
  Rust side to ignore it.

## The check this buys

For the `Hybrid` fixture — a flat child, an `Option<data_class>` child, and a
`.jobject_input()` child — the row now reproduces the walk **exactly**:

```
hFlatId        h.flat.id           jlong_to_i64_…
hMaybePresent  h.maybe != null     —
hMaybeId       h.maybe?.id ?: 0L   jlong_to_i64_…
hObject        h.object_           JObject_to_ObjectChild_…
```

So the equivalence assertion now covers names, types, accesses **and**
conversions, where before it covered the first two. That is what makes the
emitter switch a move.

`h.object_` is why the composition calls the walk's own helpers rather than
reimplementing them: `object` is a Kotlin keyword, and only
`kotlin_property_name` knows that. The same applies to `kt_leaf_default` and
`jni_field_access` for the absent-slot literal.

## The handle facts, added in the second commit

A `data_class` field that is an opaque handle crosses as that handle's `Long`,
but the Kotlin scaffold that locks and consumes it needs the handle **object**.
So a wire carries both, and `fields` states the `jlong` itself for such a part
rather than taking the conversion's destination.

Asserted on `Bag { handle: Option<Handle> }`:

```
[("bHandle", "Long", "b.handle", None, Some("b.handle"), true)]
```

— row and walk equal, including that a handle wire has no converter of its own
and that an optional field's access is nullable while its pointer is not.

**With this the row states everything `FlatLeaf` does that a caller reads.**

## What is left before the switch

Pointing `render.rs`, `fn_plan.rs`, `emit/wrapper.rs`, `emit/callback.rs` and
`emit/vec_build.rs` at the fragment, then deleting `emit/flat_input.rs`.

## Checks

- `cargo test --all` — green

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -306,18 +306,7 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
             // non-nullable slot still has to hold something when the value is
             // absent — the flag is what tells Rust to ignore it.
             wires.extend(inner_wires.iter().map(|w| Wire {
-                access: format!(
-                    "?{}{}",
-                    w.access,
-                    crate::jni::emit::kt_leaf_default(
-                        crate::jni::wire_access::jni_field_access(&w.ty)
-                            .map(|(sig, _, _)| sig)
-                            .unwrap_or(""),
-                        w.kt_ty.ends_with('?'),
-                    )
-                    .map(|d| format!(" ?: {d}"))
-                    .unwrap_or_default()
-                ),
+                access: format!("?{}{}", w.access, absent_default(w)),
                 // Anything under the gate can be absent, however the field
                 // itself was spelled.
                 handle_nullable: w.handle_target.is_some() || w.handle_nullable,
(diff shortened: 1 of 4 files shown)


milyin:
Pushed 7942d1d, which adds the handle facts this PR listed as remaining — they are small and belong with the rest of the wire, so they are here rather than in another PR.

A `data_class` field that is an opaque handle crosses as that handle's `Long`, but the Kotlin lock-and-consume scaffold needs the handle **object**. The wire carries both now, and `fields` states the `jlong` itself for such a part rather than taking the conversion's destination.

Asserted on `Bag { handle: Option<Handle> }` — the fixture these facts exist for:

```
[("bHandle", "Long", "b.handle", None, Some("b.handle"), true)]
```

Row and walk equal, including the two things easy to get backwards: a handle wire has no converter of its own, and an optional field's *access* is nullable while its *pointer* is not.

**With this the row states everything `FlatLeaf` does that a caller reads**, so the next PR is the switch itself.

milyin:
Real bug, fixed in 9934244 — though the mechanism turned out to be next to the one you named rather than the one itself.

I built the shape you describe: `Outer { mid: Option<Mid> }` over `Mid { inner: Option<Leaf> }`, so reaching `Leaf.id` passes through two gates. Composed against the walk:

```
composed  o.mid?.inner != null ?: false     o.mid?.inner?.id ?: 0L ?: 0L
walked    o.mid?.inner != null              o.mid?.inner?.id ?: 0L
```

**Not `??`.** An access composed by `fields` always starts with `.` — the presence flag's `" != null"` is composed to `.inner != null` before any outer gate sees it — so the safe-call prefix is well-formed at every depth.

**But the defaulting beside it was wrong, twice over.** The inner gate already substituted a value, so from there on the expression yields one and the outer gate has nothing left to substitute for. And a presence flag is a `!= null` comparison: non-null already, and Kotlin refuses an elvis on a non-null operand outright.

`absent_default` now states both exemptions, and the fixture above is a permanent test. Your reading of the code was right — that block was doing something unconditionally that is only sometimes correct; it was the append rather than the prefix.

PR #482, opened 2026-08-21
Complete the wire: the field it fills and whether it is staged

milyin:
Last piece of stage 3 before the switch, in #472. Byte-identical.

## Scope, measured

I audited which `FlatLeaf` fields the five input-side emitters actually read —
that is exactly the surface the switch has to serve. Thirteen are read outside
`emit/flat_input.rs`, and after
[#481](https://github.com/milyin/prebindgen/pull/481) the composed `Wire`
covered eleven. This adds the last two, both from `emit/vec_build.rs`:

- **`staged`** — whether the conversion carries Rust-side stages beyond its
  wire-facing function. That helper declines such an element rather than emit a
  call binding the representation where the value is wanted.
- **`field`** — the struct field the value fills, which the Rust-side rebuild
  binds by name.

**With this the row states every fact those readers take from a leaf.**

## The one that was not a derivation

`field` looked like pure derivation: the last segment of the path, since a path
**is** the chain of fields that reached the value. It is — for a value.

Doing it for a presence flag yields `present`, a synthetic segment naming no
field, where the walk says `None`. The equivalence check failed the moment I
wrote it that way, on two fixtures at once:

```
composed  ("oMidPresent", …, Some("present"))
walked    ("oMidPresent", …, None)
```

So `Wire::field` states the exemption — a gate says whether the fields beside
it mean anything, and fills none itself — rather than each call site
rediscovering it.

The accessors are `#[cfg(test)]` for now: only the equivalence check calls
them until the emitters take the row, and clippy's `--deny warnings` is right
that they are otherwise dead.

## Checks

- `cargo test --all` — green (266 in `prebindgen-jni`)

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -90,6 +90,15 @@ pub(crate) struct Wire {
     /// build helper declines such an element rather than emit a call that does
     /// not compile.
     pub(crate) staged: bool,
+    /// The struct field this value fills or gates, once a product says which.
+    ///
+    /// `None` until then, and permanently `None` for a gate over a whole value:
+    /// such a gate says whether the fields beside it mean anything and fills
+    /// none itself. A gate over a decoupled scalar is the other case — it gates
+    /// exactly one field, and answers with it.
+    pub(crate) field: Option<String>,
+    /// Whether this wire gates a whole value rather than one field.
+    pub(crate) whole_gate: bool,
 }
 
 impl Carrier for JFrag {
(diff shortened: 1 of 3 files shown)


milyin:
Good catch, and building the shape showed a bigger gap under it. Fixed in 016f598.

You were right that the walked side forwards `field` on a gate leaf. But for an `Option<i64>` field the two sides did not even produce the same **number** of leaves:

```
composed  sN         Long     s.n            JObject_to_Option_i64
walked    sNPresent  Boolean  s.n != null
          sNValue    Long     s.n ?: 0L      jlong_to_i64
```

A nullable primitive or enum with no niche keeps the allocation-free `(present, value)` pair rather than boxing, and the value crosses through the **inner's** conversion — there is no boxed `Option` on that wire to decode. My `optional` hook only decoupled data classes. `decoupled_optional` now states the conditions the walk applies: not a borrow, a JNI-primitive wire, and nothing that already carries the absence.

**On the field itself, normalizing the walked side to `None` would have been wrong.** The walk is right to distinguish: a gate over a whole `Option<struct>` names no field, while a gate over a decoupled scalar gates exactly one and names it. Deriving from the path could not express that — it gave `value` and `present`, the composition's own synthetic segments.

So a wire now carries its field, stated where a product knows it: a part closes what a decoupled optional left open, and a whole-value gate stays `None` for good. My first attempt at that rule gave every gate the field it sat beside, and the two fixtures that already matched failed immediately — which is the check earning its keep.

PR #483, opened 2026-08-21
Remove the extension points no adapter uses

milyin:
**Stage 7** of #472, complete. 102 lines out, 3 in, generated output byte-identical.

The [architecture review](https://github.com/milyin/prebindgen/pull/465#issuecomment-5361741324) objected to shipping unused extension points in the crate a new adapter reads first. These are the two.

## Requirements

`Cx::require` let a conversion declare that it needs a named runtime helper, and `Compiled::required` collected them for the emitter.

Nothing ever declared one. `prebindgen-c` emits `__cbg_alloc_array` through `prerequisites`, and JniGen its runtime helpers the same way — so the mechanism duplicated a route that already works, and the only caller was a test that registered a requirement in order to have something to count. `RequirementId` goes with it.

## Identity

`Construct::Identity` says *the single part named already is the value*. Nothing builds one — no adapter declares such a row and the registry never derives one.

But `Compile::identity` was a **required** trait method, so both adapters carried an implementation whose only job was to refuse a shape they could never be handed:

```rust
fn identity(&mut self, _cx: &mut Cx<'_>, at: At<'_>, _inner: &CFrag) -> Frag<Self> {
    Err(refuse(at, "Cbindgen declares no identity rows"))
}
```

The variant, the hook and both refusals go together. Keeping the hook without the shape leaves a method nothing can call; keeping the shape without the hook does not compile.

## Tests

The two that touched either were exercising the mechanism, not a binding. `a_callback_row_names_its_arguments` keeps every assertion it made about callbacks and loses only the requirement it registered to have something to count; the `Recorder` adapter and its wrapper lose an `identity` hook that was never reached.

## Checks

- `cargo test --all` — green

### DIFF
--- a/prebindgen-c/src/compile.rs
+++ b/prebindgen-c/src/compile.rs
@@ -306,10 +306,6 @@ impl<R: Conversions> Compile for CCompile<'_, R> {
         self.wrap(at, "no C representation for this run", conv)
     }
 
-    fn identity(&mut self, _cx: &mut Cx<'_>, at: At<'_>, _inner: &CFrag) -> Frag<Self> {
-        Err(refuse(at, "Cbindgen declares no identity rows"))
-    }
-
     fn construct(
         &mut self,
         _cx: &mut Cx<'_>,
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -378,10 +378,6 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
         self.wrap(at, "no JNI representation for this run", conv)
     }
 
-    fn identity(&mut self, _cx: &mut Cx<'_>, at: At<'_>, _inner: &JFrag) -> Frag<Self> {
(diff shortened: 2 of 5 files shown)


milyin:
Fixed in 8cf7aa2. `Cx` now says what it is: a read-only view of the model and the table plus the Rust-emission capability, and a note that a hook asks the registry nothing — what it produces is its fragment, and what the registry reads of that is `Carrier::yields`.

This is the third PR in a row where you have caught prose left behind by a deletion, so it is worth naming the pattern rather than just fixing the instance: `cargo doc` with `-D warnings` fails on a **dangling link** and is completely silent about a sentence that has quietly become false. Deleting a public item produces both, and only one of them is gated. I have been sweeping for the deleted identifier by name since #478; here the sentence named no identifier at all — "the one thing a hook may ask the registry to record" — so the grep could not have found it either.

I also swept `identity` while I was in there. The remaining hits are `unfold`'s `DeconRecord::Identity`, which is a live and unrelated mechanism — stage 4's territory, not this one's.

PR #488, opened 2026-08-21
The flatten emitters read the row (#472 §3)

milyin:
Follows #486, which composes the wires a `data_class` parameter occupies. This points the emitters at them.

`FlatInputPlan::leaves` **is** the fragment's wire list. `FlatLeaf` is a projection of one `Wire` paired with the one thing a wire cannot know — the parameter name the site hangs it off — and the walk that used to compute names, JNI types, Kotlin types, access expressions and nullability computes none of them. 780 lines out of `emit/flat_input.rs`, and the generated output is byte-identical: `examples/regen-check.sh` passes over `covertest-kotlin`, `perftest-kotlin` and the two C examples.

## What is left of the walk

The rebuild — the tree the generated Rust reconstructs the struct through. It reads the row rather than deciding again:

- **How many wires a field occupies** is the row's answer, taken off a cursor over the leaves.
- **Which composite shape those wires took** is the model's, taken off the field's type: a sum, a nested `data_class`, or the decoupled `(present, value)` pair.

A sum the row left whole is one wire and needs no special case at all — it lands on the same arm every other single-valued field does. That is the shape of the change generally: the emitter stopped re-deriving what the composition already decided.

## Why a cursor and not a lookup

A path is not a key. A nested `data_class` with a field called `value` reaches the same path a decoupled pair's slot does, and one called `present` reaches the same path as the gate over its own struct — `covertest-kotlin`'s `Annotated` is the first of those and it fails the golden diff immediately.

What is unambiguous is the **order**: the composition emits a gate before what it gates, fields in the model's order, and a sum's tag before its arms' slots. That is the order this walk visits them in, so a cursor cannot mistake one wire for another.

## One behaviour pinned rather than changed

A field spelled `Box<Leaf>` crosses as one value through `Box<Leaf>`'s own conversion, where a **parameter** spelled that way is peeled back to the class and flattened. `Declarations::bindings` now binds a field by its own spelling rather than by the stripped key, which is what the walk always did — binding the stripped key flattened a field the rebuild has no node for. The asymmetry is the existing one; it is stated in the code now instead of falling out of a key choice.

## The equivalence check goes with the walk it checked

`parts_vs_walk_for_test` compared two answers to the same question, and after this there is one answer. Its five fixtures now state what the row says: the whole signature, in the order the three coordinated sites read it — the native wrapper, the `JNINative` extern declaration and the Kotlin call-site destructure.

That is a stronger assertion than agreement was. Agreement passed when both halves changed together; an explicit signature does not.

## Checks

`examples/regen-check.sh` byte-identical, 270 lib tests, clippy and fmt clean.

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -115,9 +115,6 @@ pub(crate) enum Access {
 }
 
 /// The Kotlin property chain, rooted at the object a site destructures.
-///
-/// Only the equivalence check calls it until the emitters take the row.
-#[cfg(test)]
 pub(crate) fn reached(base: &str, walk: &[Nav]) -> String {
     let mut out = base.to_string();
     for nav in walk {
@@ -137,10 +134,6 @@ impl Access {
     }
 
     /// The Kotlin expression, rooted at the object this site destructures.
-    ///
-    /// Only the equivalence check calls it until the emitters take the row —
-    /// the same `#[cfg(test)]` the wire's other readers carry.
-    #[cfg(test)]
     pub(crate) fn render(&self, base: &str) -> String {
         match self {
             Access::Read { walk, tail } => format!("{}{tail}", reached(base, walk)),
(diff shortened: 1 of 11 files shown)


milyin:
Reopened as #488: [#487](https://github.com/milyin/prebindgen/pull/487) closed itself when its base branch `stage3-sum-wires` was deleted on merging [#486](https://github.com/milyin/prebindgen/pull/486), and GitHub will not reopen a pull request whose base is gone. Same commit, rebased onto `registry-api-switch`.

PR #489, opened 2026-08-21
Compose what a JNI sealed class and a data class hand out (#472 §4)

milyin:
The composition half of stage 4, and the deconstructing twin of #486. Nothing takes the row yet, so `examples/regen-check.sh` is byte-identical; what lands here is the row saying exactly what `synth_sum_leaves` says.

This is the direction with no alternative. A sum has **no** whole-value output conversion, because a tag plus one group of values per alternative is not one wire — where a `data_class` parameter could always have crossed as a single `JObject`, a sum returned to the JVM has only ever crossed decomposed.

## The row

`Declarations::recipes` declares a `parts` row for every `sealed_class` on the deconstructing side: `Deconstructing::Choice`, one arm per alternative, each taken apart into its own payload fields. `Compile::fields` answers an arm, `Compile::choice` puts the selector ahead of the groups it chooses between.

**Every payload of an alternative contributes, unconditionally.** That is the asymmetry with the constructing side, where a payload this adapter has no slot form for leaves the whole sum object-shaped. Going the other way the Rust side produces the value and the payload's own output conversion is what encodes it, so there is nothing for an arm to decline.

## What an outgoing value is

`JFrag` gains a second wire list rather than reading the first one backwards, because an outgoing value is a different shape. It is not filled from a Kotlin expression — it is produced by the Rust side and pushed — so what it carries is the builder parameter it lands in, the reading whose output conversion encodes it, the alternative that produces it, and the pattern binding it is reached through. A fragment answers one crossing and a crossing does one job, so the two lists are never both set.

The tag's type is **the sum**, not the `jint` it crosses as. Nothing converts a tag — the emitter assigns the alternative's number in each arm of its `match` — so naming the sum is the only use that type has, and it is exactly the use an emitter has for it: which enum to match over.

## Where it is driven from

`build_with`, not `compile_crossing`. A sum has no deconstructing crossing to be driven from: nothing ever asks for its whole-value output conversion, so the converter walk never reaches it.

It runs last, after `validate_resolved`, and only over sums every payload of which already resolved. A sum with an unresolved payload is a gap in the binding, and the writers report that against the payload — `Reading.Exact.v0 has no OUTPUT converter` — where this could only say that composing failed. What is left for it to catch is a composition that fails over parts which did resolve, which is a bug here rather than a gap there.

## The product beside the choice

A `data_class` states a deconstructing row too: it hands its value out as its fields, so the foreign side reassembles the object in one call and nothing builds one on the Rust side. A field that is itself a `data_class` contributes its own values under the parent's name and chain, which is why `OutFrom` carries a field chain at all — a sum's payload needs none, being bound by its arm's pattern.

That row declines for the **whole** value rather than per field, because the emitter it feeds does. A handle, an `enum_class`, a sum, or a `data_class` behind an `Option` or a `Vec` is delivered with a transform the decoupled form does not carry, and one such field sends the whole object down the whole-value `fromParts` path. A row that decomposed the rest of it would describe a shape nothing emits.

`Declarations::bindings` binds a nested `data_class` field's part in both directions now rather than only in the constructing one. Without it the nested field states one value and the flattening stops a layer down — the same failure the constructing side had before #479.

## Checks

The sum row is held to `synth_sum_leaves` on a fixture carrying a unit alternative (which contributes nothing), a positional payload, a two-field payload and a named-field payload, so the slot naming is exercised in every form it takes. The two agree on names, types, groups and pattern bindings.

The product row is held to `synth_value_struct_leaves` on a fixture with a nested class and a handle-bearing sibling, compared as options — declining is one of the two answers, and a row that states nothing and a synthesis that returns nothing are the same claim.

`examples/regen-check.sh` byte-identical, 272 lib tests, clippy and fmt clean.

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -36,6 +36,15 @@ pub(crate) struct JFrag {
     /// makes a nested `data_class` field contribute its own several rather than
     /// one.
     pub(crate) wires: Option<Vec<Wire>>,
+    /// The values this crossing hands **out**, when it hands out more than one.
+    ///
+    /// The deconstructing twin of [`Self::wires`], and a different shape rather
+    /// than the same one read backwards: an outgoing value is not filled from a
+    /// Kotlin expression, it is produced by the Rust side and named for the
+    /// builder parameter it lands in. Never set at the same time as
+    /// [`Self::wires`] — a fragment answers one crossing, and a crossing does
+    /// one job.
+    pub(crate) out_wires: Option<Vec<OutWire>>,
     /// This fragment states a wire list and nothing else — no conversion of its
     /// own, so nothing of it reaches the generated file.
     ///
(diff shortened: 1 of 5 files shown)


milyin:
Also byte-identical on the real binding: regenerated `zenoh-flat-jni` against this branch and diffed its ~40 generated Kotlin files and its generated Rust against the same regeneration from prebindgen `main`. No difference, which is what a row nobody takes should cost. (Both differ from the files **committed** in that checkout, for reasons that belong to `main` rather than to this branch — see the [note on #472](https://github.com/milyin/prebindgen/pull/472#issuecomment-5369466160).)

PR #509, opened 2026-08-21
Document the model, and fix where the code disagrees

milyin:
Adds `docs/model/`: an overview and three chapters describing the model a
binding is written in — the vocabulary, how the generated file is built, and
directions and crossings. The registry API an adapter programs against, and the
table, sites and hooks around it, are not covered here.

Writing it down exposed three places where the code said something the model
does not, so this changes the code.

**`Shape::Optional` and `Shape::Sequence` lose their `inner` field.** Both were
derivable and neither was checked: the payload of `Option<T>` is `T` and the
element of `Vec<T>` is `T`, but the resolver read the declared `inner` straight
through, so a recipe declaring `Optional { inner: X }` on an `Option<T>` crossing
compiled the wrong part. The two are unit variants now, and the type supplies
the inner.

**`Row::Callback` becomes `Shape::Invoke`.** A callback recipe was derived by
the table and could not be declared, which broke two rules the model states:
every recipe is the adapter's to declare, and every recipe carries a shape. `Invoke` is an
ordinary shape, declared through `RecipesBuilder::declare` like any other. It stays the only
shape a callback crossing may take — converting the arguments is what makes the
callable callable, so there is no second answer — and `RecipesBuilder::build`
refuses anything else there.

**The table's entries settle on one word, `recipe`.** `Recipes`, `RecipeId`,
`RecipeError`, `RecipesBuilder` and the module were named for a recipe; the
enum they hold and every doc comment around them said `Row`. One concept, two
words, and nothing distinguished them. The `Row` enum becomes `Recipe`,
`Ask::Row` becomes `Ask::Recipe`, and the prose follows. Two accessors collided
once the words merged and are renamed for what they answer: `Recipes::rows` is
`Recipes::names_of`, and `Cx::recipes` splits into `Cx::table` and
`Cx::recipe_names`. The error variants had the same confusion in their fields —
`row: CrossingKey` beside `recipe: RecipeId` — and all of them say `crossing`
now.

Two new checks come with the first two changes, since a shape that reads its
inner type off the crossing can be declared on a type that has none:
`RecipeError::WrongShape` for `Optional`, `Sequence` or `Invoke` on a type that is
not one, and `RecipeError::CallbackShape` (was `CallbackDeclared`) for a callback
crossing declared with any shape but `Invoke`.

The rename this does **not** make is `Assembly` to `Direction`. The registry's
converter table has its own `Direction { Input, Output }` naming the same axis,
and the adapters translate between them by hand
(`Direction::Input => Assembly::Construct`). Merging them is converter-table
work. The chapter uses the word and names the type, and says why they differ.

## Checks

- `cargo test --workspace` — 28 suites, all green
- `examples/regen-check.sh` — **byte-identical**, which is the claim: this
  changes how the model is stated, not what any binding generates
- clippy, rustfmt, and `cargo doc` on `prebindgen-registry` — clean


milyin (commented):
One cleanup is still needed before this vocabulary change is internally consistent.

The old public term remains in `Operation::into_row` and `RecipeError::NoSuchRow` (including its display arm and `recipe_of` call site), even though this PR removes `Row` and the new model docs define only recipe. Downstream users would therefore still encounter an undefined legacy noun in the public API. Please finish these as `into_recipe` / `NoSuchRecipe`.

The same pass left several broken rustdocs: `One table of crossings, one recipe per recipe` and `One recipe’s recipe` in `recipe.rs`, while the new `wrong_shape` insertion in `compile.rs` splits the `element_mode` explanation between two functions. The standalone chapters also retain issue-local wording (`this issue` in chapters 01/02 and `this proposal` in chapter 03), and `git diff --check` reports the extra blank line at EOF in `01-vocabulary.md`.

The model-derived Optional/Sequence types and explicit Invoke path otherwise look internally consistent. `cargo fmt --all -- --check`, `cargo test --workspace`, and `examples/regen-check.sh` pass; all GitHub checks are green.

— Codex (GPT-5)

milyin (commented):
Follow-up review of `68c44041` (`One Direction, where there were three`): the semantic mapping is consistent throughout the registry and both adapters—legacy `Input` became `Construct`, `Output` became `Deconstruct`, and callback argument edges still swap direction.

One statement introduced by the unified view is now inaccurate: `docs/model/03-directions-and-crossings.md` says “`swap` has one caller.” There are now three production call sites: recipe validation (`recipe.rs`), recipe compilation (`recipe/compile.rs`), and the legacy registry type-graph scan (`registry/scan.rs`). Saying “only callback argument crossings swap direction” would describe the invariant without asserting an obsolete caller count.

The findings from my previous review (`into_row`, `NoSuchRow`, malformed rustdocs, and issue-local wording) remain unresolved on this head.

Verification: `cargo fmt --all -- --check`, `cargo test --workspace`, and the local Kotlin/JNI covertest all pass. The hosted covertest failure happened before compilation because Maven Central returned HTTP 403 for the Kotlin plugin BOM, so it is not evidence of a code regression. The other hosted checks pass.

— Codex (GPT-5)

PR #510, opened 2026-08-22
Separate Rust planning from final emission

milyin:
## Why

`docs/model.md` is the authoritative description of the binding model. Keeping an implementation audit and migration proposal inside it mixed stable concepts with planned work. This branch keeps the model focused and gives Rust generation a separate, reviewable design with a mechanically enforceable Flat planning boundary.

## Summary

- State the Flat planning boundary in `docs/model.md` and `registry/model.rs`: planning may carry opaque `TypeRef` source descriptions and structural Flat facts, but captured Rust type syntax is available only during final file emission.
- Define construct/deconstruct as the Rust value-to-parts relation and decode/encode as the atomic terminal-to-wire relation.
- Use complete `RecipeKey = (CrossingKey, RecipeName)` row identities throughout declarations, bindings, compilation, fragment lookup and diagnostics. Fragment identity also includes the spelled directed crossing, so `T`, `&T` and `Box<T>` may reuse a row without incorrectly sharing generated behavior.
- Add `docs/rust-writing.md` with the current implementation audit and a restricted conversion architecture: every fragment receives one directional intermediate Rust type; its ABI layout is separate; adapters declare representation bridges and atomic terminal codecs; and the registry constructs product, optional, sequence and choice converter graphs automatically.
- Keep terminal conversion and target representation policy adapter-owned without allowing adapters to render complete composed converter functions during planning. Private converter names derive during final emission from semantic artifact identities rather than serving as dependency keys.
- Specify niche propagation, inactive storage, fallibility, ownership, validity and cleanup as frozen semantic operations instead of early Rust expressions.
- Give `Invoke` callbacks a dedicated frozen bridge while reusing ordinary direction-swapped deconstruct plans for callback arguments. The plan covers C closure structs, null call pointers, takeable values and panic routing, plus JNI global references, daemon attachment, local frames, descriptors, sequence folds, close-unless-taken handles, exceptions and synchronous or asynchronous invocation. A construct-direction result site is reserved for #216.
- Keep one shared Rust-writing pipeline. Adapter renderers translate frozen artifact plans, while the shared writer owns ordering, the sole late `Emit` capability, assembly, formatting, errors and destination writing.
- Preserve public C ABI, JNI ABI and Kotlin behavior semantically rather than requiring byte-identical generated source. Private helpers may be combined, inlined, reordered, deduplicated or removed.
- Migrate through reviewable vertical slices: pin contracts; introduce syntax-free plan vocabulary; build the shared composer; migrate ordinary C and C callbacks; migrate ordinary JNI and JNI callbacks; move all spelling to final emission; then delete and seal the old path.

The proposal preserves model complexity that carries meaning: Flat, crossing dependency order, recipe and fragment identity, reusable fragments versus per-site plans, adapter-specific representations and ABI layouts, recursive-cycle handling, typed failures and callback ownership. It removes generated Rust from planning and the mutable, duplicated state needed for partially generated code and callback emitters to query conversion state.

Related to #195 and the remaining emission work from #251. The JNI migration coordinates with #506 without treating its legacy plan carriers as the replacement architecture.

## Verification

Before the latest documentation expansion, the full branch passed:

- `cargo fmt --all --check`
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`
- `cargo test --workspace`
- `git diff origin/main...HEAD --check`

That run regenerated the checked-in artifacts without changing them.

For the latest documentation-only commit:

- `cargo fmt --all --check` — passed
- `git diff --check` — passed

GitHub Actions for the pushed commit are currently queued.


milyin:
Closing as superseded, with the one still-current part salvaged into #698.

**Why it is superseded.** `main` is 287 commits past this branch's merge base —
#513, #620, #660 and all of #676 have landed since 2026-08-22 — and the branch
conflicts on `docs/model.md`, the one file both sides changed.

Parts of the proposal landed independently, so re-proposing them would reopen
settled work:

- `RecipeKey = (CrossingKey, RecipeName)` as the table's primary key is on
  `main` in `prebindgen-registry/src/recipe.rs`, with the same reasoning.
- "the registry constructs product, optional, sequence and choice converter
  graphs automatically" is what #677 and #678 did; `ShapePlan` and
  `Compile::composed` are on `main`.

Parts describe code that no longer exists. `docs/rust-writing.md` is 923 lines
whose value is being an accurate audit, and of the fifty identifiers it names,
seven are gone from the workspace — `convert_with` (deleted by #677/#678),
`render_artifact`, `post_process_item`, and the `IEncoding` / `IBuffer` /
`IOptionalEncoding` trio.

One claim in it is now false. The proposal says the implementation "violates the
timing part of this invariant by handing `Emit` to conversion planning". #375 is
closed, no `Emit` capability remains in the registry, and `RustWriter` — whose
constructor is registry-private — reaches emission callbacks during final file
assembly and at no other point. The timing half holds by construction.

**What was salvaged.** #698 states the Flat planning boundary itself, in
`docs/model.md` and on `prebindgen-registry/src/registry/model.rs`: planning may
carry a `TypeRef` opaquely and use the model's answers, but may not render or
pattern-match the captured spelling, and a fact available only from the syntax
belongs in Flat. The wire-type carve-out is kept; the stale violation paragraph
is dropped.

**What carries the rest.** The remaining direction — adapters declaring
representation bridges, one shared Rust-writing pipeline, moving spelling to
final emission — is #675's step 4 and its neighbours, deferred with reasons in
#689, "What #676 found". A successor plan should be written from #689 against
the current tree rather than from this branch.

The branch is left in place if any of the audit is worth mining.

— Claude Code (Opus 5)

PR #513, opened 2026-08-22
Transition Rust converter generation to registry-composed chains

milyin:
## Goal

Complete the transition to one registry-driven Rust generation pipeline.

The registry must own recipe selection, recursive source-value traversal, child converter composition, dependency and reachability order, site binding, and the frozen artifact order. Language adapters select target representations and provide semantic bridge operations; they must not rebuild the recipe walk or retain a second converter pipeline.

The Flat-model boundary is strict: planning may inspect Flat semantic facts and retain an opaque `TypeRef`, but it may not obtain or spell the captured Rust type. Actual source Rust types become available only during final file assembly through the writer-owned `Emit`. If planning needs a fact available only from captured syntax, Flat must be extended instead.

This umbrella is complete only when generated Rust is rendered from an immutable registry-produced generation plan, and the old planning/emission path has been deleted rather than retained as a fallback.

## Foundation already in this branch

- [x] Product foundation and C/JNI adoption — #512
- [x] Reachability and generated-code performance follow-up — #515
- [x] Optional-over-Product composition, including wrapper and callback paths — #514
- [x] JNI Choice composition — #516
- [x] Adapter-neutral Choice protocol and C adoption — #517
- [x] Sequence composition for C and JNI — #518
- [x] Callback invocation (`Invoke`) composition for C and JNI — #519
- [x] One wrapper-capability rule, with mutation coverage for every JNI composed planner — #520

The branch now proves every shared shape with both adapters:

- Product: C and JNI, construct and deconstruct;
- Optional: C single-intermediate input; JNI input and output;
- Choice: C and JNI, construct and deconstruct;
- Sequence: C owned-vector output; JNI object-shaped `List<T>` input and output;
- Invoke: C and JNI callback capture, ordered argument preparation, invocation, and cleanup.

Generated foreign-language APIs and application binary interfaces remain semantically compatible. Private Rust helpers may be combined, inlined, reordered, renamed, deduplicated, or removed when that improves the generator.

## Why the migration is not complete yet

The shared composers are still embedded in the old generation architecture:

- C canonical custom conversions, all whole-value input/output terminals, both specialized data-field terminals, tagged-union pointer payloads, and all five borrow terminals now retain frozen operations and spell their source `TypeRef`s only during final rendering (#540, #541, #542, #544, #546, #547). The zero-copy slice input already spelled its element only during final wrapper emission; #548 replaces its remaining untyped dummy `ConverterImpl` carrier with a typed plan. Optional, borrowed-Sequence, and Result multi-wire markers are now typed operations, and C's generic `ConverterImpl` fragment wrapper plus syntax-based validity inference are deleted (#549). Opaque-handle declarations and value-opaque mirrors, transmute glue, destructors, and take helpers are now frozen registry artifacts (#550). Tagged-union mirrors, typed destructors, and recursive payload cleanup are now frozen registry artifacts with explicit nested-union dependencies (#551). Choice's transient Product-arm marker is now typed and `CFunction::Complete` is deleted (#552); the C adapter has no path that can hand a pre-rendered function to the writer.
- JNI Rust, Kotlin, callback-interface, symbol, and report artifacts now share one frozen `JniGenerationPlan` (#524). Exact ordinary-parameter ABI leaves, Rust decode operations, Kotlin call/locking operations, decomposed-return output operations, whole-element iterable-fold pipelines, callback Invoke delivery, and typed domain-error delivery are frozen independently (#536, #537, #538, #539), and `InputKind` is deleted. Bare scalar, all string-terminal, unit, fixed-size primitive-array, `enum_class`, and opaque-handle codecs now retain `TypeRef` until final rendering, and planned dummy markers are explicitly non-emitting (#553, #554, #555, #556, #559, #561). Owned `String`, `Box<String>`, and `Cow<str>` share one frozen value-codec carrier; the latter preserves its captured lifetime instead of normalizing it early. Bare `str` records only adapter-authored `String`/`&str` signatures, and its rank-0/rank-1 output plans share identical converter bodies. Primitive arrays retain only Flat/JNI policy until their full array and element types are spelled during final rendering; a renamed-helper covertest pins const-path qualification. One handle plan now owns consume-input, borrowed-input, owned-output, and clone-on-borrow-output policy; the three eager handle builders are deleted and the zero niche plus odd closed-pointer tag are documented at that plan. A frozen Result plan now owns the fallible-output peel, retains its success converter as a dependency, and carries Flat readings into return-surface classification instead of `syn::Type` (#562). This makes owned and borrowed opaque outputs demand-driven and deletes the wildcard selector plus eager Result builder. A frozen transparent-wrapper plan now owns erased-wrapper construction/deconstruction, retains complete child pipelines and propagates reachability, and spells its stored source `TypeRef` only during final rendering (#563). The two eager wrapper builders and duplicate stage-composition helpers are deleted; private converter names use Flat nominal identity plus an opaque-key hash. Canonical `convert!` function/trait stages now retain Flat source, representation, and error readings until final rendering; their eager body builders, terminal lookups, and wire classification are deleted (#564). Callback Invoke arguments now retain semantic Fold, Decomposed, or Whole delivery operations; their output ABI, converter pipelines, origin qualification, and Flat sum alternatives are frozen before rendering, and the callback compiler no longer receives `Emit` (#565). Whole-object struct input and output codecs now retain Flat struct shape, JNI field policy, and child converter pipelines until final rendering; their eager body builders and terminal fallbacks are deleted (#566). Private names keep only direction, Flat nominal identity when available, and an opaque-key hash. Value codecs remain emitted once compiled, and borrowed-input handle plans alone retain historical unconditional emission until the remaining compatibility parents propagate reachability. The final JNI primitive compatibility fallback is now deleted: scalar terminals are selected from Flat `ScalarKind`, `Vec<u8>` is selected from the structured `Vec<Scalar(U8)>` reading, and both directions retain late value-codec plans (#568). JNI recipe compilation has no path to obtain `Emit`; a source fence rejects both `.emit()` in `compile.rs` and `TypeKey` in the terminal-policy module. The writer-facing `JFunction` carrier can no longer retain or render a complete `syn::ItemFn`; executable converters are frozen semantic plans and legacy rows contribute only non-emitting identities (#569). The shared `ConverterImpl` and `Stage` table rows are now identity-only, `Call::complete` is deleted, JNI validity comes from Flat `Direction`/`Mode`, and `RegistryBuilder::convert_with` no longer receives `Emit` (#570). Legacy `ConverterImpl`, `Stage`, `fn_plan`, and `struct_plan` representations still retain compatibility assembly, but no complete converter syntax.
- Sealed-sum whole-object input now retains Flat alternatives, JVM property policy, and complete child pipelines until final rendering (#567). It shares frozen property planning with whole structs and deletes the eager sum decoder plus duplicate stage assembly. Rust-to-Kotlin sums remain registry-composed flattened `Choice` parts and therefore need no whole-object output codec.
- Registry conversion planning and C/JNI recipe compilation can no longer obtain `Emit`, but the compatibility writer still passes it through registry-bearing `Prebindgen::on_*` callbacks. Neither adapter function carrier nor the shared `ConverterImpl` / `Stage` table can contain a complete `syn::ItemFn`; `fn_plan` / `struct_plan` still provide compatibility assembly around frozen converter identities. C now freezes every non-Invoke `CFunction` as its registry `FragmentPlan` artifact and derives writer input from reached fragments in dependency order; its duplicate `compiled_fns` cache is deleted (#571). Registry reachability prunes unused unit, element-terminal, declared-direction, and whole-Result helpers, removing 307 generated fixture lines without changing the exported C ABI. JNI now drains its mutable compiler store into the immutable `JniGenerationPlan`, freezes each reached fragment's converter artifacts there, derives writer input from those frozen artifacts, and deletes its duplicate `compiled_fns` cache (#572). Registry-owned operation identities now replace adapter-invented JNI converter names; the final writer alone formats bounded semantic Rust symbols, and explicit conversion-contract keys preserve sharing while removing all duplicate converter bodies (#573). The writer groups operations by that identity before rendering, prefers a reachable representative, and retains rendered-name deduplication only for unidentified compatibility functions (#574). Every retained C converter now also exposes an explicit operation identity, with a fixture-wide bijection proving it preserves the old helper-sharing partition before names change (#575). Both adapters still cross an ephemeral adapter-vector seam before the shared writer. Private converter names and call sites now retain only semantic operation identity until final emission (#576), every emitted operation must provide that identity and the rendered-name compatibility deduplicator is deleted (#577), and per-item callbacks now return typed `syn::Item`s without reparsing captured or adapter-produced tokens (#578). The callbacks remain registry-bearing, and JNI still relies on the whole-file `post_process_item` rewrite.
- #506’s deletion list is now a policy inventory, and only one item on it was deletable. `unfold` is the registry’s own decomposition walk and, since #600, where the delivery walk lives; `expand` is the declared surface behind `expand_param!` / `expand_return!`; `Decompositions` is a public input, the route a registry with no adapter declarations states its demanded output decompositions by; `ConverterImpl` and `Stage` have been identity-only rows carrying no function bodies since #570; `fn_plan.rs` and `struct_plan.rs` hold JNI policy — JVM property reads, `jvalue` layout, locking — rather than a second source walk. `InputKind` is deleted (#536). What is still true of #506’s subject is narrower and belongs to parent 5: `pre_stages` is not yet folded into the converter operation graph. See the reconciliation comment on this PR.

Green tests therefore prove that the new composers coexist correctly with the compatibility pipeline; they do not yet prove that the compatibility pipeline is removable.

## Completion plan

The vocabulary foundation may land before a production path can consume it. Every adapter-migration implementation slice must move one complete production path off duplicate planning; temporary old/new comparison is allowed in tests, not in production generation.

Some compatibility helpers serve more than one production path and cannot be deleted by the first slice alone. Such paired slices form one explicit deletion unit: the first removes its migrated callers, and the second must remove the helper and remaining fallback rather than extending it.

- [x] **1. Freeze the shared generation vocabulary.** Add semantic `FragmentId`, `SiteId`, and `ArtifactId`; immutable fragment and site plans; intermediate and ABI-layout payloads; typed converter, ownership, validity, niche, failure, and cleanup operations; and deterministic reachability/artifact order. The registry owns this generic store while adapters supply associated representation payloads. — #521
- [x] **2. Move ordinary C sites onto frozen plans.** Make C function parameters, results, error routes, and cleanup consume registry-compiled site plans. Represent Optional output and pointer-plus-length Sequence layouts explicitly, preserve zero-copy slice and `MaybeUninit` policies, and remove every ordinary-path call to structural lowering or registry queries. Callback-only compatibility helpers remain until stage 3; stages 2 and 3 are one deletion unit. — #522
- [x] **3. Finish C Invoke reuse and complete the C deletion unit.** Make callback arguments reference the same deconstruct fragment/site plans as ordinary outputs. Keep NULL-call guarding, takeable graveyard slots, zero-copy borrowed slices, panic routing, and post-call cleanup as C bridge policy; delete `lower_shape`, `encode_value`, callback marker converters, callback-time registry queries, and the compatibility state they made necessary. — #523
- [x] **4. Freeze one JNI generation plan shared by Rust and Kotlin.** Promote the current function, struct, interface, projection, expansion, lock, error-channel, callback, and ABI-leaf decisions into one immutable `JniGenerationPlan` built during resolution. Rust JNI wrappers, native symbols/descriptors, Kotlin wrappers, `JNINative` declarations, data/sealed classes, callback interfaces, and the report must consume this same plan rather than lazily querying the registry or rebuilding classifications. — #524
- [ ] **5. Move every ordinary JNI representation onto fragment/site plans.** Fold `pre_stages` into the converter operation graph; cover flattened and whole-object Product/Choice forms, deep Optional forms, primitive arrays, handle vectors, borrowed sequences, multi-slot folds, and value forms. Delete each superseded `flat_input`, `delivery`, `struct_plan`, `fn_plan`, `InputKind`, selector, and compatibility branch in the same slice that replaces it. **Re-scoped after #596:** every slice below landed, `InputKind` is the only deletion target that proved to be one, and the other named modules are policy homes rather than superseded duplicates. What remains of this parent is `pre_stages`, and the whole-object output encode #602 tracks.
  - [x] Move allocation-free Optional parameter and expansion-leaf reconstruction onto a distinct registry `pair` row; delete `OptionScalarInputPlan` and its handwritten ordinary/expanded Rust assembly. Kotlin and the JNI ABI stay unchanged. — #525
  - [x] Freeze ordinary parameter, constructor-expansion-leaf, and value-return converter pipelines in JNI site plans; delete wrapper-time converter lookups and handwritten stage ordering. — #526
  - [x] Route direct owned-handle parameters and constructor-expansion leaves through their frozen site pipelines; delete duplicate inline ownership reconstruction and the obsolete `InputKind::Handle.direct` classification. — #527
  - [x] Route transient Vec-handle inputs through frozen site pipelines; retain the Kotlin push-helper ABI while deleting wrapper-local borrow/move reconstruction and Rust-only `VecBuild` fields. — #528
  - [x] Make flattened data-class inputs require an exact registry-composed Product/Optional/Choice chain; delete JniGen’s parallel `FlatStructNode` reconstruction tree and retain the specialized Vec push ABI only for terminal Products. — #529
  - [x] Include expansion-synthesized Optional readings through an on-demand-derived, representation-neutral registry iterator so nullable primitive constructor leaves use the registry `pair` row instead of boxed `JObject` unboxing. Make the shared writer reject calls to planned converters removed by reachability before it writes an invalid destination. Kotlin public declarations stay unchanged; the private JNI boundary switches affected leaves to presence plus raw-value primitives. — #530
  - [x] Freeze constructor-expansion folds and recursively qualified constructor calls in JNI site plans; remove render-time `Registry` and `Declarations` access from expanded-parameter reconstruction while keeping every Rust, Kotlin, and report consumer on the same fold. — #531
  - [x] Retain borrowed `Option<&T>` and `Option<&mut T>` handle reconstruction as a late registry plan; delete the eager `OptionRef` branch while preserving the non-owning carrier and final borrow projection. — #532
  - [x] Compose owned `Option<opaque>` through the shared Optional plan and its owned-handle child; delete the eager direct-handle Option converter and prune unreachable optional helpers. — #533
  - [x] Delete the remaining JNI structural Optional compatibility pipeline; compose borrowed `Option<&data_class>` through an owned registry carrier, preserve Kotlin-side flattening, prune unused whole-object decoders, and pin unsupported Optional shapes at the compiler seam. — #534
  - [x] Make registry Sequence fragments own ordinary `Vec<T>` and shared `&[T]` JNI conversion; key borrowed inputs on their produced Flat-model vector carrier, reuse one decoder across owned and borrowed crossings, preserve specialized Kotlin folds as composed-only Sequence fragments, and delete the parallel wrapper/selector/list-converter pipeline. — #535
  - [x] Freeze each JNI parameter leaf's exact ordered native ABI, Rust decode operation, and Kotlin call/locking operation independently; make Rust wrappers, `JNINative`, and JVM-slot validation consume those frozen answers; delete `InputKind` and pin every two-slot ABI path by mutation-resistant limit tests. — #536
  - [x] Freeze ordinary decomposed-return leaf operations and whole-element iterable-fold pipelines before rendering; reuse exact Product/Choice/value-form child operations, compile function-unique registry unfold leaves once, and retain callback/error assembly only as the explicit stage-6 compatibility boundary. — #537
- [x] **6. Finish JNI Invoke reuse.** Make every callback argument use its ordinary deconstruct plan, including stages, flattened data, optional gates, sequence folds, primitive/object `jvalue` layout, and owned-handle transfer. Freeze descriptors, local-frame sizing, daemon attachment, cached method lookup, exception handling, close-unless-taken cleanup, and Kotlin interface leaves before rendering; remove callback-time registry lookups and independent conversion assembly.
  - [x] Retain every callback argument's registry-compiled deconstruct fragment and exact outgoing ABI operation; remove the callback atomic bypass, compatibility recipe row, handwritten stage assembly, and renderer-time chain lookup. Rust-only expanded callback inputs remain ordinary composed-only fragments. — #538
  - [x] Freeze typed domain-error delivery as an exact registry output plan; delete the ABI-less `Delivered::of` constructor plus generic and Choice renderer-time converter/stage fallbacks. Generated artifacts remain byte-identical. — #539
- [x] **7. Move terminal codecs and all source spelling to final emission.** Replace complete functions inside `ConverterImpl`, `Stage`, `CFunction::Complete`, and `JFunction::Complete` with frozen terminal operations. Only the final adapter renderer receives `Emit`; source qualification happens while stored `TypeRef`s are spelled.
  - [x] Retain canonical C scalar `convert!` declarations as frozen function/trait operations with domain and niche facts; spell and qualify their source `TypeRef` only in final `RustFunction::render`, and bypass `ConverterImpl::function` for both directions. Generated artifacts remain byte-identical. — #540
  - [x] Retain all seven whole-value C output terminals (unit, `String`, scalar, owned handle, opaque error, value-opaque, and enum) as frozen semantic plans; remove C recipe compilation's last `Emit` use and spell stored source `TypeRef`s only during final rendering. Pin every operation at the planning/rendering seam. Generated artifacts remain byte-identical. — #541
  - [x] Retain all seven whole-value C input terminals (owned handle, value-opaque, enum, `String`, `str` marker, `bool`, and scalar) as frozen semantic plans; type value-opaque write-back as none/null-fields/gravestone, delete their eager `ConverterImpl` builders, and spell source `TypeRef`s only during final rendering. Generated artifacts remain byte-identical. — #542
  - [x] Retain bare, optional, boxed, and optional-boxed C tagged-union pointer payloads as frozen semantic plans in both directions; preserve NULL and ownership behavior, delete their eager `ConverterImpl` builders, and spell full/inner source `TypeRef`s only during final rendering. Generated artifacts remain byte-identical. — #544
  - [x] Retain lenient `String` field input and normalized `bool` field output as distinct frozen terminal operations; delete their eager `ConverterImpl` builders and spell their source `TypeRef`s only during final rendering. Generated artifacts remain byte-identical. — #546
  - [x] Retain `&str`, shared `&T`, mutable `&mut T`, `&mut MaybeUninit<T>`, and shared-output `&T` C crossings as five frozen borrow operations; delete their eager `ConverterImpl` builders, split the unrelated `Result<T, E>` marker fallback, and spell referents only during final rendering. Generated artifacts remain byte-identical. — #547
  - [x] Replace the C shared-slice input's dummy `ConverterImpl` with a typed plan carrying element identity, pointer wire, and the scalar/value-opaque reinterpretation decision. The zero-copy decoder already used final-emission source spelling; this slice deletes its remaining untyped compatibility carrier. Generated artifacts remain byte-identical. — #548
  - [x] Replace Optional, borrowed-Sequence, and Result output marker converters with typed operations retaining exact dependency edges; delete C's generic `ConverterImpl` fragment wrapper, syntax-based validity inference, fallback method, and final production import. Fence every C source module against reintroducing an `Option<ConverterImpl>` path. Generated artifacts remain byte-identical. — #549
  - [x] Freeze opaque-handle declarations and value-opaque mirrors, transmute glue, typed destructors, and take helpers as registry artifacts with fragment dependencies; retain source `TypeRef`s until final rendering and delete the eager prerequisite paths. Generated artifacts remain byte-identical. — #550
  - [x] Freeze C tagged-union mirrors, typed destructors, and recursive payload cleanup as registry artifacts; retain boxed payload `TypeRef`s until final rendering, express nested-union cleanup as artifact prerequisites, and delete the eager prerequisite/cleanup/tag-guard path. Generated Rust and headers change only by dependency-correct declaration ordering. — #551
  - [x] Replace Choice's transient Product-arm dummy function with a typed marker, keep its exact dependencies in the composed Choice, and delete `CFunction::Complete` plus the final C render escape hatch. Generated artifacts remain byte-identical. — #552
  - [x] Retain bare JNI scalar codecs as typed plans, select them before the compiler obtains `Emit`, spell their source `TypeRef` only during final rendering, and make copied planned markers structurally non-emitting. Generated Rust and Kotlin remain byte-identical. — #553
  - [x] Generalize the JNI scalar carrier into a terminal-value codec plan and move owned `String`, `Box<String>`, `Cow<str>`, and unit output onto it; delete their eager terminal builders and render captured source spelling only at final emission. The emitcheck golden changes only to preserve `Cow<'static, str>` and rename its private converter; Kotlin and JNI ABI stay unchanged. — #554
  - [x] Move bare unsized `str` input/output and borrowed `&str` output onto frozen value-codec plans with explicit adapter-authored `String`/`&str` signatures; keep the crossing opaque until final rendering, preserve one safe shared output identity with identical bodies, and delete the eager string-slice compatibility builders. Generated Rust and Kotlin remain byte-identical. — #555
  - [x] Move fixed-size JNI primitive arrays onto the frozen value-codec carrier; retain array/JNI policy while deferring full-array and element source spelling to final rendering, and delete both eager terminal branches. Existing generated surfaces remain byte-identical; a covertest-only helper fixture adds `[u8; cov_helpers::CONST_ARRAY_LEN]` and a JVM round trip to pin cross-source const qualification. — #556
  - [x] Retain JNI `enum_class` input/output terminals as frozen value-codec plans; freeze Flat variant identities, discriminants, source origin, `jint` representation, and nested-Optional niches, spell the source `TypeRef` only during final rendering, and delete the eager enum body builders and terminal branches. Generated Rust and Kotlin remain byte-identical. — #559
  - [x] Retain all four JNI opaque-handle terminals—owned consume input, non-owning borrowed input, owned output, and clone-on-borrow output—as one frozen handle-codec plan; delete the eager input/output/borrow builders, preserve the zero niche and odd closed-pointer tag, and spell source `TypeRef`s only during final rendering. Generated Rust and Kotlin remain byte-identical. — #561
  - [x] Retain JNI `Result<T, E>` output peeling as a frozen plan carrying source/success/error Flat readings and the success converter dependency; replace `KotlinMeta`'s stored `syn::Type` with `TypeRef`, delete the wildcard selector and eager Result builder, and make opaque output codecs demand-driven. Generated Kotlin and the JNI ABI remain semantically unchanged; private unused Rust helpers are pruned. — #562
  - [x] Retain erased JNI transparent-wrapper crossings as one frozen bidirectional plan carrying wrapper policy, the complete child pipeline, and child reachability; spell the source `TypeRef` only during final rendering, delete both eager bridge builders and duplicate stage-order helpers, and name private helpers from Flat nominal identity without reading `TypeKey` text. Generated Kotlin and JNI ABI remain semantically unchanged; one unreachable private decoder is pruned. — #563
  - [x] Retain canonical JNI `convert!` function-backed and trait-backed stages as frozen semantic operations; preserve fallibility, representation domains, niches, metadata, and composed wire converters while deleting eager body builders and terminal lookups. Flat source, representation, and error readings are spelled only during final rendering; helper names use direction, nominal model identity, and an opaque-key hash. Generated Kotlin and JNI ABI remain semantically unchanged. — #564
  - [x] Retain JNI callback arguments as semantic Fold, Decomposed, or Whole Invoke operations; freeze exact output ABI, converter pipelines, origin qualification, and Flat sum alternatives, remove callback planning’s final `Emit` access, and render the source-value walk only from the final registry Invoke pass. Generated Rust and Kotlin remain byte-identical. — #565
  - [x] Retain JNI whole-object struct input and output codecs as frozen plans carrying Flat shape, JNI property policy, and complete child pipelines; delete the eager struct-body builders and terminal fallbacks, and spell the source crossing only during final rendering. Generated Rust and Kotlin remain byte-identical. — #566
  - [x] Retain JNI sealed-sum whole-object input as a frozen plan carrying Flat alternatives, JVM property policy, and complete child pipelines; share property planning with whole structs, delete the eager sum decoder and duplicate stage assembly, and construct alternatives only during final rendering. Rust-to-Kotlin sums remain flattened `Choice` composition. Generated artifacts remain byte-identical. — #567
  - [x] Select JNI scalar terminals from Flat `ScalarKind` and byte vectors from the structured `Vec<Scalar(U8)>` reading; retain both directions as late value-codec plans, delete the final input/output terminal fallback and eager function builders, and fence recipe compilation against `Emit` and terminal policy against `TypeKey`. Generated artifacts remain byte-identical. — #568
  - [x] Delete JNI's complete-function render escape hatch; restrict `JFunction` to frozen semantic plans or identifier-only non-emitting compatibility markers, and remove the resulting dead private `__jni_parts` helper. Generated Kotlin and JNI ABI remain unchanged. — #569
  - [x] Make the shared converter table identity-only: replace complete functions in `ConverterImpl` / `Stage`, delete `Call::complete`, remove `Emit` from `RegistryBuilder::convert_with`, and derive JNI validity from Flat `Direction` / `Mode`. Fence both the carrier and planning capability boundaries. Generated artifacts remain byte-identical. — #570
- [ ] **8. Cut over the shared writer and delete the old pipeline.** Render already-ordered artifact plans directly to `syn::Item`s. Remove planning-time `Emit`, resumable mutable compiler caches, `compiled_fns`, registry-bearing item callbacks, token reparsing, rendered-name deduplication, and the whole-file post-process rewrite. Complete #506 and add call-graph/compile-fail checks enforcing both sides of the planning/rendering boundary. **After #581 and #596** every clause here has landed except one: the writer takes a frozen assembly and no registry, planning-time `Emit`, the mutable compiler caches, `compiled_fns`, registry-bearing item callbacks, token reparsing, rendered-name deduplication and the whole-file post-process rewrite are all gone, the call-graph check refuses an artifact whose declared callee is not reached, and the compile-fail doctests seal the planning side of the boundary. The rendering side is sealed by an interface rather than by a check: `RustArtifact::render` cannot reach a registry, and `write_kotlin` still receives one.
  - [x] Store each frozen C converter artifact on its registry `FragmentPlan`, derive emission from the plan’s reached dependency order, delete the C `compiled_fns` cache, and prune unreachable private helpers. The shared writer’s ephemeral vector seam and C’s mutable resolution cache remain for the following cutover. — #571
  - [x] Freeze each reached JNI fragment's converter artifacts in `JniGenerationPlan`, drain the mutable compiler store at freeze, derive writer input from the immutable plan, and delete the JNI `compiled_fns` cache. The early adapter-generated converter identifiers and shared writer seam remain for the next slice. — #572
  - [x] Move converter operation identity into the registry, format readable semantic symbols only in final emission, and replace incidental name collisions with explicit contract-based sharing. Generated externs stay unchanged and duplicate private converter bodies fall to zero. — #573\n  - [x] Deduplicate semantic operations before rendering, selecting a reachable representative for identities retained by multiple fragments; leave rendered-name deduplication only as a compatibility path for unidentified functions. — #574\n  - [x] Give every retained C converter an explicit semantic `OperationId`; prove with a fixture-wide bijection that registry identity describes exactly the existing helper-sharing partition before final-name migration. — #575
  - [x] Allocate C private converter names and composed call sites from semantic `OperationId` only during final emission; keep readable semantic stems and remove the adapter's early identifier path. — #576
  - [x] Require semantic identity for every private converter, remove rendered-name compatibility deduplication and complete-function carriers, and order emitted operations by registry identity. — #577
  - [x] Return typed `syn::Item`s from every item callback, remove writer token reparsing and whole captured-item emission helpers, and seal captured syntax from `Origin` diagnostics. — #578
  - [x] Take the whole generated file from one frozen assembly, prove dependency completeness from declared artifact edges, and reduce `Prebindgen` to validation. — #581
  - [ ] Complete #506 — re-scoped by the reconciliation comment: its carriers are current machinery, and what is left of its subject is parent 5’s `pre_stages`.

## Kotlin’s place in the model

Kotlin generation is in scope where it shares a contract with generated Rust: JNI method names and descriptors, wire arity and types, parameter/return leaves, callback interface signatures, handle ownership and locks, error channels, and expansion/fold layout must come from the same frozen JNI plan.

Kotlin language policy remains adapter-owned. The registry must not know about `KtType`, packages, imports, classes, overloads, Kotlin formatting, JVM naming policy, or filesystem layout. It should help all language generators only with generic generation infrastructure:

- stable fragment, site, and artifact identities;
- ordered immutable plan storage and reachability;
- the association between a site, its selected fragment, and adapter-defined ABI layout;
- generic arity, direction, ownership, validity, failure-route, and dependency validation; and
- a final Rust-rendering boundary that receives one frozen adapter artifact and `Emit`, but no registry lookup capability.

JniGen then owns `JniGenerationPlan` payloads and two pure consumers: the shared Rust writer through its JNI renderer, and the Kotlin writer through its Kotlin renderer. The Kotlin writer may remain a separate destination writer, but it must not derive a second JNI ABI after the plan is frozen.

## Completion gates

- **Holds.** No planning or validation path can obtain `Emit`, source `syn::Type`, or rendered source tokens — nine `compile_fail` doctests hold it across the crate boundary (three on `RustWriter`, six in the flat rendering module), over per-module source fences and a `pub(crate)` rendering capability.
- **Holds for Rust, not for Kotlin.** `RustArtifact::render(&self, emit)` is the whole Rust interface; `write_kotlin` still takes `&Registry`, though it reads the frozen plan rather than resuming compilation.
- Every reached fragment has one intermediate type; every reached site has one adapter ABI layout referencing that fragment.
- **Holds** (#523). C ordinary calls and callbacks reuse the same site/fragment plans.
- **Holds** (#524 and the slices under parent 7). JNI Rust wrappers, Kotlin declarations, native descriptors, and callbacks consume the same `JniGenerationPlan` decisions.
- **Nearly.** Product, Optional, Sequence, Choice, and Invoke source traversal exists only in the registry composer, except the whole-object output encode, which flattens a struct’s fields into the leaves a JVM constructor is called with. It is a superset of the registry-facing decomposition — which refuses a nested data class behind `Option` or `Vec` — so composing it would drop delivery coverage; #602 carries the missing support, and a standing check compares the two derivations over leaf names and nesting depth.
- **Half.** The compatibility writer path is deleted (#581). #506’s carriers are not, and are reclassified above as current machinery.

## Compatibility and verification

Each child PR must classify generated-source changes as private restructuring, qualification/formatting, an explicit optimization, or an intentional public contract change. Exported C symbols/signatures/layouts, JNI native names/descriptors, and generated Kotlin public behavior remain compatible unless a separate change explicitly says otherwise.

The current umbrella head passes GitHub Actions for stable and Rust 1.85 builds, documentation, packaging, the Kotlin/JNI covertest, and the ASan smoke test. Across the landed child PRs it has also been checked with:

- `cargo test --workspace`
- `cargo fmt --all --check`
- `git diff --check`
- `cd examples/covertest-kotlin && ./gradlew run` — all 61 runtime sections passed
- `bash examples/perftest-bench.sh --quick` — completed; C correctness and RSS checks passed

Related to #510 and #506.


milyin:
Opened #596 as the successor umbrella on this branch, stacked the way #581 was.

Reviewing this branch against the merged #581 turned up four things it did not reach — three source-value walks JniGen still performs itself (whole-object struct input and output, sealed-sum whole-object input, and decomposed delivery), and `RepresentationDomain` deciding a scalar's identity by matching its rendered spelling. #596 restates them from the code with line references and plans them as seven reviewed children.

Two notes on this description, which #596's first section also records:

- It is stale in one direction: `post_process_item`, `compiled_fns`, the `on_*` callbacks, `Emit` reaching planning through them, and the ephemeral adapter vector seam are all deleted, with fence tests keeping them deleted. `InputKind`, `Call::complete`, `CFunction::Complete` and `JFunction::Complete` are gone too.
- And in the other: items 5, 7 and 8 are unchecked parents whose listed children are all ticked. What actually holds 5 and 7 open is their deletion clause — `flat_input.rs`, `struct_plan.rs`, `fn_plan.rs` and `delivery.rs` still exist — and that is what #596 steps 2 to 5 address. Item 8 looks complete to me after #581.

I have left this description alone rather than editing it; #596's last step reconciles it once its own work lands, including a decision on whether the #506 carriers (`expand`, `unfold`, `Decompositions`, `ConverterImpl`, `Stage`, `fn_plan.rs`) are current machinery or legacy — my reading is the former, but this umbrella has not recorded it either way.

— Claude Code (Opus 5)

milyin:
## Where this branch stands after #581 and #596

**#581 made the writer a pure consumer.** `write_rust` takes one frozen
`Assembly` — the artifacts, the rendering capability minted at freeze, and
prebindgen's injected feature guards — and no registry. Dependency
completeness is proven from edges each artifact declares (`calls`, `provides`)
instead of by walking rendered syntax, and `Prebindgen` is down to `validate`
and `validate_resolved`, so no item callback carries a registry. Walking
`syn` survives only as test support.

**#596 moved the source walks the adapters still owned.** The registry gained
fallible part reads, one complete child per whole-object property, composed
`Product` and `Choice` for JVM whole-object input and sealed sums, and the
delivery walk itself. Three of the four walks over a source value that JniGen
owned are gone — whole-object struct input, sealed-sum whole-object input and
delivery; the fourth is gate 6's remainder. A representation domain now
carries its scalar kind, so no adapter decides a scalar's identity from a
rendered spelling (#589).

### The completion gates, one by one

1. **No planning or validation path can obtain `Emit`, source `syn::Type`, or
   rendered source tokens** — holds, and a compile-fail suite is what holds
   it. Nine `compile_fail` doctests cover the gate's three subjects, and
   because a doctest compiles against the public API they cover them across
   the crate boundary: three on `RustWriter` — its constructor is
   registry-private, a registry-only adapter cannot name the flat rendering
   protocol to supply a receiver of its own, and the facade exposes no
   `spell` — and six in the flat rendering module, closing `spell` on a type
   and on a function's origin, `as_syn` on a type and on an element,
   `stripped_syntax`, and `to_syn`. #375 restored the first three. Per-module
   source fences and the `pub(crate)` rendering capability sit under them.
   Three further compile-fail doctests seal model construction rather than
   rendering, so the boundary has twelve in all.
2. **No final renderer receives `&Registry`, resumes compilation, or selects a
   fallback** — holds for Rust: `RustArtifact::render(&self, emit)` is the
   whole interface. **It does not hold for Kotlin**: `write_kotlin` still takes
   `&Registry`. It reads the frozen plan rather than resuming compilation, so
   the second half of the gate holds and the first does not.
3. **One intermediate type per reached fragment, one ABI layout per reached
   site** — unchanged by these two umbrellas; whatever #521 and #536 left.
4. **C ordinary calls and callbacks reuse the same plans** — unchanged (#523).
5. **Every JNI consumer reads one `JniGenerationPlan`** — unchanged (#524 and
   the slices under parent 7).
6. **Product, Optional, Sequence, Choice and Invoke source traversal exists
   only in the registry composer** — nearly. What remains adapter-side is one
   walk: the whole-object output encode, which flattens a struct's fields into
   the leaves a JVM constructor is called with. It is a **superset** of the
   registry-facing decomposition, which refuses a nested data class behind
   `Option` or `Vec` — 11 of 29 declared structs — so composing it out of that
   decomposition would drop delivery coverage. #602 carries the missing
   support. Until then a standing check compares the two derivations on every
   declared data class, over leaf names and nesting depth.
7. **#506's legacy carriers and the compatibility writer path are deleted** —
   the writer path is (#581). The carriers are not, and the gate is the wrong
   question for them; see below.

### #506's carriers are current machinery, not legacy

#506 asked for the deletion of `expand`, `unfold`, `Decompositions`,
`ConverterImpl`, `Stage` and `fn_plan.rs` as "what the registry API replaced".
After the two umbrellas that is true of none of them:

- **`unfold`** is the registry's own decomposition walk, and since #600 it is
  where the delivery walk lives too. Deleting it would delete the composer.
- **`expand`** is the declared surface behind `expand_param!` /
  `expand_return!`, not a second pipeline.
- **`Decompositions`** is a public input: a binding states the output
  decompositions it demands before `resolve`, which is the only route for a
  registry built with no adapter declarations at all.
- **`ConverterImpl` and `Stage`** have been identity-only rows since #570 —
  they carry no function bodies, and the identity they carry is what the
  writer groups and names by.
- **`fn_plan.rs` and `struct_plan.rs`** hold JNI policy — JVM property reads,
  `jvalue` layout, locking — rather than a second source walk. #596 emptied
  them of walking, not of policy.

One carrier from that list is gone outright: `InputKind`, deleted by #536.

So #506 is not a deletion waiting to happen. What is still true of its subject
is narrower and belongs to parent 5 below: `pre_stages` is not yet folded into
the converter operation graph.

### Parents

- **Parent 7 is complete** and is ticked. Every terminal codec is a frozen
  operation, `CFunction::Complete` and `JFunction::Complete` are gone (#552,
  #569), the shared rows are identity-only (#570), and source spelling happens
  in final rendering.
- **Parent 8 is complete except its last clause.** The writer cutover, the
  removal of planning-time `Emit`, the mutable compiler caches, `compiled_fns`,
  registry-bearing item callbacks, token reparsing, rendered-name
  deduplication and the whole-file post-process rewrite all landed (#571 to
  #578, #581). What remains is "complete #506 and add call-graph/compile-fail
  checks enforcing both sides of the planning/rendering boundary". Both check
  kinds exist: the writer refuses an artifact whose declared callee is not
  reached, a test asserts the declared edges cover every rendered call, and the
  compile-fail doctests of gate 1 seal the planning side. #506 is re-scoped by
  the section above rather than completed, and the rendering side is sealed by
  an interface rather than by a check — `RustArtifact::render` cannot reach a
  registry, and `write_kotlin` still receives one.
- **Parent 5 is re-scoped.** Its slices all landed, and its deletion list
  (`flat_input`, `delivery`, `struct_plan`, `fn_plan`, `InputKind`, selectors)
  named modules that are now policy homes rather than superseded duplicates —
  `InputKind` is the only one that was deletable, and it is deleted. What is
  left of the parent is `pre_stages`, and the whole-object output encode that
  #602 tracks.

### What #596 stranded

`PartRead::statements` — a part read that binds without failing — had no
production caller and is deleted.

— Claude Code (Opus 5)

milyin:
The reconciliation comment matches the text #605 proposed, verbatim apart from
the long line being rewrapped. The five body edits all landed: parent 7 ticked,
parent 5 re-scoped, parent 8 carrying `#506` as its one remaining sub-bullet,
the "remain live" bullet replaced, and the gates annotated.

One defect the edit left behind, in the section it touched. Parent 8's sub-bullet
list contains two **literal** `\n` sequences rather than line breaks, joining
three entries into one:

```
… duplicate private converter bodies fall to zero. — #573\n  - [x] Deduplicate
semantic operations before rendering … — #574\n  - [x] Give every retained C
converter an explicit semantic `OperationId` … — #575
```

GitHub renders that as a single bullet with `\n` visible in the prose, and only
the `#573` checkbox renders as a checkbox — `#574` and `#575` become inline text.
So three completed slices read as one, and the parent's progress undercounts by
two.

Replacing the two `\n` with real newlines fixes it; nothing else in the body has
the same artifact.

This is the same edit-script failure named on #605, in its second form: there,
a script aborted before writing and the upload did not notice; here, a script
wrote an escape sequence instead of the character it stands for. Both produce a
document that looks edited and is not, which is why reading the body back after
an edit is worth the one minute it costs.

— Claude Code (Opus 5)

milyin:
## Delivery is composed now — correcting an earlier claim

The step-7 comment above lists delivery among the walks #596 moved. That was
wrong when it was written, and #607's own first step says so: #596 moved
delivery's mechanical half — folding a path, binding the hoists, the flat
leaf reach — and left five rules about Rust values in
`prebindgen-jni/src/jni/emit/delivery.rs`. The walk had no bridge to hand the
encoding half to, so nothing entangled with encoding could cross.

#607 finished it. `DeliveryBridge` (#608) gave the walk the adapter's half —
encoding a leaf, placing it as a call argument, what a slot holds unfilled,
what absence looks like. Then the rules crossed: one leaf reach with the
gating and the ownership terminal in it (#609), where a leaf's reach starts
and whether the place is moved (#610), and segment recognition with the
optional-sum gate and the conditional arm (#611).

`emit/delivery.rs` walks nothing now, and its module doc says so. It holds
the `DeliveryBridge` implementation and JNI policy: `jvalue` layout,
local-frame sizing, cached method lookup, exception routing.

**What this cost.** Measured across the whole umbrella, excluding tests and
generated files: production Rust +971 / −557, net +414 lines — the registry
gained 553 and JniGen shed 139. `emit/delivery.rs` went 1515 lines to 1360. Generated output changed in 11
places, all in #609 and all the same spelling change — a non-owned field read
is `(&v.f).clone()` everywhere now, which is the spelling that also holds
when the field is itself a reference. Cbindgen is untouched: it declares no
decomposition.

**Gate 6 of this PR now reads true.** It was written during #596's
reconciliation on the belief that delivery had moved; the whole-object output
encode it names as the exception is still the exception, and #602 still
carries it.

One thing #607 did not change, and it is worth stating rather than leaving to
be discovered: `walk.rs` has one consumer and `DecomposedLeaf` one production
implementation. Cbindgen has no decomposition feature and none is proposed.
The walk is in the registry because a decomposition walk is the registry's
subject and one home means one implementation — not because a second adapter
reuses it today.

— Claude Code (Opus 5)

PR #517, opened 2026-08-23
Make Choice chain composition usable by C and JNI

milyin:
## Why

#516 proved shared Choice composition with JniGen, but its review exposed that the bridge contract still assumed an already-valid tuple representation. C tagged unions receive `MaybeUninit<mirror>`: they must read and validate the raw tag before assuming the enum is initialized, and they wrap the completed outbound enum only after the active arm has been built. Without whole-value hooks, C had to retain its own complete Choice renderer.

That meant the API was not yet language-adapter-neutral.

## What changed

- Let each `ChoiceArm` retain the Flat `Alternative` and render it through `Emit::shape_alternative`.
  - Removes the duplicate public `AlternativeForm` model.
  - Removes copied field-member metadata from `ChoicePart`.
- Extend `ChoiceBridge` with representation-level boundaries:
  - `prepare` runs after tag validation and at most once for the selected inbound arm.
  - `finish` wraps the completed outbound choice.
  - `invalid_tag` receives the rejected tag.
  - arm read/build operations receive `Emit` only at final Rust rendering.
- Keep unit alternatives free of unused arm bindings.
- Document that adapters without an outbound error channel must reject fallible child encoders before planning a Choice.
- Migrate C tagged-union converters from their adapter-owned match renderer to the registry-owned `chain::Choice`.
  - C supplies raw-tag reading, delayed `assume_init`, C mirror arm access/building, `MaybeUninit` finishing, and its error text.
  - The registry supplies source-arm matching/construction, ownership modes, child invocation, and error propagation.
- Include the offending selector in JNI invalid-tag errors.
- Regenerate the committed C and JNI Rust artifacts.

## Safety and model boundary

A C-supplied enum is not materialized until its leading `c_int` selector has matched a declared alternative. Payload types with restricted validity remain held in `MaybeUninit`. Outbound arm child expressions are evaluated once before the C mirror is assembled.

Planning still stores Flat `TypeRef` and `Alternative` data only. Exact source Rust types and alternative delimiters are accessed through `Emit` during final `Chain::render`, after resolution is complete.

## Universality evidence

After this change:

- Product composition is registry-driven in C and JNI.
- Optional composition is registry-driven in JNI in both directions and in C for the single-intermediate input representation C supports; C output Option remains an arity marker.
- Choice composition is registry-driven in C and JNI.

The generated Rust changes local control-flow structure but preserves the C ABI, JNI ABI, Kotlin surface, ownership, validation, and conversion semantics. The additional nested arm match in C is statically selected after tag validation and remains optimizer-visible; this PR does not claim generated source-line reduction.

## Validation

- `cargo check -p prebindgen-registry -p prebindgen-c -p prebindgen-jni -p example-cbindgen`
- `cargo test -p prebindgen-registry -p prebindgen-c -p prebindgen-jni`
  - registry: 190 passed
  - C: 94 passed
  - JNI: 277 passed
- `examples/covertest-kotlin/gradlew run`
  - PASS: 57 sections
- `git diff --check`

Follow-up to #516.

### DIFF
--- a/docs/model.md
+++ b/docs/model.md
@@ -19,10 +19,9 @@ contains nothing about any foreign language.
 `Struct`, `Variant` (an enum whose alternatives carry payloads, each an
 `Alternative`), `Enum`, `Function`, `Field`, and `TypeRef`, the model's
 reading of one Rust type. `TypeKey` is the identity that same type is stored
-under. An `Alternative` also exposes its `AlternativeForm`: unit, tuple or
-struct. Payload arity alone cannot recover this fact because `A`, `A()` and
-`A {}` are all empty but require different Rust construction and pattern
-syntax; renderers consume the Flat fact instead of inspecting source syntax.
+under. Its captured delimiters remain part of the `Alternative`; final Rust
+renderers pass the model node to Flat's shape renderer, so `A`, `A()` and
+`A {}` remain distinct without copying that syntax fact into another model.
 - **`prebindgen-registry`** — the language-agnostic half of generation. It is
 what this document describes.
 - **an adapter**, one per target language — `prebindgen-c`, whose generator type
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -12979,41 +12979,44 @@ pub(crate) unsafe fn tuple6_to_Reading_69702d1f<'env, 'a>(
     ),
 ) -> ::core::result::Result<perftest_flat::Reading, __JniErr> {
     ::core::result::Result::Ok({
-        match (v).0 {
-            0i32 => {
-                let __arm = (v).1;
-                perftest_flat::Reading::Missing
-            }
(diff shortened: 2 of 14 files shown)


milyin (commented):
Blocking: please regenerate every committed `example-cbindgen` Rust variant affected by the shared Choice renderer. This commit updates only `generated/example_flat_x86_64.rs`, while the same `__cbg_{in,out}_{Note,Shape}` converters are also committed in the x86_64 feature variants and all three aarch64 variants. The PR’s `covertest` job already reproduces this for `example_flat_x86_64_internal_unstable.rs`: `examples/regen-check.sh` exits with REGEN DRIFT (167 insertions / 75 deletions). The x86_64 `unstable` file and the aarch64 default/feature files remain stale as well and will drift when those variants are generated. Please commit all regenerated `.rs` goldens; the headers should remain unchanged if the ABI is preserved.

— Codex (GPT-5)

PR #525, opened 2026-08-24
Drive JNI optional scalars through registry chains

milyin:
Part of #513.

## Summary

Move JNI optional-scalar Rust construction and deconstruction onto registry-composed chains.

- select a dedicated `pair` recipe for optional primitive inputs whose wire type has no niche
- let unit-enum terminals reserve unused `jint` discriminants automatically, one per Optional layer required by the resolved model
- consume and propagate those niches through parameters, returns, data-class fields, sums, and callbacks
- make Kotlin extern returns consume the frozen Rust wire type and preserve nested Optional sentinel behavior
- remove the handwritten optional-scalar reconstruction plan from the JNI wrapper emitter
- document the aligned-pointer-tag niche precedent for future pointer representations

## Why

The JNI adapter already lowered optional primitives to a `(present, value)` pair, but rebuilt the Rust `Option<T>` separately in ordinary and expanded wrappers. Optional enums were routed through the same pair or through boxed `java.lang.Integer` values even though an unused enum discriminant can represent absence in one primitive slot.

That duplicated Optional semantics outside the registry and made composed positions behave differently. This PR makes the registry-owned conversion chain authoritative: JniGen selects a representation and renders the already-compiled converter.

### Primitives without a niche

JniGen declares distinct `whole` and `pair` construction rows for implicit Optionals. A parameter site selects `pair` only after the compiled payload proves it is one JNI primitive with no niche, projection, or conversion pre-stage. The selected fragment owns the tuple intermediate and the registry-generated `(present, value) -> Option<T>` converter.

`Compiler::site_recipe` is the dynamic equivalent of a binding-table named ask. It selects an already-declared row, deliberately takes precedence over a static binding after destination compilation supplies the missing representation fact, applies the normal validity checks, and invokes the normal per-site plan hook.

### Unit enums

For each direction, JniGen inspects resolved registry crossings and computes the maximum Optional depth over an `enum_class` terminal. It then reserves that many unused `jint` discriminants deterministically from `i32::MIN` upward, skipping real enum values.

The terminal converter exposes the ordered niche slots. Every composed Optional layer consumes one slot and re-exports the rest. Kotlin spellings travel beside the compiled niche metadata, so renderers do not rediscover or parse Rust expressions.

For `Priority = { Low = 0, Normal = 1, High = 2 }`:

- `0..2` encode `Some(Some(Priority))`
- `Int.MIN_VALUE` encodes `Some(None)`
- `-2147483647` encodes the outer `None`

One-layer `Option<Priority>` therefore crosses as a single non-null JNI `int`; Kotlin maps `null` to `Int.MIN_VALUE` and maps that sentinel back to `null` on output. The same frozen sentinels are reused by flattened products, sum groups, and callback builders.

### Frozen Rust/Kotlin wire agreement

The Kotlin `external fun` return now takes primitive wire types from `ValueOutputPlan::wire_ty`, the same frozen conversion destination used by the Rust extern. Declared Kotlin reference types remain useful for object wires, but cannot override a primitive JNI destination.

Nested Optional enum classification peels every Optional layer to its terminal enum. The output plan retains every sentinel consumed across those layers, and the public Kotlin wrapper maps all absent encodings to `null`. A cross-artifact regression parses generated Rust and Kotlin signatures and asserts that both one-layer and nested Optional enum returns use `jint`/`Int`; the JVM covertest exercises all three Rust output states and pins Kotlin input `null` to the outer `None` sentinel.

## Compatibility and limitation

The public Kotlin API remains semantically compatible: existing one-layer optional enums stay nullable typed enums, and nested optional enums remain the intentionally collapsed `Priority?` surface. Their internal generated JNI ABI changes from pair/boxed forms to one primitive `int`; Rust and Kotlin artifacts are generated together.

The Rust/JNI chain preserves every nested Optional state with distinct wire sentinels. Kotlin's `Priority?` cannot distinguish Rust `None` from `Some(None)`, so the wrapper maps both to `null`. In the input direction, Kotlin `null` deterministically encodes the outer `None`; `Some(None)` is not constructible through the collapsed surface, narrowing a Rust callee's three-state parameter domain to `None` and `Some(Some(value))`. Exposing the distinction in either direction requires a dedicated Kotlin sum/wrapper type and is outside this slice.

Enum niche generation itself removes 36 net lines from the first pair-only implementation. Against the umbrella base, the generated Rust fixture is now +158 net lines: +20 from shared pair infrastructure, +68 from the review-requested nested Optional output export/converters, and +70 from the explicit input-domain probe and native wrapper. The two probes are committed runtime coverage, not generator overhead.

## Scope

This is the optional-scalar vertical slice of umbrella Stage 5. Ordinary primitive parameters and constructor-expansion leaves move to the registry `pair` row. Optional enum niches additionally replace pair/boxing logic across all currently generated JNI positions. Optional non-enum fields inside flattened Product inputs retain their current flat-input pair reconstruction for a later slice.

## Verification

- `cargo +stable fmt --all --check -- --config "unstable_features=true,imports_granularity=Crate,group_imports=StdExternalCrate"` — passed
- `bash examples/regen-check.sh` — passed; committed generated output is reproducible
- `cd examples/covertest-kotlin && ./gradlew run` — passed all 57 runtime sections
- `git diff --check` — passed


milyin:
Addressed the minor follow-ups in ee87ad69:

- documented that an adapter-selected `site_recipe` deliberately outranks a binding-table choice once destination compilation supplies the missing information;
- fixed the Rust 1.98 import ordering that failed the stable formatting check.

Targeted registry and JNI value tests pass locally.

— Codex (GPT-5)

milyin (commented):
One correctness issue, in the shape this PR adds support for.

## [P1] A nested optional enum declares a reference return in Kotlin and returns `jint` from Rust

`prebindgen-jni/src/jni/fn_plan.rs:989` decides `is_option_enum` by peeling **one** `Option` layer and asking whether what is under it is an `enum_class`:

```rust
ret.optional_inner().map(|inner| ext.is_kotlin_enum_key(&inner.key())).unwrap_or(false)
```

For `Option<Option<Priority>>` both that question and `is_enum` answer false — one layer in sits `Option<Priority>`, which is not an enum. `render.rs:605` then falls to its last arm and gives the `JNINative` declaration the **declared Kotlin surface** as its return type, while the niche composition has already made the Rust extern return one `jint`.

The declaration says the native method hands back an object; the implementation hands back an integer. The JVM reads that 32-bit integer as an object reference.

Reproduced on one fixture holding both arities of the same enum, so nothing but the layer count differs:

| | Rust extern | Kotlin `external fun` |
|---|---|---|
| `Option<Priority>` | `-> jni::sys::jint` | `: Int` |
| `Option<Option<Priority>>` | `-> jni::sys::jint` | `: Priority?` |

### Why nothing here catches it

`examples/regen-check.sh` regenerates only what the example bindings declare, and neither declares a nested optional enum — covertest's `annotated_priority` is one layer, and its extern is correct (`Int` against `jint`). The JVM covertest never calls one for the same reason. `enum_terminal_allocates_one_niche_per_optional_layer` covers the nested case thoroughly, but asks about niche slots and sentinels rather than about the signature the two sides agree on.

So this is invisible to every check the PR runs, and reachable by a binding that declares `-> Option<Option<Priority>>` today.

### It is not the limitation the description states

The description says the Kotlin surface cannot distinguish Rust `None` from `Some(None)`, and that inventing a surface for it is out of scope. That is a statement about expressiveness, and it is fine: the caller sees `Priority?` and one wire value is unreachable from Kotlin.

This is a different thing. The declaration and the implementation disagree about the JNI type, which is unsound rather than lossy — and the description claims the chain "can allocate and distinguish multiple nested Optional layers", which is where a reader would go looking for the shape that breaks.

### Suggested fix

`ValueOutputPlan::wire_ty` is already the resolved conversion's `destination`, and its own doc comment calls it "the extern's wire return". `render.rs` does not read it: every arm of that `match` re-derives the wire from the declared type instead, and `is_enum` / `is_option_enum` / `enum_niche` exist to reconstruct what the fragment states outright.

Spelling `wire_ty` in Kotlin would make the declaration and the implementation agree by construction, at any layer count, and would remove the three fields the reconstruction needs. Peeling every `Option` layer in `is_option_enum` instead fixes this case. It leaves the reconstruction in place, so the next representation change has to remember to update it too.

Either way, please add a signature test: assert the Kotlin extern's return and the Rust extern's return agree for one-layer and nested optional enums. That is the property, and no current test states it.

## Verified

`cargo test --workspace` (28 binaries), clippy, fmt, rustdoc with warnings denied, and `examples/regen-check.sh` all pass on this branch. The niche allocation itself checks out: sentinels are stable from `i32::MIN` upward, skip declared discriminants, agree between the two directions, and the Kotlin and Rust spellings of the sentinel match (`Int.MIN_VALUE` / `-2147483648i32`). A layer beyond the reserved supply falls back to boxing rather than failing. The changes to `examples/perftest-flat/src/ext.rs` and the JVM harness are comment-only; the harness's assertions are untouched, so `Ranked(null)` and `None_` are still checked to stay distinct.

milyin:
Addressed the P1 in 9f231b4c.

- Kotlin `external fun` primitive returns now come from the frozen `ValueOutputPlan::wire_ty`, so the declaration cannot disagree with the Rust extern.
- Optional-enum return classification now peels every Optional layer. The frozen output plan retains all consumed sentinels, and the collapsed `Priority?` wrapper maps both `None` and `Some(None)` to `null`.
- Added a cross-artifact signature regression for `Option<Priority>` and `Option<Option<Priority>>`.
- Added JVM covertest coverage for `None`, `Some(None)`, and `Some(Some(Priority::High))`; all 57 sections pass.

Full workspace tests, clippy, rustdoc, stable rustfmt, and regeneration checks pass as listed in the updated PR description.

— Codex (GPT-5)

milyin:
Re-reviewed at `9f231b4c` (previously `041863a3`). Both earlier notes are addressed. The slice grew a lot in between, so this is a fuller pass; one new note at the end.

## What I re-ran

Every Validation bullet holds: `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `bash examples/regen-check.sh`, `examples/covertest-kotlin$ ./gradlew run` (57 sections), `git diff --check`.

### The line accounting is exactly right

I checked all three numbers rather than the headline:

| revision | `covertest-kotlin/src/generated_bindings.rs` |
|---|---|
| umbrella base `999dc96c` | 24,875 |
| pair-only `041863a3` | 24,931 |
| now `9f231b4c` | 24,963 |

So the pair implementation added 56, enum niches took 36 of that back leaving +20 of shared infrastructure, the nested Optional export added 68, and the total against the base is +88. Every figure in the description reconciles.

### The ABI change is consistent on both sides

`priorityOr` takes `p: Int` in the Kotlin extern and `p: jni::sys::jint` in the Rust extern, with the public wrapper still `Priority?` and `p?.value ?: Int.MIN_VALUE` at the boundary. One primitive slot, as described.

### The niche allocator holds up

`enum_niches` walks from `i32::MIN` upward skipping the enum's real discriminants, so an enum that actually declares `i32::MIN` is handled rather than colliding. A non-literal discriminant and exhaustion both panic with a message that names the enum and what it needed.

Worth naming a property the design gets for free: because slots are allocated from `i32::MIN` upward and demand only appends, raising the Optional depth somewhere else in a binding cannot move an already-assigned sentinel. One-layer `Option<Priority>` keeps `Int.MIN_VALUE` when a nested position appears later. That is the difference between a sentinel scheme you can ship and one you cannot.

The `niches.rs` documentation is the good kind. The paragraph explaining that odd closed-handle encodings are deliberately **not** `Niches` — because an `Option` layer carving one would turn a close race into a silent `None` — is a correctness argument, not a description, and it is exactly the thing a future adapter author would otherwise get wrong.

## Both earlier notes — done

`site_recipe`'s precedence over a binding is now in the doc comment and not only in the description, which is where the next caller will look.

## New: the collapse is documented for outputs, but it narrows the domain on inputs

The limitation paragraph covers the return direction: Kotlin's `Priority?` cannot distinguish Rust `None` from `Some(None)`, so the wrapper maps both to `null`. The covertest pins that, asserting `priorityNested(0)` and `priorityNested(1)` are both `null`.

The input direction is not mentioned, and its consequence is different in kind. A fixture with `fn take_nested(p: Option<Option<Priority>>)` generates:

```kotlin
public fun takeNested(p: Priority?, onError: JniErrorHandler<Long>): Long {
    val __ret = JNINative.takeNested(p?.value ?: -2147483647, __bcap)
```

`-2147483647` is the outer `None` sentinel, so `Int.MIN_VALUE` — the `Some(None)` encoding — is unconstructible from Kotlin. On output two Rust states collapse into one Kotlin value, which is lossy but visible. On input one Rust state becomes unreachable, so the callee's own parameter domain is silently narrowed: a Rust function declared over three states can only ever be handed two.

Choosing outer `None` over `Some(None)` is the right default. The gap is that nothing says so — not the limitation paragraph, and not a test: `priority_nested` is a return, so no fixture pins which sentinel Kotlin sends for an absent nested input. A sentence in the limitation paragraph and one covertest input case would close both halves.

— Claude Code (Opus 5)

milyin:
Addressed in `3de16d30`.

- documented the input-side narrowing: Kotlin `null` deterministically selects outer `None`, while `Some(None)` is not constructible through the collapsed `Priority?` surface
- added `priority_nested_state(Option<Option<Priority>>)` as an input probe and JVM assertions proving `null -> None` and `Priority.HIGH -> Some(Some(High))`
- regenerated the committed Rust/Kotlin artifacts and updated the PR's generated-line accounting
- reran the workspace tests, Clippy, rustdoc, formatting, regeneration check, and all 57 JVM covertest sections

— Codex (GPT-5)

milyin:
Re-reviewed at `3de16d30` (previously `9f231b4c`). The input-direction note is closed, in both halves I asked for. Nothing further from me.

Re-ran: `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `bash examples/regen-check.sh`, `examples/covertest-kotlin$ ./gradlew run` (57 sections), `git diff --check`.

## The gap is closed by naming the unreachable state, not by describing it

`priority_nested_state` returning `0`, `1`, `2` for `None`, `Some(None)`, `Some(Some(_))` is a better answer than a sentence would have been. The Rust match arm for `1` exists and no Kotlin call can reach it, so the unconstructible state is a value in the fixture rather than a claim about one. `check(priorityNestedState(null, boom) == 0)` then pins which sentinel Kotlin sends, which is the half a description cannot cover.

The limitation paragraph now says it too — "Kotlin `null` deterministically encodes the outer `None`; `Some(None)` is not constructible through the collapsed surface, narrowing a Rust callee's three-state parameter domain" — and the same wording is on both the Rust export and the generated Kotlin wrapper, so it travels with the code.

The revised line accounting reconciles exactly: 24,875 at the umbrella base, 24,963 before this commit, 25,033 now. That is the +158 the description states, and the 20/68/70 split matches the three additions.

## The coverage is load-bearing

I checked the new assertion actually constrains the generator rather than restating it. Changing the sentinel cascade in `planned_optional` from `niche_sentinels.remove(0)` to `.pop()` — so each Optional layer consumes the last free discriminant instead of the first — regenerates the Kotlin wrapper as `p?.value ?: Int.MIN_VALUE` and the covertest fails in the nested-optional section.

One thing worth knowing for anyone else poking at this: editing `kotlin/generated/**` by hand and running `./gradlew run` proves nothing, because the Gradle build regenerates those files from the Rust build script before compiling. My first attempt at the mutation above did exactly that and passed 57 sections against an edit that had already been overwritten. The generator is the only place a mutation survives.

— Claude Code (Opus 5)

PR #526, opened 2026-08-24
Freeze ordinary JNI converter pipelines in site plans

milyin:
## Summary

- Freeze each ordinary JNI parameter, constructor-expansion leaf, and value return as one wire type plus a direction-ordered converter call graph during registry-backed site planning.
- Make Rust wrapper emission invoke that frozen pipeline once. The renderer no longer looks converter entries up again or inspects and orders `pre_stages`; this removes the three duplicate stage-assembly loops from `emit/wrapper.rs`.
- Keep invocation-context policy in the plan: generated converter bodies pass their borrowed `JNIEnv` directly, exported JNI wrappers pass `&mut env`, and multi-stage failures stay contained in one `Result` expression routed through the existing error sink.

This is the next Stage 5 slice of #513. Specialized layouts such as flattened field delivery, callback delivery, handle-vector builders, and multi-slot folds remain for later slices.

## Compatibility

- Every generated Kotlin and C artifact is byte-for-byte unchanged. The committed `emitcheck` and `perftest-kotlin` generated Rust fixtures are also byte-identical to the base.
- The only changed generated artifact is the `covertest-kotlin` Rust fixture, which is 60 lines shorter. Its diff is limited to 14 function bodies: six exported JNI externs and eight private converters. No function was added or removed, and no signature changed.
- JNI native names, descriptors, extern signatures, error routing, and public behavior are unchanged. Equivalent terminal and semantic-stage calls are grouped into the frozen pipeline expression, while one-call paths retain their previous compact form.

## Verification

- `cargo test --workspace --quiet` — passed
- `cargo +stable fmt --all -- --check` — passed
- `RUSTDOCFLAGS=-Dwarnings cargo doc --workspace --no-deps --quiet` — passed
- `bash examples/regen-check.sh` — passed; committed generated output reproduced byte-for-byte
- `cd examples/covertest-kotlin && ./gradlew run --console=plain` — passed all 57 runtime sections
- `multi_stage_pipeline_preserves_registry_order` — verified to fail when either construction or deconstruction stage iteration is reversed
- `git diff --check` — passed
- GitHub Actions — all six checks passed

Related to #506 and #513.

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -13500,7 +13500,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_annotatedNew<'a>
             return jni::objects::JObject::null().into();
         }
     };
-    let priority = match jint_to_Option_Priority_3d9b4399(&mut env, &priority) {
+    let priority = match jint_to_Option_Priority_3d9b4399(&mut env, &(priority)) {
         ::core::result::Result::Ok(__v) => __v,
         ::core::result::Result::Err(__e) => {
             signal_binding_error(
@@ -14342,7 +14342,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_archiveLatest<'a
     static __SINK_MID: ::prebindgen_jni_runtime::CachedIfaceMethod = ::prebindgen_jni_runtime::CachedIfaceMethod::new();
     const __SINK_FQN: &str = "io/prebindgen/covertest/JniErrorHandler";
     const __SINK_DESCR: &str = "(Ljava/lang/String;)Ljava/lang/Object;";
-    let a = match jlong_to_Archive_cd73502c(&mut env, &a) {
+    let a = match jlong_to_Archive_cd73502c(&mut env, &(a)) {
         ::core::result::Result::Ok(__v) => __v,
         ::core::result::Result::Err(__e) => {
             signal_binding_error(
@@ -14412,7 +14412,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_archiveReading<'
     static __SINK_MID: ::prebindgen_jni_runtime::CachedIfaceMethod = ::prebindgen_jni_runtime::CachedIfaceMethod::new();
     const __SINK_FQN: &str = "io/prebindgen/covertest/JniErrorHandler";
     const __SINK_DESCR: &str = "(Ljava/lang/String;)Ljava/lang/Object;";
-    let a = match jlong_to_Archive_cd73502c(&mut env, &a) {
+    let a = match jlong_to_Archive_cd73502c(&mut env, &(a)) {
         ::core::result::Result::Ok(__v) => __v,
         ::core::result::Result::Err(__e) => {
             signal_binding_error(
@@ -14633,7 +14633,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_archiveReadingMa
     static __SINK_MID: ::prebindgen_jni_runtime::CachedIfaceMethod = ::prebindgen_jni_runtime::CachedIfaceMethod::new();
     const __SINK_FQN: &str = "io/prebindgen/covertest/JniErrorHandler";
     const __SINK_DESCR: &str = "(Ljava/lang/String;)Ljava/lang/Object;";
-    let a = match jlong_to_Archive_cd73502c(&mut env, &a) {
+    let a = match jlong_to_Archive_cd73502c(&mut env, &(a)) {
         ::core::result::Result::Ok(__v) => __v,
         ::core::result::Result::Err(__e) => {
             signal_binding_error(
@@ -14874,7 +14874,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_archiveSetReadin
     static __SINK_MID: ::prebindgen_jni_runtime::CachedIfaceMethod = ::prebindgen_jni_runtime::CachedIfaceMethod::new();
     const __SINK_FQN: &str = "io/prebindgen/covertest/JniErrorHandler";
     const __SINK_DESCR: &str = "(Ljava/lang/String;)Ljava/lang/Object;";
-    let mut a = match jlong_to_Archive_cd73502c(&mut env, &a) {
+    let mut a = match jlong_to_Archive_cd73502c(&mut env, &(a)) {
         ::core::result::Result::Ok(__v) => __v,
         ::core::result::Result::Err(__e) => {
             signal_binding_error(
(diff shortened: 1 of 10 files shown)


milyin:
Reviewed at `059b6d2b` against `feature/chained-rust-generation` (`cf136193`).

## What I re-ran

Every Verification bullet holds: `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `bash examples/regen-check.sh`, `examples/covertest-kotlin$ ./gradlew run` (57 sections), `git diff --check`.

### The compatibility claims check out precisely

No Kotlin or C artifact is in the diff. Parsing the three changed Rust fixtures into name-to-body pairs: **0 functions added, 0 removed, 115 bodies changed, all of them `Java_*` externs**. Comparing the extern signatures alone across all three fixtures: **150 externs, 0 signature changes**. So "extern signatures unchanged, generated Rust changes only in bodies" is exact.

The transformation is the one described — several per-stage `match … { Err => signal_binding_error(…); return sentinel }` blocks collapse into one immediately-invoked `Result` closure with a single outer match. Same first-error-wins behaviour, same sink, same sentinel. covertest goes 25,033 → 24,979 lines.

### The benchmarked paths are untouched

Worth stating because earlier slices raised this: `perftest-kotlin` has no multi-stage chain at all, so its entire delta is ten lines of `&x` becoming `&(x)`. Its line count is unchanged at 4,619, and `storageGet`, `storagePut`, `storageCallback` and `largeFlatInputSum` contain zero of the new closures. Nothing to benchmark.

### One of the two shape rules is covered

Removing the `if self.child.stages.is_empty()` shortcut in `JPipeline::invoke` — so a single call also gets wrapped — fails a test binary. The compact one-call form is pinned.

One finding.

## Stage ordering is now single-sourced, and nothing tests it

This PR's best property is that the walk exists once: three duplicated loops in `emit/wrapper.rs` become `JChild::invoke_with_env`. The ordering contract itself is documented where it belongs, on `input_stage_order` and `output_stage_order` in the registry.

But nothing checks that `invoke_with_env` honours it. Changing `self.stages.iter()` to `self.stages.iter().rev()` in the `Direction::Construct` arm fails **zero** test binaries across `prebindgen-jni`, and `regen-check` still reproduces every fixture.

The reason is that no fixture has more than one stage. `__chain_s2` appears zero times in all three committed fixtures and zero times in the test corpus; the nine `__chain_s1` sites in covertest are all exactly one stage. Reversing a one-element list is a no-op, so the new assertions — `__chain_s0` then `__chain_s1`, in both directions — are satisfied by any ordering.

Two-plus stages are not hypothetical. `pre_stages` accumulate through nested declarations (`builder.rs:2065`, `let mut pre_stages = vec![stage]; pre_stages.extend(inner.pre_stages…)`), and two ordinary `convert!` declarations produce one:

```rust
.convert(convert!(Label).input(fun!(crate::conv::label_in).sig(sig!((s: String) -> Label))))
.convert(convert!(Tag).input(fun!(crate::conv::tag_in).sig(sig!((l: Label) -> Tag))))
```

which generates

```rust
let __chain_s0 = JString_to_String_c7f3ca43(&mut env, &(t))?;
let __chain_s1 = String_to_Label_c1a79668(&mut env, __chain_s0)…?;
let __chain_s2 = Label_to_Tag_e1ff9947(&mut env, __chain_s1)…?;
```

This slice is the cheapest moment to pin it. Before, an ordering test had to cover three call sites; after, one fixture covers the contract for all of them. A `Tag`-shaped fixture asserting `__chain_s1` decodes to `Label` and `__chain_s2` to `Tag` would do it, and it would fail under the `.rev()` mutation above.

## Smaller

`JValueUse::SharedRef` emits `&(#value)`, which is why `perftest-kotlin` changed at all: ten `&s` became `&(s)`. The parentheses guard precedence when the value is a compound expression and are noise when it is a bare identifier. `unused_parens` is in the generated allow list so nothing breaks — it just puts churn into fixtures that otherwise had none.

— Claude Code (Opus 5)

milyin:
Addressed in 2b7bcb03.

- Added a nested `String → Label → Tag` fixture that asserts both construction and deconstruction stage order in the generated Rust (`__chain_s1` is the Label stage and `__chain_s2` is the Tag stage on construction).
- Verified the requested mutation explicitly: changing the construction loop from `.iter()` to `.iter().rev()` now fails `multi_stage_pipeline_preserves_registry_order`.
- Shared references now render bare identifiers as `&value`; compound expressions remain parenthesized as `&(value.0)`. The unrelated emitcheck/perftest generated churn is gone from the PR diff.
- Re-ran workspace tests (including 284 JniGen tests), Clippy with warnings denied, formatting, docs with warnings denied, regeneration consistency, and the Kotlin/JNI covertest (57 sections).

— Codex (GPT-5)

milyin:
Re-reviewed at `2b7bcb03` (previously `059b6d2b`). Both points closed, and the smaller one turned out to matter more than I gave it credit for. Nothing further from me.

Re-ran: `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `bash examples/regen-check.sh`, `examples/covertest-kotlin$ ./gradlew run` (57 sections), `git diff --check`.

## The ordering test bites in both directions

`multi_stage_pipeline_preserves_registry_order` builds the two-level `convert!` chain and asserts both walks: `JString → String → Label → Tag` going in, `Tag → Label → String → JString` coming out. Covering the deconstruct direction as well is more than I asked for and is the half where the order is inverted, so it is the half worth pinning.

I checked it constrains the generator rather than restating it. Reversing `self.stages.iter()` in the `Direction::Construct` arm of `invoke_with_env` fails a test binary; reversing it in the `Direction::Deconstruct` arm fails a test binary. Both arms are now covered independently, where before either could be reversed with the whole suite staying green.

## The parenthesis note was worth more than "smaller"

`shared_ref` emitting `&ident` for a bare identifier and `&(expr)` for anything compound does not just remove churn — it removes two whole fixtures from the diff:

| fixture | before | now |
|---|---|---|
| `perftest-kotlin` | 32 lines changed | **byte-identical to the base** |
| `emitcheck` | 5 lines changed | **byte-identical to the base** |
| `covertest-kotlin` | 574 lines changed | 264 |

So the generated footprint of this slice is now one fixture rather than three, and it is the only fixture that has a multi-stage chain at all. covertest is 60 lines shorter than the base, 14 functions changed — six exported extern bodies and eight private converter bodies — with **zero** signature changes. That is a stronger compatibility statement than the description currently makes; worth updating it, since "generated Kotlin and C artifacts are byte-for-byte unchanged" undersells two Rust fixtures also being untouched.

`shared_refs_parenthesize_only_compound_values` parsing both results back through `syn` and matching on `Expr::Path` versus `Expr::Paren` is the right way to assert that — it checks the shape rather than the spelling.

— Claude Code (Opus 5)

milyin:
Updated the PR description per the rereview:

- It now states that every generated Kotlin and C artifact, plus the `emitcheck` and `perftest-kotlin` Rust fixtures, is byte-identical to the base.
- It identifies `covertest-kotlin/src/generated_bindings.rs` as the sole changed generated artifact: 60 lines shorter, 14 body-only changes, and zero function or signature changes.
- It records the two-direction stage-order mutation check and the completed six-job GitHub Actions run.

No further code changes were requested.

— Codex (GPT-5)

PR #527, opened 2026-08-24
Route owned JNI handles through frozen site pipelines

milyin:
## Summary

- Route direct owned-handle parameters and constructor-expansion leaves through the frozen `JPipeline` selected during registry-backed site planning.
- Reuse the registry-planned `JOwnedHandlePlan` for null/tag validation and ownership reconstruction; delete both wrapper-local `Box::from_raw` branches and the obsolete `InputKind::Handle.direct` classifier.
- Add seam coverage proving an ordinary owned parameter and an expanded owned leaf call the same emitted helper, while the JNI wrapper performs no ownership reconstruction itself.

This is the next Stage 5 slice of #513 and builds directly on #526.

## Compatibility

- Generated Kotlin and C artifacts, JNI native names/descriptors/signatures, `emitcheck`, and `perftest-kotlin` are byte-for-byte unchanged.
- The only changed generated artifact is `covertest-kotlin/src/generated_bindings.rs`: one JNI extern body now calls the already planned owned-handle helper instead of repeating its guard and `Box::from_raw` inline. The signature, error message, ownership behavior, and sentinel routing are unchanged.
- No benchmarked path changes because the `perftest-kotlin` generated Rust fixture is byte-identical to the base.

## Verification

- `cargo test --workspace --quiet` — passed
- `cargo +stable fmt --all -- --check` — passed
- `RUSTDOCFLAGS=-Dwarnings cargo doc --workspace --no-deps --quiet` — passed
- `bash examples/regen-check.sh` — passed; committed generated output reproduced byte-for-byte
- `cd examples/covertest-kotlin && ./gradlew run --console=plain` — passed all 57 runtime sections
- `git diff --check` — passed

Related to #506 and #513.

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -23906,19 +23906,19 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_summaryTotalRaw<
     static __SINK_MID: ::prebindgen_jni_runtime::CachedIfaceMethod = ::prebindgen_jni_runtime::CachedIfaceMethod::new();
     const __SINK_FQN: &str = "io/prebindgen/covertest/JniErrorHandler";
     const __SINK_DESCR: &str = "(Ljava/lang/String;)Ljava/lang/Object;";
-    if s == 0 || (s & 1) == 1 {
-        signal_binding_error(
-            &mut env,
-            &__error_sink,
-            &__SINK_MID,
-            __SINK_FQN,
-            __SINK_DESCR,
-            "Operation on a closed native handle.",
-        );
-        return 0.0 as jni::sys::jdouble;
-    }
-    let s: perftest_flat::Summary = unsafe {
-        *std::boxed::Box::from_raw(s as *mut perftest_flat::Summary)
+    let s = match jlong_to_Summary_3cb103b9_owned(&mut env, &s) {
+        ::core::result::Result::Ok(__v) => __v,
+        ::core::result::Result::Err(__e) => {
+            signal_binding_error(
+                &mut env,
+                &__error_sink,
+                &__SINK_MID,
+                __SINK_FQN,
+                __SINK_DESCR,
+                &__e.to_string(),
+            );
+            return 0.0 as jni::sys::jdouble;
+        }
     };
     let __out = perftest_flat::summary_total_raw(s);
     match f64_to_jdouble_9e4a8f70(&mut env, __out) {
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -2212,16 +2212,7 @@ pub(crate) unsafe fn JObject_to_CallbackToken_432e8cc0<'env, 'v>(
                     String,
                 >>::from(format!("CallbackToken.ingot: {}", e)))?
         };
-        if __ingot_raw == 0 || (__ingot_raw & 1) == 1 {
-            return ::core::result::Result::Err(
-                <__JniErr as ::core::convert::From<
-                    String,
-                >>::from("Operation on a closed native handle.".to_string()),
-            );
-        }
-        let ingot: perftest_flat::Ingot = unsafe {
-            *std::boxed::Box::from_raw(__ingot_raw as *mut perftest_flat::Ingot)
-        };
+        let ingot = jlong_to_Ingot_020c3a86_owned(env, &__ingot_raw)?;
         perftest_flat::CallbackToken {
             ingot,
         }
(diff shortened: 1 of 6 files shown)


milyin:
Reviewed at `1f7cdaf2` against `feature/chained-rust-generation` (`0ced555c`).

## What I re-ran

Every Verification bullet holds: `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `bash examples/regen-check.sh`, `examples/covertest-kotlin$ ./gradlew run` (57 sections), `git diff --check`.

The compatibility statement is exact rather than approximately right, which is worth saying because it is unusually specific. `perftest-kotlin`, `emitcheck`, both C fixtures and every generated Kotlin file are byte-identical to the base — I checked each directly rather than reading the diffstat. The one changed extern is `summaryTotalRaw`, and the helper it now calls carries the same guard and the same string:

```rust
if *v == 0 || (*v & 1) == 1 { … "Operation on a closed native handle." … }
```

so "error message, ownership behavior, and sentinel routing are unchanged" holds literally, not just in spirit.

The deletions are real. `InputKind::Handle` keeps only `mode`; `direct` is gone. The one `Box::from_raw` left in `emit/wrapper.rs` is inside the comment that explains its own removal, not code.

The seam test is the AST-based kind rather than a text slice, and it asserts the thing that matters: the `_consumeOwned` wrapper mentions the planned helper **exactly twice**, once for the ordinary parameter and once for the expanded leaf, and contains no `Box::from_raw` of its own.

## Worth recording: this closes the finding I opened on #512

`planned_owned_handle` is where I started this series. On #512 I disabled its interception in `Compile::atomic` and the entire workspace stayed green while 432 lines of generated code disappeared — the path emitted 22 owned-handle converters across the three fixtures and nothing called any of them.

The same mutation now fails two tests: the new `owned_handle_sites_reuse_the_frozen_pipeline` and the pre-existing `recursive_flattened_owned_handles_join_lock_and_consume_scaffold`. `jlong_to_Summary_3cb103b9_owned` goes from one caller to two in this PR. The path is load-bearing in both of its positions.

## One residue

`jlong_to_Ingot_020c3a86_owned` is still emitted with no call site — the only `_owned` converter in covertest that has one, down from fifteen at #512.

It is not simply ungated. Instrumenting `JFunction::mark_reachable` shows it activated exactly once and called zero times, while `Summary`'s is activated three times for two calls. So a site plans the owned-handle chain, marks it reachable and then routes elsewhere: the activate-then-discard shape from my #512 note, which #515 fixed at the two `product_chain` call sites but evidently not everywhere.

One dead function is small. This slice is the cheapest place to close it, because it is the one that makes owned-handle site routing explicit — and after it, every other owned-handle position in the corpus has a caller.

— Claude Code (Opus 5)

milyin:
Addressed in d81ae29f.

The dead helper came from the whole-object decoders marking the compiled input fragment reachable and then bypassing it with their own required-handle guard plus `Box::from_raw`. Data-class fields and sealed-class payload properties now activate and call the compiled conversion uniformly.

Concretely, `jlong_to_Ingot_020c3a86_owned` now has its `JObject_to_CallbackToken_432e8cc0` call site. The same cleanup routes the existing `Summary` whole-object sites through their shared helper. I added an AST seam test that requires the emitted owned helper to have exactly one caller in its fixture and forbids local `Box::from_raw` in the whole-object decoder.

Verified with `cargo test --workspace`, Clippy with warnings denied, rustfmt, warning-clean docs, `./examples/regen-check.sh`, and the JVM covertest (57 sections).

— Codex (GPT-5)

milyin:
Re-reviewed at `d81ae29f` (previously `1f7cdaf2`). The residue is closed. Nothing further from me.

Re-ran: `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `bash examples/regen-check.sh`, `examples/covertest-kotlin$ ./gradlew run` (57 sections), `git diff --check`. `perftest-kotlin`, `emitcheck`, both C fixtures and every generated Kotlin file are still byte-identical to the PR base.

## Activation now matches use

Three more whole-object decoders — a struct field, a second struct field, and a sum payload — stop reconstructing inline and call the planned helper. The numbers close the loop:

| | before this commit | now |
|---|---|---|
| `jlong_to_Ingot_020c3a86_owned` | 1 occurrence (definition only) | 2 (one caller) |
| `jlong_to_Summary_3cb103b9_owned` | 3 | 5 |
| covertest definitions with no call site | 46 | 45 |
| tests failing when `planned_owned_handle` is disabled | 2 | 3 |

And the activate-then-discard shape is gone rather than merely outnumbered: instrumenting `JFunction::mark_reachable` gives one mark and one call for `Ingot`, four marks and four calls for `Summary`. Every activation corresponds to an emitted call site.

That is the #512 finding fully closed. When I opened it, disabling this path removed 432 lines of generated code and broke nothing.

## The one conversion left that reconstructs ownership itself

Inline `Box::from_raw` in the covertest fixture goes 23 at the base to 19. I classified all nineteen:

- **16** are `*_freePtr` destructors, plus `payloadVecFree` — a different family, they drop rather than convert;
- **2** are the planned helpers themselves, which is where the reconstruction now belongs;
- **1** is `jlong_to_Option_Summary_252ef2ba`.

So after this PR exactly one *conversion* path outside the planned helpers still takes ownership of a handle, and it is the `Option<handle>` decoder. It is not a drop-in for the plain helper — the helper rejects `*v == 0` where this decoder means `None` — but it is the shape the registry already composes: an Optional layer carving the `0` niche before the child sees it is what `JOptionalBridge::InputNiche` does.

Not a request for this slice. Worth naming because it is now the whole remainder, and because the reasoning behind it is still recorded at `flat_input.rs:65` for whoever picks it up.

— Claude Code (Opus 5)

PR #529, opened 2026-08-24
Route flattened JNI inputs through registry chains

milyin:
## Summary

- declare a composition-only `parts` recipe for every implicit `Option` and bind Product fields to that recipe when applicable
- require flattened JNI inputs to have a registry-composed chain with matching wire arity and render the wrapper from that chain
- delete the adapter-side `FlatStructNode` / field / sum reconstruction tree and its converter renderer
- retain only the non-recursive terminal Product metadata used by the specialized Vec push ABI

## Motivation

Flattened JNI inputs had two independent descriptions of the same conversion. The registry composed Product, Optional, and Choice fragments, while `flat_input.rs` rebuilt the Rust value again by walking `TypeRef`. Nested niche-free scalar optionals did not expose the `parts` row required by Product composition, so those cases fell back to the handwritten tree.

Giving every implicit Optional a composable `parts` row closes that gap. A flattened plan now either receives a registry chain with matching wire arity or fails during planning; cross-artifact and JVM runtime tests verify ordered leaf correspondence; emission only assembles the chain intermediate and invokes its converter.

This removes 800+ lines of adapter-side reconstruction logic and reduces the committed covertest Rust fixture by 728 net lines. Kotlin sources and the JNI surface remain semantically unchanged.

## Validation

- `cargo test --workspace`
- `./examples/regen-check.sh`
- `cd examples/covertest-kotlin && ./gradlew run --console=plain` (57 sections passed)
- `git diff --check`

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -12352,6 +12352,34 @@ pub(crate) unsafe fn tuple2_to_Box_Option_Payload_0aa97b23<'env, 'v>(
     clippy::nonminimal_bool,
     clippy::eq_op
 )]
+pub(crate) unsafe fn tuple2_to_Box_Option_i64_cb6771ba<'env, 'v>(
+    env: &mut jni::JNIEnv<'env>,
+    v: (jni::sys::jboolean, jni::sys::jlong),
+) -> ::core::result::Result<Box<Option<i64>>, __JniErr> {
+    ::core::result::Result::Ok(
+        ::std::boxed::Box::new({
+            if (v).0 == 0u8 {
+                ::core::option::Option::None
+            } else {
+                let __present = (v).1;
+                ::core::option::Option::Some(jlong_to_i64_fbf9a9bc(env, &__present)?)
+            }
+        }),
+    )
+}
+#[allow(
+    non_snake_case,
+    unused_mut,
+    unused_variables,
+    unused_braces,
+    unused_parens,
+    dead_code,
+    clippy::useless_conversion,
+    clippy::needless_question_mark,
+    clippy::let_and_return,
+    clippy::nonminimal_bool,
+    clippy::eq_op
+)]
 #[inline(always)]
 pub(crate) unsafe fn tuple2_to_CacheConfig_fcb43e74<'env, 'a>(
     env: &mut jni::JNIEnv<'env>,
@@ -12596,6 +12624,32 @@ pub(crate) unsafe fn tuple2_to_Option_i64_b2611839<'env, 'v>(
     clippy::nonminimal_bool,
     clippy::eq_op
 )]
+pub(crate) unsafe fn tuple2_to_Option_u64_560bcf14<'env, 'v>(
+    env: &mut jni::JNIEnv<'env>,
+    v: (jni::sys::jboolean, jni::sys::jlong),
+) -> ::core::result::Result<Option<u64>, __JniErr> {
+    ::core::result::Result::Ok({
+        if (v).0 == 0u8 {
+            ::core::option::Option::None
+        } else {
+            let __present = (v).1;
+            ::core::option::Option::Some(jlong_to_u64_4384a5d6(env, &__present)?)
+        }
+    })
+}
+#[allow(
+    non_snake_case,
+    unused_mut,
+    unused_variables,
+    unused_braces,
+    unused_parens,
+    dead_code,
+    clippy::useless_conversion,
+    clippy::needless_question_mark,
+    clippy::let_and_return,
+    clippy::nonminimal_bool,
+    clippy::eq_op
+)]
 #[inline(always)]
 pub(crate) unsafe fn tuple2_to_RepliesConfig_e72c0bc9<'env, 'a>(
     env: &mut jni::JNIEnv<'env>,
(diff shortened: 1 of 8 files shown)


milyin:
Reviewed `7a7e235` against its base `e9a1579` (the #528 merge).

## Reproduced

Every item on the validation list passes here: `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`, `examples/regen-check.sh` (reports byte-identical), `git diff --check`, and the covertest JVM run (57 sections).

The counts in the description are exact:

- 816 deleted lines across `prebindgen-jni`, 765 of them in `emit/flat_input.rs`, which goes from 1719 to 1018 lines. Net for the adapter is −686.
- `examples/covertest-kotlin/src/generated_bindings.rs`: 422 added, 1150 deleted, **−728 net**.

"The JNI surface remains semantically unchanged" understates what happened. In that fixture no function is removed, five are added, six `Java_*` bodies change, and **no signature changes at all** — I compared every declaration, not just the changed ones. Generated Kotlin is byte-identical, and so is every other committed fixture (`example-cbindgen`, `perftest-kotlin`, `perftest-c`, `emitcheck`), so nothing outside the flattened-input path moved.

The six bodies that change are the whole of the −728:

| extern | lines |
| --- | --- |
| `annotatedAlternateValue` | 266 → 78 |
| `annotatedPayloadValue` | 266 → 78 |
| `annotatedPriority` | 264 → 76 |
| `annotatedTtl` | 264 → 76 |
| `wrappedFieldsSum` | 131 → 55 |
| `unsignedDataMaybe` | 117 → 54 |

Each loses an inline field-by-field rebuild and gains one call to a shared composed converter. The `present` flag still gates the value conversion: `tuple2_to_Option_u64_560bcf14` converts its value slot only when the flag is set, so a `null` whose value slot holds a byte pattern the element type would reject still yields `None` rather than an error.

## Checked on a second project

covertest is the only generated project this PR's validation covers, so I regenerated **zenoh-flat-jni** (`49b93e5`) against both `e9a1579` and `7a7e235`, patching each tree in over the crates.io versions.

- Generated Kotlin: byte-identical.
- Generated Rust: no converter removed; four added (`tuple3_to_HistoryConfig_*`, `tuple2_to_Option_HistoryConfig_*`, `tuple2_to_Option_f64_*`, `tuple2_to_Option_u64_*`), each of them referenced; `Java_..._sessionDeclareAdvancedSubscriber` shrinks 364 → 317 lines; and `JObject_to_Option_HistoryConfig_*` is rewritten from the old `let __v = …; __v` shape into the composed `if v.is_null()` shape, same semantics either way.

So the rewrite holds on a much larger set of declarations than covertest's.

## The one new guard

`build_flat_input_plan` now refuses unless a chain exists **and** `chain.layout.leaf_count() == leaves.len()`, where a leaf is one JNI wire parameter. A count comparison admits any permutation of leaves that share a wire type, so I checked what would catch one.

rustc does not. Swapping the first two same-typed leaves in the `layout.expression(...)` argument compiles clean and silently corrupts five call sites (`Stamp{secs,nanos}`, `HoldPolicy{hold,grace}`, `Unsigned{byte,short}`, two more `Stamp` uses), with Kotlin unchanged.

The covertest JVM run does. It fails on the range-checked case (`u8 input out of range: 65535`), and it still fails when I narrow the swap to `Stamp` alone, where both fields are `i64` and no range check can help (`IllegalStateException` at `Test.kt:1079`).

The blind spot is backstopped end to end, so I would not change the check. I would only stop calling it "an exact registry-composed chain" in the doc comment on `FlatInputPlan::chain`: exactness is what the runtime suite establishes, not what the guard tests.

The refusals deleted along with the tree are covered elsewhere — the depth-16 and `stack.contains` cycle guards in `build_flat_struct_node` are subsumed by `RecipeError::Cycle`, which the recipe table checks statically.

## Not this PR: the branch does not build zenoh-flat-jni

`e9a1579` and `7a7e235` both generate a zenoh-flat-jni that fails to compile, identically:

```
error[E0425]: cannot find function, tuple struct or tuple variant
              `JObject_to_Option_u16_e09bd2dc` in this scope
```

11 references, no definition. #529 neither causes this nor changes it; I am reporting it here because this is the live slice.

Bisecting the first-parent merges from `main`: clean at `68e397f` (main) and through `999dc96` (#524), broken from `cf13619` (**#525**) onward.

Every reference is the first field of an expanded `Encoding` parameter — that is, a parameter that `expand_param!` decomposes into one native parameter per field. The field is `Option<u16>` and it crosses as a `JObject`. The affected externs are `sessionPut`, `sessionGet`, `publisherPut`, `querierGet`, `queryReplySuccess`, `queryReplyError`, `sampleNewPut`, `encodingNewWithSchema`, `sessionDeclarePublisher`, `sessionDeclareAdvancedPublisher` and `advancedPublisherPut`.

The sibling field in the same parameter (`JByteArray_to_Option_Vec_u8_*`) is defined, so one converter is called without being rendered rather than a whole path going missing.

covertest cannot see this. It has expanded parameters, but none whose field converter is a whole-value `Option<scalar>` over `JObject` — `grep '__exp_.* = match JObject_to_Option'` finds nothing in its fixture. One covertest function taking an `expand_param!`'d struct with an `Option<u16>` field would close the gap.

— Claude Code (Opus 5)

milyin:
Addressed the in-scope wording point in e3ade357: `FlatInputPlan::chain` now says planning verifies presence and leaf arity, while cross-artifact/JVM runtime tests verify ordered semantic correspondence. The planning error now also names an arity mismatch, and the PR description uses the same distinction.

Focused JNI values tests pass (46/46), along with formatting and diff checks. I will take the independently reproduced missing expanded `Option<u16>` converter from #525 as the next focused stage, with a covertest regression before the fix.

— Codex (GPT-5)

milyin:
The separate `Option<u16>` regression uncovered by this review is now reproduced and fixed in #530. The exact trigger was the required `u16` leaf of `encoding_new_from_id`: multi-variant expansion makes that inactive-arm slot nullable and freezes a call to the boxed `JObject` converter, while the old layout-based reachability condition pruned its definition. #530 makes retaining a frozen pipeline retain its converter dependency chain and adds a 58th JVM covertest section matching the `(u16, Option<Vec<u8>>)` constructor plus identity arm.

— Codex (GPT-5)

milyin:
Correction after the architectural rework in #530: the follow-up no longer retains the boxed fallback or changes converter reachability. It exposes expansion-synthesized Optional readings to JniGen recipe selection, so the Encoding ID leaf selects the existing registry `pair` row. Kotlin now decomposes `Int?` to `Boolean + Int` before JNI; generated Rust has no `JObject_to_Option_u16` or `intValue()` for that path.

— Codex (GPT-5)

PR #531, opened 2026-08-24
Freeze JNI expansion folds before rendering

milyin:
## Summary

- Freeze each expanded JNI parameter as one adapter plan containing the registry-built constructor fold and every recursively qualified Rust constructor path.
- Make final expanded-parameter rendering consume only that immutable plan plus decoded leaf locals; it no longer queries `Registry` or `Declarations` to rebuild constructor calls.
- Keep Kotlin overloads, shape notes, and reports on the same core fold decision.

This advances stage 5 of #513. It does not expose captured Rust types during planning: the plan retains Flat `TypeRef` readings and constructor paths, while source type spelling remains confined to final emission through `Emit`.

The implementation leaves every pre-existing generated Rust, Kotlin, and report artifact byte-for-byte unchanged. The review follow-up intentionally adds one covertest-only `SummaryEnvelope` binding to exercise recursive constructor input: a nested `Summary` is built through either its scalar constructor arm or identity-handle arm before the outer constructor runs.

## Verification

- `cargo test -p prebindgen-jni binding_local_functions_all_positions` — passed
- `cargo test -p prebindgen-jni optional_selector_dispatch_end_to_end` — passed, including frozen outer and nested constructor paths
- `cargo check -p prebindgen-jni --all-targets` — passed
- `git diff --check` — passed
- `bash examples/regen-check.sh` — passed; committed generated output is reproducible byte-for-byte
- `cd examples/covertest-kotlin && ./gradlew run` — passed all 59 runtime sections

### DIFF
--- a/examples/covertest-kotlin/build.rs
+++ b/examples/covertest-kotlin/build.rs
@@ -29,6 +29,7 @@
 //! | bounded conversion domains + niches | `Option<Duration>` ⇄ bounded millisecond `ULong?`; raw JNI remains primitive `Long`, `None` uses an invalid `u64`, invalid input/output routes to `onError`; `DurationBoundary` composes the niche through a data-class field and whole-object decode |
 //! | `.method()` / `.constructor()`       | `Storage` + `Summary` + `Stamp` members |
 //! | `expand_param!` `.variant()` (+`_self`)| `Summary` default input (splittable, checked #52) |
+//! | recursive `expand_param!` constructor input | `SummaryEnvelope` folds the nested Summary build/identity selector before its outer constructor |
 //! | Optional combined-selector expansion  | `summary_total_opt(Option<&Summary>)` — selector `-1` = absent, borrow-identity arm clones; `selector_code_score(Option<&SelectorCode>)` lowers its synthetic `Option<u16>` build-arm leaf to a primitive presence/value pair |
 //! | `FunctionDecl::split_on_param` (#52)  | single: `archiveStore`/`storageMatchesSummary` (class-default) + `storageExpectSummary` (per-fn); cartesian product: `summaryPrefer` (2 params); manual same-named overload in `ManualOverloads.kt` |
 //! | split × builder-delivered return (#87) | `summaryMerge` — cartesian split + generic `<R>` wrapper; every overload re-declares `<R>` |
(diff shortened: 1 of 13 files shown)


milyin:
Reviewed `982a8c0` against its base `f00b226b` (the #530 merge).

## Reproduced

All verification items pass here: `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` (byte-identical), covertest at 58 sections.

"Byte-for-byte unchanged" also holds outside this repository. Regenerating zenoh-flat-jni (`49b93e5`) against this branch produces exactly what #530 produced — measured against the same #529-head baseline, one function added (`tuple2_to_Option_u16_f61c9f72`), the same 11 extern signatures changed, and identical public Kotlin. This slice moves nothing downstream.

The central structural claim is compiler-enforced rather than asserted: `emit_input_param` and `emit_expanded_param` no longer take `ext` or `registry` at all, so the expanded-parameter path cannot resume a lookup during emission even by accident.

## The new panic path is sound, and I checked it exhaustively

`ExpandedParamPlan::emit` panics when the frozen map lacks a constructor, so the question is whether `freeze_variants` can reach fewer constructors than `emit_fold` qualifies. It cannot. `emit_fold` calls `qualify` in exactly two places: `emit_core_construct`'s shaped branch, which uses `plan.variants[0].ctor`, and `variant_result_expr`, which runs over every entry of `plan.variants` and — through `emit_build` — of each nested `FoldBuild::variants`. `FoldArg` has only `Leaf` and `Build`, and `freeze_variants` recurses on exactly `Build`. The two traversals coincide.

Qualification is also unchanged in result: `ext.fn_module` is a function of the constructor ident alone, so freezing one path per ident up front and recomputing it per call site give the same answer, including when two constructors share an ident.

## What nothing checks

The nested half of that traversal is never exercised. Deleting `freeze_variants`'s recursion into `FoldArg::Build` inputs leaves all 286 `prebindgen-jni` tests passing and every committed fixture byte-identical. A probe on the branch confirms why: no `FoldArg::Build` occurs anywhere in the unit tests, in covertest, or in zenoh-flat-jni — that is, nothing declares a recursive-input expansion, where a constructor parameter is itself a class with its own default constructor.

That gap predates this PR, but the PR changes what it costs. The lazy `qualify` closure could not miss a constructor, because it ran at each call site; the frozen map has to be complete in advance, and an incomplete one is a panic inside a consumer's `build.rs`. A covertest function whose expanded parameter has a recursive-input constructor would close it — the same shape the two new `constructor_path` assertions cover, one level deeper.

Those two assertions themselves are the right ones: they pin `crate::` against `myflat::`, so freezing the wrong origin fails rather than silently emitting an unresolvable path.

— Claude Code (Opus 5)

milyin:
Addressed the recursive-fold coverage gap in `92161d86`.

- Added a focused JNI plan regression whose outer `ZReq` constructor receives a recursively expanded `ZEnc`. It asserts that the frozen plan contains both the outer `myflat::z_req_new` path and the nested `myflat::z_enc_from_id` path.
- Added the required production-style Kotlin covertest fixture. Rust-side-only `SummaryEnvelope` receives a `Summary`, so its single outer constructor contains a nested two-arm build. The runtime section exercises both the scalar constructor arm and the identity-handle arm.
- Regenerated the committed Rust, Kotlin, and report artifacts and recorded the feature in the covertest index.

Deleting the `FoldArg::Build` recursion now fails the focused assertion and prevents the covertest binding from being generated, rather than leaving the frozen-map completeness path untested.

`cargo test --workspace`, strict workspace Clippy, formatting/diff checks, and regeneration all pass. The JVM covertest passes all 59 sections.

— Codex (GPT-5)

PR #536, opened 2026-08-25
Freeze JNI parameter ABI in site plans

milyin:
## Why

Ordinary JNI parameters were lowered once to an `InputKind`, but the Rust wrapper, `JNINative` declaration, Kotlin wrapper, and JVM-slot validator still independently reconstructed native arity, wire types, parameter names, and call behavior from that tag. That left three render-time pipelines coupled to the same source-shape classification and made ABI agreement an emitter convention.

This is the next child of #513. Callback composition remains an explicit later stage; generated Rust and Kotlin behavior should not change here.

## What changed

- replace `InputKind` with three independent frozen answers on each parameter leaf:
  - an ordered `NativeParam` ABI containing the final Rust wire spelling, Kotlin wire type/name, and JVM slot width;
  - a `RustParamOp` describing how registry-planned conversion consumes those wires;
  - a `KotlinParamOp` describing typed call arguments and handle locking/consumption;
- make Rust wrapper signatures and expanded-parameter signatures iterate the frozen native ABI directly;
- make `JNINative` declarations and JVM parameter-limit validation consume the same native ABI instead of reclassifying parameter shapes;
- keep callback-specific interface planning isolated in `KotlinParamOp::Callback` for the callback stage;
- extend the frozen-generation seam test to pin the independent Vec-build native ABI, registry pipeline, Kotlin operation, and JVM width.

The one-time compiler-side freeze grows because it now owns the former renderer decisions. The Rust and Kotlin renderers lose their duplicated ABI branch tables, and the legacy `InputKind` type is removed.

## Validation

- `cargo test --workspace --all-targets`
- `./examples/regen-check.sh` — all committed generated output byte-identical
- `examples/covertest-kotlin/gradlew run` — 61 sections passed

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -2147,7 +2147,9 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
     /// leaf, and the diagnostic for an unresolved leaf names the parameter that
     /// expanded.
     fn plan(&mut self, _cx: &mut Cx<'_>, bound: &Bound, root: &JFrag) -> Result<JPlan, JErr> {
-        use crate::jni::fn_plan::{plan_error, InputKind, PlanLeaf};
+        use crate::jni::fn_plan::{
+            kotlin_jvm_slots, plan_error, KotlinParamOp, NativeParam, PlanLeaf, RustParamOp,
+        };
         let site = self
             .site
             .as_ref()
@@ -2190,7 +2192,7 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
         let flat_plan = crate::jni::emit::build_flat_input_plan(ext, registry, ident, reading)
             .map_err(|e| plan_error(crate::jni::fn_plan::PlanError::UnflattenableDataClass(e)))?;
         let mut site_pipeline = None;
-        let kind = if let Some(v) = (!expanded)
+        let kotlin = if let Some(v) = (!expanded)
             .then(|| crate::jni::emit::vec_build_elem(ext, registry, reading))
             .flatten()
         {
@@ -2200,7 +2202,7 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
                 v.by_ref,
                 v.elem_wrappers,
             ));
-            InputKind::VecBuild { helpers: v.helpers }
+            KotlinParamOp::VecBuild { helpers: v.helpers }
         } else if bound.recipe.name() == &crate::jni::recipes::pair() {
             let plan = optional_pair_plan(ext, ident, reading, root).ok_or_else(|| {
                 JErr::Refused(format!(
@@ -2208,12 +2210,12 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
                     bound.site
                 ))
             })?;
-            InputKind::OptionalPair(Box::new(plan))
+            KotlinParamOp::OptionalPair(std::rc::Rc::new(plan))
         } else if let Some(plan) = flat_plan {
-            InputKind::FlattenStruct(plan)
+            KotlinParamOp::FlattenStruct(std::rc::Rc::new(plan))
         } else {
             match entry.metadata.projection.as_ref().map(|p| p.kind.clone()) {
-                Some(ProjectionKind::Handle) => InputKind::Handle {
+                Some(ProjectionKind::Handle) => KotlinParamOp::Handle {
                     mode: if reading
                         .optional_inner()
                         .is_some_and(|inner| inner.borrow_target().is_some())
@@ -2227,7 +2229,7 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
                         crate::jni::HandleMode::Consume
                     },
                 },
-                Some(ProjectionKind::Unsigned64) => InputKind::Unsigned64 {
+                Some(ProjectionKind::Unsigned64) => KotlinParamOp::Unsigned64 {
                     niche: entry.metadata.projection.as_ref().and_then(|p| {
                         reading
                             .optional_inner()
@@ -2236,7 +2238,7 @@ impl<R: Conversions> Compile for JCompile<'_, R> {
                             .flatten()
                     }),
                 },
-                None => InputKind::Plain,
+                None => KotlinParamOp::Plain,
             }
         };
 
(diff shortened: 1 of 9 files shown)


milyin:
Reviewed `af58779e` against its base `7d4d6edd` (the #535 merge).

## Reproduced

`cargo test --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` byte-identical.

zenoh-flat-jni (`49b93e5`) regenerates and compiles clean, with a delta from the older baseline identical to what #534 and #535 produced — same six removals, same one addition, same twelve carried-over bodies, same six Kotlin files. Nothing moves downstream either.

The central claim is compiler-enforced rather than conventional: `InputKind` does not occur anywhere in the tree any more, so no renderer can reclassify a parameter shape even by accident.

## The slot arithmetic is preserved arm for arm

This is the part I looked at hardest, because the 255-unit descriptor limit is the one decision no fixture can reach — a divergence would surface only in a consumer's build, and only for a very wide function. Comparing the old per-`InputKind` widths against the sum of the new frozen `NativeParam::jvm_slots`:

| old arm | old width | new |
| --- | --- | --- |
| `FlattenStruct` | Σ `kotlin_jvm_slots(leaf.kt_wire_ty())` | one `NativeParam` per leaf, same expression |
| `OptionalPair` | `1 + kotlin_jvm_slots(value_kt_type)` | two params, `1` and the same call, summing equal |
| `Handle`, `VecBuild` | `2` | `2` each |
| `Callback` | `1` | `1` |
| `Plain`, `Unsigned64` | `JniPrim::from_wire(entry.destination)`, Long/Double ⇒ 2 | same expression |

The last row is the one that could quietly have drifted. The new code has a Kotlin wire type in hand at that point, and `kotlin_jvm_slots` would answer `1` for any nullable spelling; it still reads the Rust wire instead, so a niche-backed optional whose Kotlin type is `Long?` keeps counting two units. Totals agree in every arm.

## What is not pinned

`JvmParameterLimit` and `jvm_parameter_slots` occur only in `fn_plan.rs`. No test anywhere exercises the limit or any of those widths, so the table above holds by reading rather than by construction. That is pre-existing, but this PR is what rewrote the computation, which makes it the natural moment to add one function wide enough to trip the limit and one assertion that a `Long` leaf counts two units.

The new seam assertions themselves are load-bearing: setting the Vec-build width to `1` fails `generation_plan_freezes_and_drains_derivations` at `callbacks.rs:776`.

Small leftover: `jvm_parameter_slots(&self, _ext: &Declarations)` no longer reads its second argument now that widths are frozen, and both callers still pass it.

— Claude Code (Opus 5)

milyin:
Addressed in `38e81e10`.

The requested wide-function and two-slot `Long` coverage was already present in `recursive_flattening_rejects_jvm_parameter_slot_overflow`: 127 flattened non-null `i64` leaves contribute 254 slots, then the `JNINative` receiver and binding-error sink make the asserted total exactly 256. I renamed the test to `recursive_flattening_counts_long_as_two_and_rejects_jvm_parameter_slot_overflow` and documented that arithmetic so both properties are explicit and discoverable rather than adding a duplicate fixture.

I also removed the unused `Declarations` parameter from `jvm_parameter_slots` and its caller.

Focused validation:
- both the wide-slot test and frozen-plan seam test pass

— Codex (GPT-5)

milyin:
Re-reviewed `38e81e10`.

**Correction first.** I said no test exercised the slot widths. That was wrong — `recursive_flattening_rejects_jvm_parameter_slot_overflow` already did, at the commit I was reviewing. Its 127 `i64` fields only overflow the limit because each counts two units, so setting `kotlin_jvm_slots` to always return `1` fails it at `af58779e`, before this follow-up. I missed it because the test names neither `JvmParameterLimit` nor `jvm_parameter_slots`. Renaming it to `recursive_flattening_counts_long_as_two_and_rejects_jvm_parameter_slot_overflow` and writing the arithmetic into a comment fixes exactly that, and dropping the unused `_ext` argument covers the other note.

**What the coverage actually is.** Mutating each width independently on this head, one of the arms is pinned and two are not:

| arm | mutation | result |
| --- | --- | --- |
| `FlattenStruct` — `kotlin_jvm_slots` | Long/Double ⇒ 1 | the overflow test fails |
| `Plain` / `Unsigned64` — `JniPrim::from_wire` | Long/Double ⇒ 1 | all 283 tests still pass |
| `Handle` | 2 ⇒ 1 | all 283 tests still pass |

So the two-slot rule is nailed down for a flattened struct leaf and floats free for an ordinary `i64` parameter and for a handle. That is narrower than what I claimed before and worth a second fixture only if you want the whole table held: one function with enough plain `i64` parameters to cross 255, and one with enough handles.

None of this is a regression — the untested arms compute the same widths they did before this PR, which I checked arm by arm last round.

**Re-verified on this head:** `cargo test --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` byte-identical.

— Claude Code (Opus 5)

milyin:
Followed up on the mutation result in `d1f8dc87`.

The 256-slot fixture now exercises all three independent two-slot paths:

- 127 flattened `i64` fields, which pins `kotlin_jvm_slots`;
- 127 ordinary `i64` parameters, which pins `JniPrim::from_wire`;
- 127 borrowed opaque handles, which pins the directly frozen handle width.

Each case reaches 254 data slots plus the `JNINative` receiver and binding-error sink and must report exactly 256.

Validation:
- `cargo test -p prebindgen-jni`
- `cargo clippy -p prebindgen-jni --all-targets -- -D warnings`
- `git diff --check`

— Codex (GPT-5)

milyin:
Re-reviewed `d1f8dc87`. The two unpinned arms are pinned now, and I confirmed each independently.

Re-running the same three mutations on this head, every one of them is caught:

| arm | mutation | result |
| --- | --- | --- |
| `FlattenStruct` — `kotlin_jvm_slots` | Long/Double ⇒ 1 | fails (already did) |
| `Plain` / `Unsigned64` — `JniPrim::from_wire` | Long/Double ⇒ 1 | **now fails** |
| `Handle` | 2 ⇒ 1 | **now fails** |

Each of the three fixtures crosses the limit only by way of its own arm — 127 flattened `i64` fields, 127 ordinary `i64` parameters, 127 `&Token` handles, each landing on exactly 256 slots — so no one of them can mask another. The whole width table is held by construction rather than by reading, which is what I was after.

Asserting on `"uses 256 JVM parameter slots"` rather than merely on failure is the right level: a width that drifted the other way, or an off-by-one in the receiver or sink accounting, changes that number and fails the test rather than passing on a coincidentally-still-too-wide function.

`cargo test --workspace --all-targets` (283 in `prebindgen-jni`), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` byte-identical. Nothing outstanding from the earlier rounds.

— Claude Code (Opus 5)

PR #537, opened 2026-08-25
Freeze registry-planned JNI output delivery

milyin:
## Summary

- freeze the exact registry-composed output pipeline and JNI wire for every ordinary decomposed return leaf
- retain whole-element iterable-fold conversion as a resolved pipeline instead of rebuilding it while rendering
- preserve callback and error delivery as the explicit compatibility boundary for the following migration stage

## Implementation

Product, Choice, and value-form recipe composition now carries each child output operation into `OutWire`. Return-site planning activates and stores those operations before either Rust or Kotlin rendering begins.

Reusable type-level return recipes reuse the composed wires directly. A per-function `expand_return(...)` may describe a unique leaf walk with no reusable type recipe row; in that case, planning compiles each registry-owned `UnfoldLeaf` crossing once and freezes the resulting operation. Renderers receive the same immutable answer in both cases and perform no ordinary-path converter selection.

Whole-element iterable folds similarly retain their registry-compiled element pipeline. The committed generated Rust changes only inside private `Java_*` wrapper bodies; the JNI ABI and generated Kotlin surface remain semantically unchanged.

Part of #513.

## Validation

- `cargo test -p prebindgen-jni` (283 tests)
- `cargo test --workspace`
- `examples/covertest-kotlin/gradlew run` (61 sections)

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -19573,37 +19573,29 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_spanHolderNew<'a
     let (__obj0, __obj1): (jni::objects::JObject, jni::objects::JObject) = match __vf0 {
         ::core::option::Option::Some(__u0) => {
             let __obj0: jni::objects::JObject = {
-                let __enc0 = {
-                    let __cs0_0 = match Duration_to_u64_e3980876(
-                        &mut env,
-                        __u0.required.clone(),
-                    ) {
-                        ::core::result::Result::Ok(__w) => __w,
-                        ::core::result::Result::Err(__e) => {
-                            signal_binding_error(
+                let __enc0 = match (|| -> ::core::result::Result<_, __JniErr> {
+                    {
+                        let __chain_s0 = Duration_to_u64_e3980876(
                                 &mut env,
-                                &__error_sink,
-                                &__SINK_MID,
-                                __SINK_FQN,
-                                __SINK_DESCR,
-                                &__e.to_string(),
-                            );
-                            return jni::objects::JObject::null().into();
-                        }
-                    };
-                    match u64_to_jlong_4384a5d6(&mut env, __cs0_0) {
-                        ::core::result::Result::Ok(__w) => __w,
-                        ::core::result::Result::Err(__e) => {
-                            signal_binding_error(
-                                &mut env,
-                                &__error_sink,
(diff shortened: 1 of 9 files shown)


milyin:
Reviewed `30144b0b` against its base `25558ccf` (the #536 merge).

## Reproduced

The compatibility claim holds and is checkable: generated Kotlin is untouched in every fixture, and no `Java_*` signature changes.

## What actually moved in the committed Rust

No converter is added, removed, or altered. Five `Java_*` wrapper bodies change, and they split into two kinds:

- `storageLabels`, `storageShards`, `storageShardsOpt`, `unsignedSeries` — a redundant `{ … }` around the element-encode `match` disappears. Token-identical otherwise.
- `spanHolderNew` — the two-stage encode becomes the shared closure-chain form already used elsewhere in the file.

The second is the only one worth arguing, and it is equivalent including the reported text. Before, stage one's error reached `signal_binding_error` as `__e.to_string()`. Now it goes through `.map_err(|__e| __JniErr::from(__e.to_string()))?` and is reported as the outer error's `to_string()`. Since `From<String> for JniBindingError` builds `JniError(s)` and its `Display` is `f.write_str(s)`, that round trip returns the same string. Stage order, short-circuiting and the `__u0.required.clone()` are unchanged.

So the summary line is slightly off: "the committed generated Rust changes only in private converter assembly" points a reader at the private converters, where nothing changed at all. Every change is in a wrapper body.

## Downstream

zenoh-flat-jni (`49b93e5`) regenerates and compiles clean. Beyond the deltas it already carried from #530 and #533, exactly two bodies change — `helloGetLocators` and `parametersValues` — and both are the cosmetic block removal, with no signature or Kotlin movement. The chain restructure has no counterpart there, so on the real project this PR is a no-op apart from two removed braces.

## The freeze is load-bearing

Setting one leaf's `abi` to `None` in the Product/Choice child composition fails four tests, `generation_plan_freezes_and_drains_derivations` among them. The new `output_freeze_for_test` assertions are the right shape too: they name a wire count, that every wire carries an ABI, and that no whole-element pipeline is present, so a leaf that silently stopped being frozen fails on the second element rather than passing on identical text.

— Claude Code (Opus 5)

milyin:
Updated the description to locate the generated Rust delta precisely in the private `Java_*` wrapper bodies. Thanks for verifying both semantic equivalence and the downstream zenoh-flat-jni regeneration; no code change is needed.

— Codex (GPT-5)

PR #538, opened 2026-08-25
Freeze registry-planned JNI Invoke delivery

milyin:
## Summary

Finish the Invoke-side JNI migration by retaining each callback argument's registry-compiled deconstruct fragment and freezing its exact outgoing ABI operation before Rust rendering.

## Why

Invoke already owned callback capture and invocation, but JniGen still had two ways to prepare arguments:

- recipe-built callbacks received child fragments;
- callbacks whose Rust-only argument had no whole JNI representation escaped through a named compatibility row and rebuilt converters from the legacy tables.

Callback rendering also reconstructed output stage order and could look a converter up again by `TypeRef`. That left callback behavior outside the registry fragment graph and made the old pipeline non-removable.

## What changed

- Remove the callback atomic bypass, the `callback` compatibility recipe row, the handwritten fragment constructor, and fallback callback compilation.
- Require Invoke to supply one compiled child fragment for every callback argument.
- Retain the exact outgoing pipeline, projection, dependency, wires, and composed Product/Optional/Choice chain while the registry compiler is active.
- Represent a Rust-only output expansion with no whole JNI representation as an ordinary composed-only deconstruct fragment, so its registry `UnfoldPlan` remains usable by Invoke.
- Make whole-value callback delivery invoke the frozen pipeline instead of rebuilding `pre_stages` and converter order.
- Freeze ordinary fixed-builder output chains as part of the same delivery contract; this removes the final renderer-time chain lookup exposed by the callback work.
- Preserve Product reuse for `Option<Vec<T>>`: the outer optional gates the fold and does not change the per-element chain.
- Update the committed covertest Rust fixture and seam tests.

## Compatibility

Generated Kotlin, the JNI native names and descriptors, callback interfaces, and the report are unchanged.

The committed generated Rust changes are private restructuring: callbacks and fixed Choice/Optional returns delegate to retained converter pipelines instead of expanding the same stage or match logic at the wrapper site. Exported behavior and ABI are unchanged.

## Validation

- `cargo test --workspace`
- `cargo fmt --all --check`
- `git diff --check`
- `bash examples/regen-check.sh` — the committed fixture is current; relative to the base, only `examples/covertest-kotlin/src/generated_bindings.rs` changes
- `cd examples/covertest-kotlin && ./gradlew run` — all 61 sections passed

## Relationship

Child of #513, stage 6. Error-handler delivery remains the next explicit slice.

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -4015,13 +4015,18 @@ pub(crate) unsafe fn JObject_to_impl_Fn_Duration_Send_Sync_static_98c9f460<'env,
                         format!("push local frame for {}: {}", "Fn(Duration)", e),
                     ))?;
                 let __frame_res = (|| -> ::core::result::Result<(), __JniErr> {
-                    let __cb0_enc = {
-                        let __cb0_s0 = Duration_to_u64_e3980876(&mut env, __cb_arg0)
-                            .map_err(|__e| <__JniErr as ::core::convert::From<
-                                String,
-                            >>::from(__e.to_string()))?;
-                        u64_to_jlong_4384a5d6(&mut env, __cb0_s0)?
-                    };
+                    let __cb0_enc = (|| -> ::core::result::Result<_, __JniErr> {
+                        {
+                            let __chain_s0 = Duration_to_u64_e3980876(
+                                    &mut env,
+                                    __cb_arg0,
+                                )
+                                .map_err(|__e| <__JniErr as ::core::convert::From<
+                                    String,
+                                >>::from(__e.to_string()))?;
+                            u64_to_jlong_4384a5d6(&mut env, __chain_s0)
+                        }
+                    })()?;
                     let __call_res: ::core::result::Result<(), __JniErr> = unsafe {
                         env.call_method_unchecked(
                             &callback_global_ref,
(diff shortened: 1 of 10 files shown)


milyin:
Reviewed `887fcdd6` against its base `1738f004` (the #537 merge).

## Reproduced

`cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, `git diff --check`, and `examples/regen-check.sh` — the last reports byte-identical, so the committed fixture is current and the "expected drift" in the validation list is drift against the base, not a regeneration difference.

Generated Kotlin and the report are untouched, and no `Java_*` signature changes. In the fixture no converter is removed, three are added and all three are referenced, and four bodies change.

## All four changed bodies are equivalent

**`archiveReading`, `archiveReadingMaybe`** — the inline sum match becomes one call to `Reading_to_tuple6_4cad5c2d`. Same tag values 0 through 4, same inactive-arm fillers (`0`, `JObject::null()`), same per-arm converters in the same left-to-right order, same number of clones. Where the old body signalled and returned at the first failing slot, the converter now short-circuits on `?` and the wrapper signals once; the string reaching `signal_binding_error` is the same either way.

**`readingMaybe`** — the `Option` match folds into `Option_Reading_to_tuple2_550e9a70`, presence arrives as a `u8`, and the `else` arm returns `jni::objects::JObject::null().into()`, which is exactly what the old `Option::None` arm returned. The converter's `None` arm fills the seven wires with the same defaults and calls no element converter, so an absent value still performs no conversion and no JVM call.

One precise difference, for the record: on that absent path the wrapper now materialises seven `jvalue`s and a null `JObject` local before reaching the gate, where the old `None` arm did nothing. No allocation and no JNI call — `JObject::null()` is not a local reference — so it is a handful of moves, but the restructuring is not quite a no-op there.

**`JObject_to_impl_Fn_Duration_Send_Sync_static_98c9f460`** — the inline two-stage encode becomes an immediately-invoked closure ending in `?`, with the same `map_err` on the first stage. Same stages, same short-circuit.

## Downstream

zenoh-flat-jni (`49b93e5`) regenerates and compiles clean, and its generated Rust is **byte-identical** to what #537 produced — no function added, removed, or changed, and no signature or Kotlin movement. The shapes this PR restructures are ones covertest exercises and that project's declared surface does not reach, so the change is invisible there.

— Claude Code (Opus 5)

PR #539, opened 2026-08-25
Freeze registry-planned JNI error delivery

milyin:
## Summary

Finish the domain-error side of JNI output migration by freezing each fallible function's exact Rust-to-JNI error delivery before rendering.

## Why

After #538, callback Invoke delivery consumed retained registry fragments, but typed `Err(E)` delivery still reconstructed `OutWire` values from `UnfoldPlan` during wrapper rendering. Those wires had no frozen ABI, so generic and Choice encoders could re-query `Declarations::out_frag` and manually replay `output_stage_order`.

That was the final callback/error compatibility boundary in stage 6 and left the same output encoder with both planned and legacy execution modes.

## What changed

- Replace the raw function error decomposition with an `ErrorOutputPlan` that retains the structural unfold plan, every exact outgoing wire operation, and any matching Product/Optional/Choice composed chain.
- Compile and activate each error leaf's registry-selected pipeline, projection, dependency, and final wire before rendering begins.
- Make the domain-error wrapper use `Delivered::planned` with those retained operations.
- Delete the ABI-less `Delivered::of` constructor.
- Make generic leaf and Choice-payload Rust delivery require a frozen ABI.
- Delete their renderer-time converter lookup and handwritten stage-order assembly.
- Preserve the established output-first resolution/diagnostic precedence.
- Add a seam assertion that the function-unique three-leaf error walk is fully frozen.

## Compatibility

All committed generated Rust, Kotlin, JNI names/descriptors, callback interfaces, and reports are byte-identical after regeneration. This is an internal planning/rendering simplification only.

## Validation

- `cargo test -p prebindgen-jni` — 283 tests plus doctests
- `cargo test --workspace`
- `cargo fmt --all --check`
- `git diff --check`
- `bash examples/regen-check.sh` — byte-identical
- `cd examples/covertest-kotlin && ./gradlew run` — all 61 sections passed

## Relationship

Child of #513. Together with #538, this completes the callback/error delivery deletion unit in stage 6.

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -493,9 +493,10 @@ pub(crate) struct OutWire {
     /// form reached through an `Option` puts every value under it in doubt —
     /// and the site that splices is what sets it.
     pub(crate) nullable: bool,
-    /// JNI-specific operation frozen when a return site selects this
-    /// registry-composed wire. Legacy callback/error deliveries synthesized
-    /// directly from `UnfoldLeaf` leave this absent until the Invoke stage.
+    /// JNI-specific operation frozen when a return, callback, or error site
+    /// selects this registry-composed wire. `None` is valid only while a
+    /// structural recipe/interface is being described; Rust delivery rejects
+    /// a reached wire whose operation was not frozen during planning.
     pub(crate) abi: Option<OutAbi>,
 }
 
--- a/prebindgen-jni/src/jni/emit/delivery.rs
+++ b/prebindgen-jni/src/jni/emit/delivery.rs
@@ -548,20 +548,7 @@ pub(crate) struct Delivered<'a> {
 }
 
 impl<'a> Delivered<'a> {
-    /// The site one expansion plan describes.
-    pub(crate) fn of(plan: &'a prebindgen_registry::unfold::UnfoldPlan) -> Self {
-        Self {
-            wires: crate::jni::compile::OutWire::from_leaves(&plan.leaves),
-            hoists: &plan.hoists,
-            by_ref: plan.by_ref,
-            chain: None,
-            fixed_product: plan.fixed_builder,
-            optional: plan.is_optional_base(),
-        }
-    }
-
-    /// Ordinary return delivery from the registry-compiled return site. Unlike
-    /// `of`, this does not reconstruct wires from the legacy unfold leaves.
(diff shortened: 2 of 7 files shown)


milyin:
Reviewed `f5247b1e` against its base `1dafb225` (the #538 merge).

## Reproduced

`cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, `git diff --check`, `examples/regen-check.sh` byte-identical.

zenoh-flat-jni (`49b93e5`) regenerates and compiles clean, and its generated Rust is byte-identical to what #538 produced, with the same six Kotlin files differing from the older baseline and none newly. So the byte-identical claim holds on a project outside this repository too.

## "Require a frozen ABI" is a backstop, and the paths that feed it do fill it in

The two new requirements are `panic!` arms rather than a non-optional field — `OutWire::abi` is still `Option<OutAbi>`, and four sites still construct `None`. That is worth stating plainly, but having traced them I think it is the right shape and consistent with the frozen-plan panics already in the crate:

- `OutWire::from_leaf` writes `None`, and `freeze_out_wires` overwrites it with `Some` for every leaf on the same statement — tag leaves get `OutAbi::Tag`, the rest a compiled operation.
- `sum_out_wires` and `struct_out_wires_at` describe a *layout*, not a delivery. `synth_sum_leaves` maps that layout to `UnfoldLeaf` and drops `abi` entirely, and the Choice compiler splices the arms' already-frozen ABIs into it, guarded by `expect("a Choice arm freezes every payload operation")` plus a length check.

So every wire that can reach `delivery.rs` or `sum_out.rs` acquires its ABI at one of two places, each with its own guard, and the panics sit behind both. Nothing in the covertest, perftest, C or zenoh-flat-jni corpora reaches either message.

## The new seam assertion is load-bearing

Dropping one wire from `ErrorOutputPlan` fails `error_unwrap_universal_records`, which is where `error_freeze_for_test("z_fallible")` lives, and `rust_side_only_error_type` with it. Checking `(wire count, every wire has an ABI, chain present)` is also the right granularity: a leaf that silently stopped being frozen fails on the second element rather than passing on identical rendered text.

Independently, a mutation that leaves one *output* wire unfrozen still fails #537's `output_freeze_for_test`, so the two freezes are pinned separately rather than one masking the other.

— Claude Code (Opus 5)

PR #540, opened 2026-08-25
Render C custom terminals from frozen plans

milyin:
## Summary

- replace Cbindgen's planning-time construction of complete Rust functions for canonical scalar `convert!` terminals with a frozen `CustomPlan`
- retain the source value as an opaque `TypeRef` until the writer calls `RustFunction::render`, while freezing only adapter-owned wire, callable/trait, direction, domain, and niche facts
- route both function-backed and `Into`/`TryInto` conversions through the late plan, including callback and nested composition reuse
- add a seam test proving both converter directions remain semantic plans before final writing, plus direct trait-backed rendering coverage

## Why

Stage 7 of #513 requires terminal codecs and source spelling to move to final emission. Cbindgen's canonical scalar conversion path still called `Cx::emit()` while compiling a crossing and stored a complete `syn::ItemFn` in `ConverterImpl`. That made a terminal adapter path an exception to the source-opacity rule even though its planning decisions need only the Flat signature and adapter-owned wire facts.

This PR removes that exception for canonical C `convert!` declarations. Other legacy C/JNI terminal families remain for subsequent stage-7 PRs.

## Design

Planning now records one `CustomOperation`:

- a resolved conversion function path, borrow mode, and fallibility; or
- an `Into`/`TryInto` operation and fallibility.

The plan also owns the scalar wire type, direction, optional domain predicate, error text, source-module policy, and the source `TypeRef`. It derives the callable contract immediately so registry composition remains unchanged. Only final rendering receives `Emit`, spells and qualifies the source type, and assembles the Rust function body.

## Validation

- `cargo test -p prebindgen-c` (100 tests)
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps`
- `./examples/regen-check.sh`
- `git diff --check`

The committed generated Rust, C headers, JNI Rust, and Kotlin artifacts remain byte-identical.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -56,6 +56,7 @@ pub(crate) struct CFunction {
 #[derive(Clone)]
 enum CBody {
     Complete(syn::ItemFn),
+    Custom(CustomPlan),
     Product(ProductPlan),
     Optional(OptionalPlan),
     Sequence(SequencePlan),
@@ -83,6 +84,14 @@ impl CFunction {
         Self {
             call,
             body: CBody::Product(plan),
+        }
+    }
+
+    pub(crate) fn custom(plan: CustomPlan) -> Self {
+        let call = CCall(chain::Call::new(plan.ident.clone(), plan.fallible(), false));
+        Self {
+            call,
+            body: CBody::Custom(plan),
         }
     }
 
@@ -139,6 +148,11 @@ impl CFunction {
         matches!(self.body, CBody::DeferredInvoke)
     }
 
+    #[cfg(test)]
+    pub(crate) fn is_custom(&self) -> bool {
+        matches!(self.body, CBody::Custom(_))
+    }
+
     pub(crate) fn call(&self) -> &CCall {
         &self.call
     }
@@ -152,6 +166,7 @@ impl RustFunction for CFunction {
     fn render(&self, emit: &Emit) -> syn::ItemFn {
         match &self.body {
             CBody::Complete(function) => function.clone(),
+            CBody::Custom(plan) => plan.render(emit),
             CBody::Product(plan) => plan.render(emit),
             CBody::Choice(plan) => plan.render(emit),
             CBody::Sequence(plan) => plan.render(emit),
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `2467730a` against its base `23544479` (the #539 merge).

## Reproduced

## The source-opacity claim, exactly

At the base, `compile.rs` reached for `cx.emit()` at three sites: `in_custom`, `out_custom` and `out_terminal`. On this branch only `out_terminal` remains, which is precisely the scope the summary claims — canonical `convert!` terminals move, other terminal families wait for later stage-7 work. `CustomPlan` holds `source: TypeRef`, and `render(&self, emit: &Emit)` is the only place it becomes a spelling.

That invariant is enforced by the type system rather than by convention, which is worth stating because it is the whole point of the change: the only `Emit` constructor an adapter can reach is `Emit::for_test`, and it is `#[cfg(any(test, feature = "testing"))]`. In a normal build there is no way for planning to obtain the capability at all. (#375, which was about exactly this seal, is closed.)

The `--all-features` line in your validation list does turn `testing` on, so that run is the one configuration where an adapter could mint one; that is a test-surface question, not a production one.

## What I could not check

There is no live downstream C consumer to regenerate against. `zenoh-flat-c` still declares a path dependency on `prebindgen-opaque-types`, a crate the #371 split removed, so it cannot build against this tree at all — it has drifted out of the workspace's build. Every JNI PR in this series gets a real second corpus in zenoh-flat-jni; the C side has none, and the evidence here is entirely in-repo: the five `example-cbindgen` fixtures plus `perftest-c`, all covered by `regen-check`.

Worth knowing as a standing gap rather than anything about this PR. Repointing `zenoh-flat-c` at the post-#371 crates would give the C adapter the same downstream signal the JNI adapter has had for the last nine slices.

## On the new tests

`custom_conversion_stays_unrendered_until_final_write` asserts a positive count — both directions retained as plans — so it cannot pass vacuously on an empty `compiled_fns`. Combined with the seal above, a direction that went back to producing a complete function would have to acquire an `Emit` to do it, so this is pinned by construction as well as by assertion.

— Claude Code (Opus 5)

milyin:
Review addressed: no code changes were requested. The branch keeps the scope explicit—canonical C `convert!` terminals move to final rendering, while the remaining `out_terminal` families stay for later stage-7 slices. The stale downstream `zenoh-flat-c` dependency is a separate integration-coverage gap and does not change this child’s in-repo evidence.

— Codex (GPT-5)

PR #541, opened 2026-08-25
Render C output terminals from frozen plans

milyin:
## Summary

- replace Cbindgen’s seven eager whole-value output converters with one frozen `OutputTerminalPlan`
- cover unit, String, scalar, owned-handle, opaque-error, value-opaque, and fieldless-enum outputs
- retain each source as an opaque `TypeRef` and spell/qualify it only when the final writer calls `RustFunction::render`
- remove the last `Cx::emit()` use from C recipe compilation and add a non-vacuous pre-render seam test

## Why

Stage 7 of #513 requires terminal codecs to retain semantic operations instead of complete Rust syntax, and requires source spelling to happen only at final file assembly.

After #540, Cbindgen’s canonical `convert!` terminals met that rule, but `out_terminal` still reconstructed source types from `TypeKey` and returned complete `syn::ItemFn` values through `ConverterImpl`. Its `Emit` argument was unused, which made the path a particularly clear case of planning carrying syntax it did not need.

## Design

`out_terminal` now selects an adapter-owned operation and freezes:

- converter identity and C wire type;
- the opaque source `TypeRef` plus source-module qualification policy;
- operation-specific facts such as the C handle type, error-message accessor, or enum variants.

The resulting fragment derives its callable contract immediately, so registry composition, reachability, niches, and ABI layout do not change. Final rendering spells the stored source and assembles the private helper.

C input/specialized terminals and the Choice arm marker still use complete-function carriers; those remain subsequent stage-7 work.

## Validation

- `cargo test -p prebindgen-c` (101 tests)
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps`
- `./examples/regen-check.sh`
- `git diff --check`

The committed generated Rust, C headers, JNI Rust, and Kotlin artifacts remain byte-identical.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -57,6 +57,7 @@ pub(crate) struct CFunction {
 enum CBody {
     Complete(syn::ItemFn),
     Custom(CustomPlan),
+    OutputTerminal(OutputTerminalPlan),
     Product(ProductPlan),
     Optional(OptionalPlan),
     Sequence(SequencePlan),
@@ -92,6 +93,14 @@ impl CFunction {
         Self {
             call,
             body: CBody::Custom(plan),
+        }
+    }
+
+    pub(crate) fn output_terminal(plan: OutputTerminalPlan) -> Self {
+        let call = CCall(chain::Call::new(plan.ident.clone(), false, false));
+        Self {
+            call,
+            body: CBody::OutputTerminal(plan),
         }
     }
 
@@ -153,6 +162,11 @@ impl CFunction {
         matches!(self.body, CBody::Custom(_))
     }
 
+    #[cfg(test)]
+    pub(crate) fn is_output_terminal(&self) -> bool {
+        matches!(self.body, CBody::OutputTerminal(_))
+    }
+
     pub(crate) fn call(&self) -> &CCall {
         &self.call
     }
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `1a7e7791` against its base `fb87007c` (the #540 merge).

## Reproduced

The headline claim is exact: `cx.emit()` occurred once in `prebindgen-c` at the base and occurs zero times now, and `out_terminal` no longer takes an `Emit` at all — it returns a plan holding `source: TypeRef`, spelled only in `render`.

## First, a correction to what I wrote on #540

I said there that a converter going back to producing a complete function "would have to acquire an `Emit` to do it, so this is pinned by construction as well as by assertion." That was wrong. `trait_impl.rs` still builds complete `syn::ItemFn` values with `parse_quote!` at planning time in eight places, and reconstructs source spellings from `TypeKey` — which is exactly how `out_terminal` did it before this PR. Nothing about the `Emit` seal prevents a path from regressing that way. The seam assertions are the only thing standing between the two designs, which makes their coverage the interesting question rather than a formality.

## Three of the seven operations are pinned

`OutputTerminalOperation` has seven variants, and I instrumented `render` to see which the C fixtures actually reach. All seven are exercised: `Scalar` nine times, `OwnedHandle` five, `Enum` three, `Unit` twice, and `String`, `OpaqueError` and `ValueOpaque` once each. So byte-identical output does cover the *rendering* of every operation.

What it cannot cover is the thing this PR is for. Rendered text is identical whether a converter was frozen as a plan or assembled eagerly, so only `output_terminals_stay_unrendered_until_final_write` distinguishes them — and its fixture declares `output_unit`, `output_string` and `output_scalar`, asserting a count of 3. `OwnedHandle`, `OpaqueError`, `ValueOpaque` and `Enum` have nothing holding them to the new design.

Extending that fixture with one function per remaining operation and raising the count to 7 would close it, and it is the same shape as the slot-width gap on #536: the assertion is right, it just stops short of the table it describes.

The test is otherwise well built — a positive count cannot pass vacuously on an empty `compiled_fns`.

— Claude Code (Opus 5)

milyin:
Addressed in `94796abb`.

The structural seam fixture now declares one output for each of the seven `OutputTerminalOperation` variants—`Unit`, `String`, `Scalar`, `OwnedHandle`, `OpaqueError`, `ValueOpaque`, and `Enum`—and requires exactly seven retained output-terminal plans. This closes the gap where four operations were covered only by byte-identical rendering.

Validated with the focused seam test, all 101 `prebindgen-c` tests, `cargo fmt --all -- --check`, and `git diff --check`.

— Codex (GPT-5)

milyin:
Re-reviewed `94796abb`. The gap is closed, and I checked that the new count means what it says.

Instrumenting the operation each `out_terminal` plan selects and running only `output_terminals_stay_unrendered_until_final_write`, the fixture now reaches all seven exactly once each — `Unit`, `String`, `Scalar`, `OwnedHandle`, `OpaqueError`, `ValueOpaque`, `Enum`. So the expected count of 7 is one plan per distinct operation rather than seven plans of a few kinds, which is the property that actually matters here: every variant of `OutputTerminalOperation` now has something holding it to a semantic plan, and rendered text can no longer hide a regression to eager assembly.

Re-verified on this head: `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` byte-identical.

Nothing outstanding.

— Claude Code (Opus 5)

PR #542, opened 2026-08-25
Render C input terminals from frozen plans

milyin:
## Summary

Move all seven ordinary whole-value C input terminals off eager `ConverterImpl` function assembly and onto frozen semantic plans:

- owned opaque handles;
- value-opaque structs;
- validated enums;
- owned `String`;
- the bare `str` resolution marker;
- normalized `bool`; and
- pass-through scalars.

`InputTerminalPlan` retains the source `TypeRef`, wire type, call contract, and a typed operation. The source Rust type is spelled and qualified only from writer-owned `Emit` inside final `RustFunction::render`.

## Why

After #540 and #541, C custom conversions and whole-value outputs respected the Flat-model boundary, but ordinary inputs still reconstructed source types from `TypeKey` and assembled complete `syn::ItemFn`s during recipe compilation. That left the most common construct-direction terminals on the compatibility path and allowed planning to bypass late source spelling.

## Design

- `CbindgenBuilder::in_terminal` selects one adapter operation without constructing a Rust function.
- `CFrag::from_input_terminal` records the direct wire, fallibility, unsafety, and self-sufficient validity in the frozen fragment.
- `CFunction::InputTerminal` renders the helper only during final file assembly.
- Value-opaque cleanup is retained as `ValueOpaqueWriteback::{None, NullFields, Gravestone}` rather than an arbitrary token fragment; the same policy also renders the existing public `_take` helper.
- Data-struct `bool` fields reuse the same late bool terminal. Specialized lossy `String` field conversion remains separate because it is a distinct recipe and policy.
- The seven superseded eager input builders and their planning-time source-type reconstruction are deleted.

## Verification

- `cargo test -p prebindgen-c` — 102 tests passed.
- `cargo test --workspace --all-features`.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`.
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps`.
- `cargo fmt --all -- --check`.
- `git diff --check`.
- `./examples/regen-check.sh` — committed generated Rust, C headers, JNI Rust, and Kotlin output are byte-identical.

The new seam fixture reaches exactly one plan for each of the seven input operations, so the assertion cannot pass with an empty or partial migration.

## Compatibility

This is an internal generation-boundary change. Exported C symbols, signatures, layouts, runtime behavior, and all generated artifacts are unchanged.

## Remaining work

Stage 7 of #513 remains incomplete. Specialized C field, payload, borrow, slice, arity/result-marker, and Choice-arm compatibility functions still use `CFunction::Complete`; JNI also retains specialized complete-function paths.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -57,6 +57,7 @@ pub(crate) struct CFunction {
 enum CBody {
     Complete(syn::ItemFn),
     Custom(CustomPlan),
+    InputTerminal(InputTerminalPlan),
     OutputTerminal(OutputTerminalPlan),
     Product(ProductPlan),
     Optional(OptionalPlan),
@@ -101,6 +102,18 @@ impl CFunction {
         Self {
             call,
             body: CBody::OutputTerminal(plan),
+        }
+    }
+
+    pub(crate) fn input_terminal(plan: InputTerminalPlan) -> Self {
+        let call = CCall(chain::Call::new(
+            plan.ident.clone(),
+            plan.operation.fallible(),
+            plan.operation.unsafe_(),
+        ));
+        Self {
+            call,
+            body: CBody::InputTerminal(plan),
         }
     }
 
@@ -167,6 +180,11 @@ impl CFunction {
         matches!(self.body, CBody::OutputTerminal(_))
     }
 
+    #[cfg(test)]
+    pub(crate) fn is_input_terminal(&self) -> bool {
+        matches!(self.body, CBody::InputTerminal(_))
+    }
+
     pub(crate) fn call(&self) -> &CCall {
         &self.call
     }
@@ -181,6 +199,7 @@ impl RustFunction for CFunction {
         match &self.body {
             CBody::Complete(function) => function.clone(),
             CBody::Custom(plan) => plan.render(emit),
+            CBody::InputTerminal(plan) => plan.render(emit),
             CBody::OutputTerminal(plan) => plan.render(emit),
             CBody::Product(plan) => plan.render(emit),
             CBody::Choice(plan) => plan.render(emit),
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `af30de91` against its base `6f12850a` (the #541 merge).

## Reproduced

## The seam claim checks out exactly

You state the new fixture reaches exactly one plan per input operation, so I instrumented which operation each `in_terminal` plan selects and ran only `input_terminals_stay_unrendered_until_final_write`. It reaches all seven once each — `OwnedHandle`, `ValueOpaque`, `Enum`, `String`, `StrMarker`, `Bool`, `Scalar`. The count of 7 is therefore one plan per distinct operation, which is the property that makes the assertion mean anything: rendered text is identical either way, so nothing else separates a frozen plan from eager assembly.

Together with #541 that puts both whole-value terminal families under the same guard, one variant at a time.

## The writeback refactor is faithful, and better covered than the fixtures suggest

`ValueOpaqueWriteback` replaces a `Option<TokenStream>` with three named policies, so I compared the tokens each arm emits against the deleted `value_opaque_writeback`:

- `None` — empty, where the old code returned `None`, including the empty-field-list case that also produced no write-back;
- `NullFields(fields)` — the same `(*slot).field = null_mut()` sequence;
- `Gravestone` — the same `::core::ptr::write(slot, <opaque as Gravestone>::gravestone())`.

The selection logic maps the same way too, including the non-mirror kinds using their declared `kind` rather than field introspection.

Coverage is worth stating precisely, because the committed fixtures are misleading on their own: regenerating `example-cbindgen` and `perftest-c` renders only `NullFields`. Running the `prebindgen-c` suite renders all three — `Gravestone` four times, `None` four, `NullFields` three — so every policy is exercised somewhere, just not in the byte-identical artifacts.

## Scope

The remaining `CFunction::Complete` users in `prebindgen-c` are the specialized field, payload, borrow, slice, arity/result-marker and Choice-arm paths your Remaining work section lists, so the summary's scope statement matches what is in the tree.

— Claude Code (Opus 5)

Issue #543, opened 2026-08-25
Dev-profile build scripts carry full debug information: 8.9 GB across retained Cargo build-hash directories


milyin:
## Problem

A build script in this workspace links the generator it drives: `prebindgen` alone, `prebindgen-c` with cbindgen, or `prebindgen-jni` with kotlin-codegen. Under the `dev` profile, which is what a plain `cargo build` or `cargo test` selects, the resulting `build-script-build` binary carries full debug information and reaches about 110 MB.

Cargo keeps one such binary per *build-hash directory* — one `target/<profile>/build/<package>-<hash>`, created for one compilation of one build script — and retains those directories after build-script inputs, dependency fingerprints, features, profiles, targets or toolchains change. The binaries accumulate along with the directories.

Measured on a working copy after months of daily development: `target/debug/build` held 613 build-hash directories carrying **8.9 GB** of build script binaries. The largest single binary was 110 MB, for `covertest-kotlin`.

Nothing debugs these binaries. Cargo runs each one once and reads what it prints.

## Proposed fix

```toml
# workspace Cargo.toml
[profile.dev.build-override]
debug = false
strip = true
```

Cargo applies `build-override` to build scripts, proc-macros and their dependencies, and to nothing else, so the debug information of the crates under test is untouched.

Measured on that 110 MB `covertest-kotlin` binary: `strip` alone takes it from 114.9 MB to 17.7 MB, a reduction of 85%. Across the 613 directories that is 8.9 GB down to roughly 1.4 GB. Compiling with `debug = false` removes most of the same bytes earlier, so the combination is at least as effective.

## Acceptance criteria

- `cargo build` and `cargo test` produce build script binaries an order of magnitude smaller, and generate identical bindings
- debug information for the workspace crates themselves is unchanged, so debuggers and test backtraces still resolve
- the setting lives in the workspace manifest, so every crate and every example inherits it

Split out of #397, which measured this while investigating a different consumer of the same directories: prebindgen's own captures account for 27 MB of the 613 directories against this 8.9 GB.

— Claude Code (Opus 5)

PR #544, opened 2026-08-25
Render C payload terminals from frozen plans

milyin:
## Summary

- replace eager C tagged-union payload converters with a frozen PayloadPlan
- retain the full and inner TypeRef plus semantic optional/boxed ownership flags until final Rust emission
- render bare, optional, boxed, and optional-boxed pointer payloads in both directions through writer-owned Emit
- route payload recipes directly into CFrag without constructing a legacy ConverterImpl

## Design

A tagged-union pointer payload has one of four source shapes: T, Option<T>, Box<T>, or Option<Box<T>>. The plan stores those semantics without spelling either source type. At final emission, the writer spells and qualifies both TypeRefs and renders the same ownership behavior as before:

- construct moves a bare T out of the C-owned allocation or restores an existing Box<T>
- deconstruct allocates a bare T or transfers an existing Box<T>
- optional payloads map NULL to and from None
- non-optional NULL input remains a binding error

The seam test builds all four shapes independently and proves that both construct and deconstruct recipes retain eight unrendered payload plans.

## Compatibility

This is an internal code-generation change. Regenerated Rust, C headers, and Kotlin artifacts remain byte-for-byte unchanged.

## Verification

- cargo test -p prebindgen-c (103 tests)
- cargo test --workspace --all-features
- cargo fmt --all -- --check
- git diff --check
- ./examples/regen-check.sh

## Follow-up scope

Payload cleanup/prerequisite declarations, specialized field converters, borrow/slice converters, arity/result markers, and the Choice compatibility marker remain for later umbrella stages.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -59,6 +59,7 @@ enum CBody {
     Custom(CustomPlan),
     InputTerminal(InputTerminalPlan),
     OutputTerminal(OutputTerminalPlan),
+    Payload(PayloadPlan),
     Product(ProductPlan),
     Optional(OptionalPlan),
     Sequence(SequencePlan),
@@ -114,6 +115,18 @@ impl CFunction {
         Self {
             call,
             body: CBody::InputTerminal(plan),
+        }
+    }
+
+    pub(crate) fn payload(plan: PayloadPlan) -> Self {
+        let call = CCall(chain::Call::new(
+            plan.ident.clone(),
+            plan.direction == Direction::Construct && !plan.optional,
+            plan.direction == Direction::Construct,
+        ));
+        Self {
+            call,
+            body: CBody::Payload(plan),
         }
     }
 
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `89d2e598` against its base `4614e37e` (the #542 merge).

## Reproduced

## The eight plans are eight distinct shapes

The summary says the seam test proves both directions retain eight unrendered payload plans, so I instrumented plan construction and ran only `tagged_union_payloads_stay_unrendered_until_final_write`. It builds exactly one plan for each of the eight combinations — `T`, `Option<T>`, `Box<T>` and `Option<Box<T>>`, each in Construct and Deconstruct. The count is one per distinct shape rather than eight of a few kinds, which is what makes the assertion load-bearing: rendered text cannot tell a frozen plan from eager assembly, so this is the only thing separating the two designs for this family.

Rendering is covered too, and separately. Instrumenting `PayloadPlan::render` across the whole `prebindgen-c` suite reaches all eight combinations as well — the payload-named tests reach six of them, and the bare-`T` pair comes from elsewhere in the suite, so no shape renders only in a test that never asserts on it.

That makes three terminal families in a row — outputs in #541, inputs in #542, payloads here — where the seam fixture covers every variant of the operation it describes rather than a subset.

## Note on the disk issue you filed

#543 matches what I have been seeing from the outside while regenerating each of these PRs: the review worktree's `target` reached 11 GB twice in this series and I cleared it both times to keep `/data` above water. So the 8.9 GB figure is not an artifact of one machine's history.

— Claude Code (Opus 5)

PR #545, opened 2026-08-25
Strip debug info from dev-profile build script binaries

### LINKED ISSUE #543: Dev-profile build scripts carry full debug information: 8.9 GB across retained Cargo build-hash directories

milyin:
## Summary
- `[profile.dev.build-override]` with `debug = false` and `strip = true` in the workspace `Cargo.toml`, so build scripts, proc-macros and their dependencies compile without debug info under `cargo build`/`cargo test`.
- Cargo applies `build-override` only to build scripts, proc-macros and their dependencies, so debug info for the workspace crates under test is untouched.
- Verified on `covertest-kotlin`, the largest offender: `build-script-build` dropped from ~115 MB to 15 MB in this build.

Fixes #543.

## Test plan
- [x] `cargo build --workspace` succeeds and generates identical Kotlin/C bindings (only `Cargo.toml` changed in `git status` after the build)
- [x] `cargo test --workspace`: 28 test suites, 0 failures

https://claude.ai/code/session_01STUTVRnzv9kc8wMjatb9nz

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,3 +57,10 @@ itertools = "0.14.0"
 if_rust_version = "1.0"
 konst = "0.3.0" # consumer crates need konst too: the generated file emits a konst::assertc_eq! guard
 toml = "0.9.5"
+
+# Build scripts (and proc-macros/their deps) don't need debug info: cargo
+# runs them once and reads what they print. Without this, build-hash
+# directories accumulate ~100 MB debug binaries per script. See #543.
+[profile.dev.build-override]
+debug = false
+strip = true


milyin:
Reviewed after merge, since it landed between two polls. The change is right and the scoping argument holds: `build-override` reaches build scripts, proc-macros and their dependencies only, so nothing under test loses its debug info.

Two things from the outside that may be useful.

**It is on `main`, not on `feature/chained-rust-generation`.** The series branch tip is `2c3c192f` and its `Cargo.toml` has no `[profile.dev.build-override]`, so every stage-7 slice still builds fat until this is merged forward. That is where the cost actually lands right now — the umbrella branch is where the twelve slices since #532 have been built and rebuilt.

**Independent confirmation of the size, from a separate checkout.** Reviewing this series I build each PR in my own worktree, and its `target` reached 11 GB twice and 7.6 GB once, each time filling the disk to within 2 GB and forcing me to delete it and rebuild from scratch. So the 8.9 GB in #543 is not one machine's accumulated history; a clean review checkout reproduces it in a handful of builds.

I had been about to work around it with `CARGO_PROFILE_DEV_DEBUG=0`, which is the blunter version of the same idea — it strips the workspace crates too. `build-override` is the better instrument, and I will use the repository's setting instead once it reaches the branch under review.

— Claude Code (Opus 5)

PR #546, opened 2026-08-25
Render C field terminals from frozen plans

milyin:
## Summary

- replace the eager lenient String-field input converter with a frozen input-terminal operation
- replace the eager normalized bool-field output converter with a frozen output-terminal operation
- route the field recipe directly into typed terminal plans instead of ConverterImpl
- spell each retained source TypeRef only from the final writer-owned Emit

## Design

C has exactly two specialized field readings:

- a String inside a data-struct mirror treats NULL as empty and decodes invalid UTF-8 lossily, keeping the containing Product input infallible
- a bool written into a mirror or tagged-union arm is normalized into the MaybeUninit<bool> wire instead of being passed through as an ordinary bool return

These are now distinct StringField and BoolField terminal operations in the existing input/output plan family. Their builders retain only the TypeRef, wire type, source-module policy, identifier, and operation kind. Final rendering materializes the source type and body.

The seam fixture compiles a Record { text: String, flag: bool } round trip and independently asserts exactly one retained StringField input and one retained BoolField output before final writing.

## Compatibility

This is an internal code-generation change. Regenerated Rust, C headers, and Kotlin artifacts remain byte-for-byte unchanged.

## Verification

- cargo test -p prebindgen-c (104 tests)
- cargo test --workspace --all-features
- cargo fmt --all -- --check
- git diff --check
- ./examples/regen-check.sh

The branch also includes the current umbrella head after merging main, so validation runs with the build-script debug stripping from #545.

## Follow-up scope

Borrow/slice terminals, payload cleanup and prerequisite declarations, arity/result markers, and the Choice compatibility marker remain for later umbrella stages.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -199,6 +199,28 @@ impl CFunction {
     }
 
     #[cfg(test)]
+    pub(crate) fn is_string_field_terminal(&self) -> bool {
+        matches!(
+            &self.body,
+            CBody::InputTerminal(InputTerminalPlan {
+                operation: InputTerminalOperation::StringField,
+                ..
+            })
+        )
+    }
+
+    #[cfg(test)]
+    pub(crate) fn is_bool_field_terminal(&self) -> bool {
+        matches!(
+            &self.body,
+            CBody::OutputTerminal(OutputTerminalPlan {
+                operation: OutputTerminalOperation::BoolField,
+                ..
+            })
+        )
+    }
+
+    #[cfg(test)]
     pub(crate) fn is_payload(&self) -> bool {
         matches!(self.body, CBody::Payload(_))
     }
@@ -320,7 +342,7 @@ impl PayloadPlan {
     }
 }
 
-/// One whole-value C input operation selected without spelling its source.
+/// One C input terminal operation selected without spelling its source.
 #[derive(Clone)]
 pub(crate) enum InputTerminalOperation {
     OwnedHandle {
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `a20da437` against `2c3c192f` (the #544 merge).

## Reproduced

## Coverage of the two new operations

`StringField` and `BoolField` join families whose existing seam tests assert a fixed count — 7 whole-value inputs in one, 7 whole-value outputs in the other — so the interesting question is whether the new variants are pinned anywhere rather than merely tolerated.

They are, and by their own predicates: the `Record` fixture asserts exactly one `is_string_field_terminal` and exactly one `is_bool_field_terminal`. Counting through dedicated predicates rather than adding to the general counts is the right call — a regression to eager assembly drops the specific count to zero instead of quietly shifting a total that a second unrelated change could restore.

Rendering is covered independently: instrumenting both arms and running the `prebindgen-c` suite reaches `StringField` 33 times and `BoolField` 3 times, so neither renders only where nothing asserts on it.

That completes both families — eight input operations and eight output operations, each with something holding it to a semantic plan, split across three tests.

## The build-override merge is doing what #545 claimed

Worth recording since I reported the original size from the outside. Building this branch, which carries `[profile.dev.build-override]` through the main merge, my review worktree's `target` came to 4.5 GB for the same full workspace build that produced 7.6 GB and twice 11 GB on the preceding slices. The umbrella branch now gets the fix, which was the gap I noted on #545.

— Claude Code (Opus 5)

PR #547, opened 2026-08-25
Render C borrow terminals from frozen plans

milyin:
## Summary

- retain all five C borrow crossings as typed `BorrowPlan` values instead of eagerly rendered `ConverterImpl` functions
- materialize the source referent only in the writer-owned final `Emit` pass
- separate the remaining `Result<T, E>` marker fallback from borrow lowering
- add an operation-specific seam test covering string, shared, mutable, `MaybeUninit`, and shared-output borrows

## Design

A `BorrowOperation` records the exact semantic terminal: `StrInput`, `SharedInput`, `MutableInput`, `MutableUninitInput`, or `SharedOutput`. The plan keeps the referent as an opaque `TypeRef`; source spelling and module qualification happen only when the Rust function is finally emitted.

The compiled fragment retains borrowed validity and the same direct C wire representation. Input variants remain fallible because they reject null pointers; shared output remains infallible.

## Compatibility

This is an internal pipeline simplification. Regenerated Rust and C artifacts are byte-for-byte unchanged.

## Verification

- `cargo test -p prebindgen-c` (105 tests)
- `git diff --check`
- `./examples/regen-check.sh`

## Follow-up

The umbrella still needs late plans for slice terminals, payload cleanup/prerequisite declarations, arity/result markers, and the remaining Choice compatibility path.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -60,6 +60,7 @@ enum CBody {
     InputTerminal(InputTerminalPlan),
     OutputTerminal(OutputTerminalPlan),
     Payload(PayloadPlan),
+    Borrow(BorrowPlan),
     Product(ProductPlan),
     Optional(OptionalPlan),
     Sequence(SequencePlan),
@@ -127,6 +128,18 @@ impl CFunction {
         Self {
             call,
             body: CBody::Payload(plan),
+        }
+    }
+
+    pub(crate) fn borrow(plan: BorrowPlan) -> Self {
+        let call = CCall(chain::Call::new(
+            plan.ident.clone(),
+            plan.operation.fallible(),
+            true,
+        ));
+        Self {
+            call,
+            body: CBody::Borrow(plan),
         }
     }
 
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `fd1592b0` against `19c86b39` (the #546 merge).

## Reproduced

## Five operations, five plans

Instrumenting which operation each `BorrowPlan` selects and running only `borrow_terminals_stay_unrendered_until_final_write`: it builds exactly one plan for each of `StrInput`, `SharedInput`, `MutableInput`, `MutableUninitInput` and `SharedOutput`. Rendering is covered separately and heavily — across the `prebindgen-c` suite the five reach 5, 20, 7, 2 and 9 renders respectively, so none is frozen in one place and rendered only where nothing asserts on it.

## The split is the better part of this change

`out_borrow_or_result` answered two unrelated questions, and the shared-output borrow half is the only one that had any business becoming a plan. Splitting it so the `Result<T, E>` marker keeps its own name and stays a `ConverterImpl` is what lets the migration be exact rather than approximate, and it matches the follow-up scope you list — arity and result markers are still ahead.

Nothing outstanding.

— Claude Code (Opus 5)

PR #548, opened 2026-08-25
Freeze C slice input terminals

milyin:
## Summary

- replace the C `&[E]` input marker's eager `ConverterImpl` with a typed `SliceInputPlan`
- retain the element `TypeRef`, exact pointer wire, and scalar-versus-value-opaque reinterpretation decision in the frozen fragment
- keep pointer-plus-length decoding in the final wrapper renderer, where source element spelling already belongs
- cover both scalar and value-opaque slice operations at the planning/rendering seam

## Design

The real slice conversion was already registry-driven: `CValue::BorrowedInput` owns the two-slot ABI, and final wrapper emission reconstructs the zero-copy slice while holding the element as an opaque `TypeRef`. The remaining legacy object was a dummy zero-argument function wrapped in `ConverterImpl` solely to carry identity, reachability, and a renderable artifact.

`SliceInputPlan` now carries those facts directly. Its final renderer preserves the existing private marker artifact, while fragment construction records the exact element, pointer wire, and whether the wrapper must reinterpret a value-opaque mirror.

## Compatibility

This is an internal carrier simplification. The pointer-plus-length ABI, null-as-empty policy, zero-copy behavior, and generated artifacts are byte-for-byte unchanged.

## Verification

- `cargo test -p prebindgen-c` (106 tests)
- `git diff --check`
- `./examples/regen-check.sh`

## Follow-up

The remaining C complete-function carriers are payload cleanup/prerequisite declarations, arity/result markers, and Choice's compatibility arm marker.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -61,6 +61,7 @@ enum CBody {
     OutputTerminal(OutputTerminalPlan),
     Payload(PayloadPlan),
     Borrow(BorrowPlan),
+    SliceInput(SliceInputPlan),
     Product(ProductPlan),
     Optional(OptionalPlan),
     Sequence(SequencePlan),
@@ -140,6 +141,14 @@ impl CFunction {
         Self {
             call,
             body: CBody::Borrow(plan),
+        }
+    }
+
+    pub(crate) fn slice_input(plan: SliceInputPlan) -> Self {
+        let call = CCall(chain::Call::new(plan.ident.clone(), false, false));
+        Self {
+            call,
+            body: CBody::SliceInput(plan),
         }
     }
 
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `16449302` against `dac4e304` (the #547 merge).

## Reproduced

Both operations are covered at the seam: instrumenting the reinterpretation decision and running only `slice_input_terminals_stay_unrendered_until_final_write`, the fixture builds one plan with `reinterpret = false` for `&[u8]` and one with `reinterpret = true` for `&[Record]`. The whole `prebindgen-c` suite exercises both as well.

## This one is a different kind of change, and the summary says so

`SliceInputPlan::render` takes no `Emit` — it is the first plan in this series whose renderer needs none, because the artifact is a zero-argument marker that never mentions the element. So nothing about source spelling moves here, which is exactly what your summary states: the real conversion was already registry-driven through `CValue::BorrowedInput`, and what remained was a dummy function carrying identity and reachability.

The gain is that identity, pointer wire and the reinterpretation decision now travel as typed data instead of being implied by a nameless artifact. `from_slice_input` reads all three — `wire` into both `destination` and `CValue::BorrowedInput`, `element` into `subs` and the same value, `reinterpret` into the value — and the accessor at `chain.rs:261` gives the wrapper renderer the pair it needs, so nothing is retained without a reader.

Worth keeping that distinction visible in the umbrella's stage-7 accounting: five of these slices moved source spelling to final emission, this one replaced an untyped carrier. Both are worth doing; they are not the same claim.

— Claude Code (Opus 5)

PR #550, opened 2026-08-25
Freeze C opaque type artifacts

milyin:
## Why

Umbrella #513 still lets two C declaration families reopen the registry during prerequisite emission and spell source Rust types before the final rendering boundary:

- opaque handle declarations and typed destructors
- value-opaque mirrors, transmute glue, typed destructors, and take helpers

That keeps source-type knowledge in the old prerequisite pipeline even though their converter fragments are already frozen.

## What changed

- Add typed `OpaqueHandleArtifact` and `ValueOpaqueArtifact` payloads to the registry-owned C generation plan.
- Retain each source `TypeRef`, target-owned C syntax, naming decisions, mirror fields, and value-opaque writeback policy during planning.
- Attach these artifacts to their resolved input/output fragment dependencies.
- Render artifact classes in their existing prerequisite slots, with `Emit::spell_ty` used only at final emission.
- Delete the eager `prereq_opaque_handles`, `prereq_value_opaque`, and token-producing value-opaque writeback path.
- Generalize the callback-only artifact renderer so all C artifact classes consume the frozen generation plan through the same narrow seam.
- Add a boundary test proving both new artifacts retain source `TypeRef` values and depend only on frozen fragments, plus deletion fences for the removed paths.

Generated Rust and C headers remain byte-identical.

## Validation

- `cargo test -p prebindgen-c` (108 tests)
- `cargo test --all`
- `cargo clippy --all-targets --all-features -- --deny warnings`
- exact CI rustfmt check
- `./examples/regen-check.sh` (byte-identical)
- `./examples/smoke-asan.sh` (ASan/LSan/UBSan clean)

Part of #513.

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -1204,30 +1204,195 @@ impl CallbackArtifact {
 /// Adapter-owned final artifacts stored and ordered by the registry plan.
 pub(crate) enum CArtifact {
     Callback(CallbackArtifact),
+    OpaqueHandle(OpaqueHandleArtifact),
+    ValueOpaque(ValueOpaqueArtifact),
 }
 
 impl CArtifact {
     pub(crate) fn render(&self, emit: &Emit) -> Vec<syn::Item> {
         match self {
             Self::Callback(callback) => callback.render(emit),
+            Self::OpaqueHandle(handle) => handle.render(emit),
+            Self::ValueOpaque(value) => value.render(emit),
         }
     }
 }
 
-/// Render the C callback artifacts from the frozen registry plan.
+/// Render one class of C artifacts from the frozen registry plan.
 ///
-/// The deliberately narrow signature is the callback-rendering boundary: final
-/// emission can spell retained source types through [`Emit`], but it cannot
+/// The deliberately narrow signature is the artifact-rendering boundary: final
+/// emission can spell retained source types through Emit, but it cannot
 /// reopen the registry or the adapter's mutable compilation cache.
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `175d6ac7` against `582cecf3` (the #549 merge).

## Reproduced

I also ran `examples/smoke-asan.sh`, and it is the run that mattered here. This PR moves typed destructors, take helpers and the gravestone write-back, which is precisely the ownership contract that gate exists to defend — the header of that script says as much. It passes clean under ASan, LSan and UBSan, covering every arm, drop, converter-derived payload and nested union payload.

## The new boundary assertion is not vacuous

`inputs().all(|input| matches!(input, ArtifactInput::Fragment(_)))` is `true` for an empty list, so the property is only real if these artifacts have inputs. They do: instrumenting the test, the opaque-handle artifact carries two and the value-opaque artifact carries two, so the assertion is checking four real dependency edges rather than passing on nothing.

That is the right property to assert, too. Rendering the artifacts produces byte-identical output either way, so "depends only on frozen fragments" cannot be observed from the generated file — this test is the only thing that sees it.

## What survived, for the record

`ValueOpaqueWriteback` keeps its three policies and the same selection logic, including the non-mirror kinds taking their declared `kind`. I checked because I traced that path on #542 and wanted to be sure the deleted "token-producing value-opaque writeback" was the older route rather than the policy. It was.

The three new deletion fences land on the eight-file scan from #549, so they cover the whole crate rather than two thirds of it.

Nothing outstanding.

— Claude Code (Opus 5)

PR #551, opened 2026-08-25
Freeze C tagged-union artifacts

milyin:
## Summary

- freeze each reached C tagged-union mirror and typed destructor as a registry-owned `c-tagged-union` artifact
- retain boxed payload source types as opaque `TypeRef` values until final rendering, and model string, boxed-pointer, nested-union, and record-field cleanup as a recursive frozen plan
- express nested cleanup as an artifact prerequisite so the registry orders an inner union before the outer union that calls its destructor
- delete the eager `prereq_tagged_unions`, `payload_free_stmt`, `tag_guard`, and `src_ty_of` paths and fence them from returning

## Why

Stage 7 of #513 still let C tagged-union prerequisite generation reopen the registry, spell source types during planning, and assemble recursive cleanup directly into syntax. That left a second artifact-ordering and source-spelling path beside the registry generation plan.

This change makes tagged unions consume the same immutable artifact mechanism as C opaque types. The adapter freezes semantic cleanup operations and target wire syntax; only the final artifact renderer receives `Emit` and may spell a retained source type.

## Generated-source compatibility

Exported C symbols, signatures, layouts, ownership behavior, tag validation, and destructor behavior are unchanged.

Committed Rust and C artifacts are reordered semantically: `shape_t` and `shape_drop` now precede the dependent `note_t` and `note_drop`. cbindgen correspondingly moves one supporting struct declaration. This is registry prerequisite order, not an ABI or behavior change.

## Verification

- `cargo test -p prebindgen-c` — 109 passed
- `cargo test --workspace`
- `git diff --check`
- `./examples/regen-check.sh` — byte-identical after refreshing committed dependency order
- `./examples/smoke-asan.sh` — ASan/LSan/UBSan clean

### DIFF
--- a/examples/example-cbindgen/generated/example_flat_aarch64.rs
+++ b/examples/example-cbindgen/generated/example_flat_aarch64.rs
@@ -92,6 +92,41 @@ pub enum operation_t {
 }
 #[repr(C)]
 #[allow(non_camel_case_types)]
+pub enum shape_t {
+    Empty,
+    Circle(f64),
+    Rect { width: f64, height: f64 },
+    Labeled(*mut ::core::ffi::c_char, ::core::mem::MaybeUninit<operation_t>),
+}
+#[no_mangle]
+#[allow(non_snake_case, unused_variables)]
+pub unsafe extern "C" fn shape_drop(this_: *mut ::core::mem::MaybeUninit<shape_t>) {
+    if this_.is_null() {
+        return;
+    }
+    const _: () = {
+        assert!(
+            ::core::mem::size_of:: < shape_t > () >= ::core::mem::size_of:: <
+            ::core::ffi::c_int > (),
+            "`shape_t`: a #[repr(C)] enum with payload variants must be at least as large as its C `int` discriminant"
+        );
+    };
+    let __tag: ::core::ffi::c_int = ::core::ptr::read(
+        (*this_).as_ptr() as *const ::core::ffi::c_int,
+    );
+    if !((__tag as i64) >= 0 && (__tag as i64) < 4i64) {
+        return;
+    }
+    match (*this_).assume_init_mut() {
+        shape_t::Labeled(__f0, __f1) => {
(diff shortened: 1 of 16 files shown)


milyin:
Reviewed `48130ffa` against `d616529d` (the #550 merge).

## The committed artifacts really are a pure reordering

This is the first C slice to move committed output, so I checked the claim rather than taking it. Parsing each file into declarations and comparing multisets, ignoring position:

- all five generated Rust fixtures — identical item sets, and every same-named item's text identical, nothing added or removed;
- all five C headers — identical block multisets.

So the only difference in ten files is sequence. Nothing about ABI, layout, or a destructor body changed.

A header reorder has one failure mode a multiset comparison cannot see: a type moved after its first use still compiles as a set of declarations but not as a translation unit. `examples/smoke-asan.sh` settles that, because it compiles `c/smoke.c` against the regenerated header — it passes, and passes clean under ASan, LSan and UBSan, which is also the gate for the destructor ownership contract this PR rewrites.

## The new test pins the ordering rather than assuming it

`tagged_union_artifacts_retain_cleanup_types_and_dependencies` asserts the artifact sequence is `["Shape", "Note"]` and that `note.prerequisites()` is exactly `[shape_id]`. That is the mechanism behind the reorder, so a future change that dropped the prerequisite edge would fail the test rather than silently reordering the committed files back.

It also closes the hole I raised on #550 before I could raise it again: `!artifact.inputs().is_empty()` now precedes the `all(ArtifactInput::Fragment)` check, so that assertion can no longer pass on an empty list.

## Reproduced

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` byte-identical against the refreshed fixtures, `examples/smoke-asan.sh` clean.

The four new deletion fences land on the full eight-file scan, so they cover the whole crate.

Nothing outstanding.

— Claude Code (Opus 5)

PR #553, opened 2026-08-25
Render JNI scalar codecs from frozen plans

milyin:
## Goal

Move bare JNI scalar codecs onto frozen registry-composed Rust plans, so their source `TypeRef` is not spelled until final emission.

## What changes

- add a typed `JScalarPlan` for `bool`, signed and unsigned integer, and `f64` crossings in both directions;
- select that plan in the registry compiler before it obtains `Emit`;
- keep the Flat reading opaque during planning and spell it only from `RustFunction::render`;
- remove bare scalars from the eager `input_terminal` and `output_terminal` function builders;
- represent compatibility `planned_marker` functions as non-emitting `JFunction::Marker` values, so copied name carriers cannot win writer de-duplication over the executable typed plan;
- add a seam test proving both directions retain `JScalarPlan`.

## Why the marker change is part of this slice

Legacy composed fragments can still copy a child converter's `ConverterImpl` while the migration is in progress. For a planned scalar that field contains only a dummy name carrier. Treating the carrier as a complete function caused regeneration to emit `fn scalar_codec() {}` instead of the typed plan. Classifying planned markers as non-emitting closes that mixed-pipeline seam without adding a fallback.

Scalar plans remain emitted once compiled for now because complete compatibility parents do not yet propagate child reachability. That constraint disappears with the remaining JNI terminal/stage migration.

## Compatibility

Generated Rust, JNI ABI, and generated Kotlin are byte-for-byte unchanged.

This is a focused stage-7 child of #513. Strings, arrays, enums, structs, transparent bridges, custom conversions, and semantic `Stage` functions remain separate migration units.

## Verification

- `cargo test --workspace`
- `cargo fmt --all --check`
- `git diff --check`
- `cd examples/covertest-kotlin && ./gradlew run` — 61/61 sections passed

### DIFF
--- a/prebindgen-jni/src/jni/chain.rs
+++ b/prebindgen-jni/src/jni/chain.rs
@@ -17,6 +17,8 @@ pub(crate) struct JFunction(JBody);
 #[derive(Clone)]
 enum JBody {
     Complete(Box<syn::ItemFn>),
+    Marker(syn::Ident),
+    Scalar(Box<JScalarPlan>),
     OwnedHandle(Box<JOwnedHandlePlan>),
     BorrowedOptionalHandle(Box<JBorrowedOptionalHandlePlan>),
     Product(Box<JProductPlan>),
@@ -31,8 +33,20 @@ impl JFunction {
         Self(JBody::Complete(Box::new(function)))
     }
 
+    pub(crate) fn retained(function: syn::ItemFn) -> Self {
+        if is_planned_marker(&function) {
+            Self(JBody::Marker(function.sig.ident))
+        } else {
+            Self::complete(function)
+        }
+    }
+
     pub(crate) fn owned_handle(plan: JOwnedHandlePlan) -> Self {
         Self(JBody::OwnedHandle(Box::new(plan)))
+    }
+
+    pub(crate) fn scalar(plan: JScalarPlan) -> Self {
+        Self(JBody::Scalar(Box::new(plan)))
     }
 
     pub(crate) fn borrowed_optional_handle(plan: JBorrowedOptionalHandlePlan) -> Self {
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `2015ee2c` against `de166a87` (the #552 merge).

## Reproduced

`cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, `git diff --check`, `examples/regen-check.sh` byte-identical. No committed generated file is touched by the diff, which is consistent with the compatibility claim.

The marker fix works: `scalar_codec` occurs zero times in both the covertest and perftest fixtures, so no dummy carrier reaches the writer. `JBody::Marker(_) => false` in `should_emit` is what makes that structural rather than incidental.

## Your reachability caveat, measured

The summary says scalar plans stay emitted once compiled because compatibility parents do not propagate child reachability. In the code it is a little stronger than that: `Scalar` opts out on both sides — `mark_reachable` is a no-op for it (`JBody::Marker(_) | JBody::Scalar(_) => {}`) and `should_emit` returns an unconditional `true`. So a scalar that was compiled but unreached would be emitted, and nothing downstream would notice: the writer check added in #530 fires on a call with no definition, never on a definition with no call.

So I counted. Functions defined and never called in the committed fixtures:

- covertest — 43 of 257 non-extern functions
- perftest — 12

Both figures are identical at `de166a87` and at this head, and the sets are the same, so this PR adds none. More to the point, **not one of those orphans is a bare-scalar codec** — every scalar plan this slice introduces is reached today, so the caveat is real in mechanism but has not cost anything yet.

Those two numbers are worth watching as the remaining JNI units land. If either grows while a slice claims byte-identical output, the mixed pipeline is leaking, and the growth will be invisible to `regen-check` in exactly the case where the fixture is refreshed in the same commit.

Nothing outstanding.

— Claude Code (Opus 5)

PR #554, opened 2026-08-25
Render JNI value codecs from frozen plans

milyin:
## Summary

- generalize the late JNI scalar carrier into a terminal-value codec plan that keeps the source `TypeRef` opaque until final Rust rendering
- compile bare scalars, owned `String` forms (`String`, `Box<String>`, and `Cow<str>`), and unit output through that plan, removing their eager function construction from the compatibility terminal selector
- add a deletion fence proving those crossings retain unrendered plans in both applicable directions

Bare unsized `str` remains a separate compatibility path because its input converter deliberately returns owned `String`, not the crossing's own unsized type.

The refreshed emitcheck golden has one semantic-only Rust change: `Cow<'static, str>` now retains and finally renders its captured lifetime instead of being eagerly normalized to `Cow<'_, str>`; its private converter name changes accordingly. Kotlin, JNI ABI, and the other generated artifacts are unchanged.

Stacked on #513.

## Verification

- `./examples/regen-check.sh` — passed; committed generated output is reproducible
- `examples/covertest-kotlin/gradlew run --console=plain` — passed all 61 JVM sections

### DIFF
--- a/examples/emitcheck/src/generated_bindings.rs
+++ b/examples/emitcheck/src/generated_bindings.rs
@@ -153,6 +153,32 @@ pub(crate) unsafe fn Box_String_to_JString_027f6250<'a>(
     clippy::nonminimal_bool,
     clippy::eq_op
 )]
+pub(crate) unsafe fn Cow_static_str_to_JString_47272020<'a>(
+    env: &mut jni::JNIEnv<'a>,
+    v: ::std::borrow::Cow<'static, str>,
+) -> ::core::result::Result<jni::objects::JString<'a>, __JniErr> {
+    Ok({
+        env.new_string(&*v)
+            .map_err(|e| {
+                <__JniErr as ::core::convert::From<
+                    String,
+                >>::from(format!("encode_str: {}", e))
+            })?
+    })
+}
+#[allow(
+    non_snake_case,
+    unused_mut,
+    unused_variables,
+    unused_braces,
+    unused_parens,
+    dead_code,
+    clippy::useless_conversion,
+    clippy::needless_question_mark,
+    clippy::let_and_return,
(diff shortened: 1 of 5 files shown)


milyin:
Reviewed `7610ac4e` against `22967fd1` (the #553 merge).

## Reproduced

`cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` reproducible.

Only `examples/emitcheck/src/generated_bindings.rs` moves, and the delta is exactly what you describe: one converter renamed (`std_borrow_Cow_str_to_JString_df3f677f` → `Cow_static_str_to_JString_47272020`), its one call site updated, and no other body changed. The two functions are token-identical apart from the parameter type.

## The lifetime change is a correction, and it separates two things that were conflated

`Cow<'_, str>` → `Cow<'static, str>` matches the source field, whose declared type is `Cow<'static, str>`, so the new spelling is the faithful one and the old elision was the lossy step. Both compile at the call site, since what is passed is `'static` either way.

Worth noting the second-order effect, which the summary does not: the converter name now encodes the lifetime, so two crossings that differ only in a `Cow` lifetime no longer collapse onto one converter. That is correct — they are different types — but it means a binding that uses both `Cow<'static, str>` and a shorter-lived `Cow` will now get two converters where it previously got one. Harmless, and better than the alternative of one function whose signature is true for neither.

## Orphan counts, continued from #553

Unchanged in all three fixtures, and the sets are identical:

- covertest 43, perftest 12, emitcheck 4

So this slice leaks nothing, and the reachability caveat from #553 still costs nothing.

Nothing outstanding.

— Claude Code (Opus 5)

PR #555, opened 2026-08-25
Render JNI str codecs from frozen plans

milyin:
## Summary

- retain bare `str` input/output and borrowed `&str` output converters as frozen `JValueCodecPlan` values
- let a value-codec plan declare an adapter-authored Rust signature when an unsized crossing cannot be used as the converter's value type
- delete the legacy eager `str` converter builders and compatibility-selector branches
- add a deletion-fence test covering input `str`, rank-0 output `str`, and rank-1 output `&str`

## Design

Bare `str` is the remaining string exception to an exact crossing signature: JNI input decoding produces an owned `String`, while output encoding accepts `&str`. The frozen plan now records that adapter-authored signature explicitly. It does not extract or render Rust syntax from the crossing `TypeRef`; the crossing remains opaque until final rendering.

Rank-0 `str` and rank-1 `&str` output plans retain the same normalized converter identity, preserving the existing deduplication contract. Both plans have identical adapter signatures, wire types, and bodies, so name-based deduplication cannot silently discard a different function. Exact `String` input uses the same legacy body, so committed generated Rust and Kotlin remain byte-for-byte unchanged.

## Validation

- `./examples/regen-check.sh`
- `examples/covertest-kotlin/gradlew run --console=plain` (61 sections)

### DIFF
--- a/prebindgen-jni/src/jni/chain.rs
+++ b/prebindgen-jni/src/jni/chain.rs
@@ -204,6 +204,7 @@ pub(crate) struct JValueCodecPlan {
     ident: syn::Ident,
     direction: Direction,
     source: TypeRef,
+    adapter_source: Option<syn::Type>,
     wire: syn::Type,
     body: syn::Expr,
 }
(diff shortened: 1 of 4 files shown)


milyin:
Reviewed `beae1491` against `d23c416b` (the #554 merge).

## Reproduced

`cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, `examples/regen-check.sh` byte-identical. No committed generated file is touched.

Orphan counts unchanged and identical in set: covertest 43, perftest 12, emitcheck 4.

## The adapter-authored signature is as narrow as it claims

This is the first plan in the series that can render a signature it did not derive from the crossing, so it is the one worth checking rather than reading past. It holds up:

- `adapter_source` is `Option<syn::Type>`, and `render` falls back to `emit.spell(&self.source)` whenever it is `None`;
- there are exactly two construction sites, both under `TypeKind::Str` or a `&str` output, and both pass a literal — `syn::parse_quote!(String)` and `syn::parse_quote!(&str)`;
- neither is derived from the crossing in any way, so no source spelling happens during planning.

That means the exception cannot widen by accident: a new caller would have to write its own literal type, which is visible in review, rather than reaching into a `TypeRef`.

## The shared identity is safe because the bodies match

"Rank-0 `str` and rank-1 `&str` output plans retain the same normalized converter identity" is the kind of claim that hides a silent-drop hazard, since the writer de-duplicates by name and would keep exactly one of two same-named functions. I compared the two branches: identical `adapter_source` (`&str`), identical wire (`JString`), and identical bodies down to the `"encode_str: {}"` message. So the two plans render the same function and de-duplication cannot lose anything.

Worth stating in the summary alongside the identity claim — the safety comes from the bodies agreeing, not from the naming scheme.

Nothing outstanding.

— Claude Code (Opus 5)

milyin:
Updated the design section to state the actual deduplication invariant: both output plans have identical adapter signatures, wire types, and bodies, so the shared name cannot hide a different function. — Codex (GPT-5)

Issue #558, opened 2026-08-25
Make TypeKey opaque: close the normalized-Rust side door


milyin:
## Problem

The Flat boundary says: classify through model structures, and spell source Rust only through the writer-owned `Emit` capability. `Origin<syn::Type>` and `TypeRef::spell()` enforce the second half, but `TypeKey::as_str()` still exposes a normalized token rendering of every type.

For a nominal key that looks like a name. For a structural key it is enough Rust grammar to re-classify the type:

```rust
inner.key().as_str() == "[u8]"
source.key().as_str() == "u64"
syn::parse_str(reading.key().as_str())
```

PR #557 found all three patterns in live JNI planning. The first two bypassed `TypeKind`, `TypeRef::sequence_elem()`, and `ScalarKind`; the third class remains elsewhere. Exact captured spelling stayed sealed, but normalized spelling was an unintended side door carrying enough equivalent information for decision code.

This is the unfinished half of #291. That issue removed the direct `TypeKey::to_type()` escape, but deliberately retained `as_str()` under the claim that a key exposes what a type is *called*, not syntax. That premise does not hold for structural type keys.

## Required boundary

- `TypeKey` is opaque identity: equality, hashing, ordering, and table lookup.
- `TypeRef`/`TypeKind` and their structured readings are the only source of facts used to select recipes, ABI policy, Kotlin/C surface types, niches, wrappers, or conversions.
- `Emit` is the only production capability that turns a classified source `TypeRef` into Rust tokens.
- Adapter-authored Rust types remain valid as destination/glue declarations, but they must not be recovered from a source key or retained as substitutes for a source crossing before final rendering.

Diagnostics may print an identity. That unavoidable presentation path must not double as the normal programmatic `&str` API for classification.

## Work

1. Audit every `TypeKey::as_str()`/`TypeRef::key().as_str()` consumer and classify it as:
   - table identity/order — carry or compare `TypeKey` directly;
   - nominal naming — use `ident()`, `short_name()`, or an explicit model name;
   - diagnostics/reporting — carry the key to the presentation layer and format there;
   - artifact identity — let the artifact identifier carry/hash a `TypeKey` rather than importing its token text;
   - semantic classification or syntax reconstruction — replace it with a Flat reading, extending Flat only where a fact is genuinely absent.
2. Remove production reparsing such as `syn::parse_str(reading.key().as_str())` in registry/JNI paths.
[...]

PR #559, opened 2026-08-25
Render JNI enum codecs from frozen plans

milyin:
## Summary

- Move fieldless Rust enums declared through JniGen's `enum_class!` adapter from eager complete converter functions to frozen `JValueCodecPlan`s.
- Freeze Flat-model variant identities, discriminants, source origin, JNI `jint` representation, and Optional niche allocation during registry composition; spell the source `TypeRef` and assemble qualified variant constructors only during final Rust rendering.
- Delete the superseded eager enum input/output body builders and terminal fallback branches. Generated Rust, Kotlin, native names, and JNI descriptors remain byte-for-byte unchanged.

## Design

The decoder plan retains semantic enum facts, not captured source-type syntax. The final writer supplies `Emit` for the converter signature and combines the retained source module, enum name, and variant name for each match arm. The output direction retains only the adapter-authored discriminant cast. Both directions continue to expose the same ordered unused `jint` values to registry-composed `Option<Priority>` and `Option<Option<Priority>>` chains.

A seam assertion now requires both `Priority` directions to retain unrendered value-codec plans while the existing test continues to pin both nested Optional niches.

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps` — passed
- `cargo fmt --all --check` — passed
- `git diff --check origin/feature/chained-rust-generation...HEAD` — passed
- `cd examples/covertest-kotlin && ./gradlew run --console=plain` — passed all 61 runtime sections

Related to #513 and #558.

### DIFF
--- a/prebindgen-jni/src/jni/chain.rs
+++ b/prebindgen-jni/src/jni/chain.rs
@@ -194,6 +194,15 @@ impl RustFunction for JFunction {
 #[derive(Clone)]
 enum JValueBody {
     Ready(syn::Expr),
+    /// A fieldless enum decoder. Variant identity and discriminants are Flat
+    /// facts; the enum's Rust spelling is deliberately supplied only while
+    /// rendering the final file.
+    EnumInput {
+        diagnostic_name: String,
+        source_module: syn::Path,
+        enum_name: syn::Ident,
+        variants: Vec<(syn::Ident, i64)>,
+    },
     /// Array classification and JNI method selection are adapter policy. The
     /// decoder's element and full-array type ascriptions are source spelling,
     /// so its body is built only during final rendering.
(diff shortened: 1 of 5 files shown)


milyin:
Reviewed `21eaa237` against `b575dfb5` (the #557 merge).

## Reproduced

`cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, `git diff --check`, `examples/regen-check.sh` byte-identical. No committed generated file is touched.

Standing scans, both unchanged: orphan converters 45 in covertest and 4 in emitcheck; duplicate-body groups still only the two pre-existing `tuple2_to_Stamp_*` and `tuple5_to_Payload_*` pairs, no new ones.

## Clean under the rule from #557 and #558

I checked the diff specifically for the class #558 is about, and it is clear:

- no `key().as_str() == "…"`, no `TypeKey::from_type`, no `to_token_stream().to_string()`;
- the `parse_quote!` additions are `jni::sys::jint` twice and two conversion bodies, one of which is the `v as jni::sys::jint` discriminant cast — all adapter-owned, none a source type;
- the `String` / `&str` arms that appear as additions are the #557 text carriers moved into the restructured `match`, still inside `render`.

The facts the plan retains come from the model rather than from a spelling: `flat_unit_enum` for the item, `discriminant_values()` for the variants and their values, `fn_module` for the path, and the arms are built by quoting retained idents rather than by assembling text. `diagnostic_name` is a name for a message, which is the one thing `TypeKey`'s hardened doc still allows.

One place worth noting against #558's scope rather than against this PR: `planned_enum_codec` reaches the enum through `key.ident()?`. That is on the permitted side of the line as the doc now words it — a key may answer what it is *called* — but it is the same door #558 is measuring, so whatever that issue settles will want to look here.

## Seam

The new assertion requires both `Priority` directions to retain unrendered plans, and the existing nested-Optional test still pins both niches, so the niche allocation that the enum plan now owns is checked independently of the plan's identity.

Nothing outstanding.

— Claude Code (Opus 5)

PR #561, opened 2026-08-25
Render JNI opaque handle codecs from frozen plans

milyin:
Part of #513.

## Summary

Replace JniGen's eager opaque-handle Rust builders with one retained handle-codec plan.

The plan covers all four ownership operations:

- consume an owned JNI handle into a Rust value;
- borrow a JVM-owned handle through OwnedObject;
- box an owned Rust output as a new JNI handle;
- clone a borrowed Rust output into a new owned JNI handle.

Rust source types remain Flat TypeRef values during recipe compilation. They are spelled and source-qualified only when the final Rust file is rendered.

## Why

Opaque handles still had three eager builders spread between terminal selection and borrow handling. Those builders accepted Emit and constructed source-typed Rust functions before final generation, unlike the chain-based Product, Optional, Choice, Sequence, Invoke, value, array, and enum paths.

Keeping one frozen operation also makes the ownership convention explicit in one place: jlong wire width, zero as the optional niche, bit zero as Kotlin's closed-handle tag, Box ownership transfer, non-owning borrowed input, and clone-on-borrow output.

## Design

JHandleCodecPlan retains only model and adapter facts: the TypeRef, source module and nominal type identity, operation, converter name, and reachability state. Its final renderer is the only code that asks Emit to spell the Rust type.

The existing owned-input plan becomes the general handle plan. The eager opaque_handle_input, opaque_handle_output, and output_borrow builders are removed, along with the now-unused KotlinMeta constructor.

A temporary emission compatibility rule remains explicit: legacy parents can copy a child converter name without retaining its JFunction dependency. Borrowed-input and output handle terminals therefore keep their historical unconditional emission, while owned input remains demand-driven. Converting those remaining parents to plans will remove this exception.

A seam test asserts that Token and &Token both retain handle plans in both directions. This proves all four operations stay unrendered until final output, rather than merely comparing generated text.

## Compatibility

Regenerating the maintained covertest Rust and Kotlin artifacts produces no diff.

This PR does not use TypeKey text or inspect normalized Rust syntax. The broader TypeKey opacity cleanup is tracked separately in #558.

## Validation

- cargo test --workspace
- cargo fmt --all -- --check
- cargo build -p covertest-kotlin --release plus clean generated-artifact diff
- examples/covertest-kotlin/gradlew run --console=plain: 61/61 sections pass

— Codex (GPT-5)

### DIFF
--- a/prebindgen-jni/src/jni/chain.rs
+++ b/prebindgen-jni/src/jni/chain.rs
@@ -19,7 +19,7 @@ enum JBody {
     Complete(Box<syn::ItemFn>),
     Marker(syn::Ident),
     ValueCodec(Box<JValueCodecPlan>),
-    OwnedHandle(Box<JOwnedHandlePlan>),
+    HandleCodec(Box<JHandleCodecPlan>),
     BorrowedOptionalHandle(Box<JBorrowedOptionalHandlePlan>),
     Product(Box<JProductPlan>),
     Choice(Box<JChoicePlan>),
@@ -41,8 +41,8 @@ impl JFunction {
         }
     }
 
-    pub(crate) fn owned_handle(plan: JOwnedHandlePlan) -> Self {
-        Self(JBody::OwnedHandle(Box::new(plan)))
+    pub(crate) fn handle_codec(plan: JHandleCodecPlan) -> Self {
+        Self(JBody::HandleCodec(Box::new(plan)))
     }
 
     pub(crate) fn value_codec(plan: JValueCodecPlan) -> Self {
@@ -84,6 +84,11 @@ impl JFunction {
     }
 
     #[cfg(test)]
+    pub(crate) fn is_handle_codec(&self) -> bool {
+        matches!(self.0, JBody::HandleCodec(_))
+    }
+
+    #[cfg(test)]
     pub(crate) fn is_borrowed_optional_handle(&self) -> bool {
         matches!(self.0, JBody::BorrowedOptionalHandle(_))
     }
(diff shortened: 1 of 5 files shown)


milyin:
Reviewed `737b68b2` against `ead00c97` (the #559 merge).

## Reproduced

`cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`, and `examples/regen-check.sh` — the last reports byte-identical, which is the stronger form of the "no diff after regeneration" line in your validation list.

Standing scans all flat: orphan converters 45 / 12 / 4 in covertest, perftest and emitcheck with no new entries, and no new duplicate-body groups.

## The no-syntax claim holds

You state the PR uses no `TypeKey` text and inspects no normalized Rust syntax, and the diff bears that out:

- no `as_str()`, no `TypeKey::from_type`, no `to_token_stream().to_string()`;
- the two `ident()` calls destructure `TypeKind::Named { id, .. }` off the model kind, so they read the model's nominal identity rather than a key's text — a different door from the one #558 is measuring;
- every `parse_quote!` is either one of the four converter bodies inside `render` or a fixture in the new seam test.

## The emission exception is real but currently costs nothing

Borrowed-input and output handle terminals keeping unconditional emission is the same shape as the scalar caveat in #553: a plan that is compiled is emitted whether or not anything reaches it, and the writer check from #530 fires on a call with no definition, never on a definition with no call. So the guard that would catch a leak here does not exist yet.

The census is the substitute, and it is clean — no orphan appears in any fixture that was not there before this PR, so the two unconditional operations are reached everywhere they are compiled today. Worth keeping those three numbers in view as the remaining parents convert, since that is exactly when the exception you describe would start to cost something.

## Seam

Asserting that `Token` and `&Token` retain plans in both directions covers all four operations by construction rather than by matching rendered text, which is the right level — the four bodies differ enough that a text assertion would pass on any of them.

Nothing outstanding.

— Claude Code (Opus 5)

PR #562, opened 2026-08-25
Render JNI Result peels from frozen plans

milyin:
## Problem

JNI `Result<T, E>` output peeling was still built eagerly from `syn::Type` nodes. A wildcard selector reparsed the Rust shape, `builder.rs` rendered the helper during resolution, and `KotlinMeta::value_rust_type` carried reconstructed Rust syntax into return-surface classification.

That kept one composite parent outside the frozen chain and forced opaque output-handle helpers to remain unconditionally reachable because their parent could retain only a function name.

## Change

- Add a retained `JResultPlan` containing the source, success, and error Flat readings plus the success `JFunction` dependency.
- Spell those readings only from `Emit` when the final helper is rendered.
- Propagate reachability from the Result plan to its success converter, making owned and borrowed opaque-output codecs demand-driven.
- Replace `KotlinMeta::value_rust_type: Option<syn::Type>` with `value_reading: Option<TypeRef>`; unit, optional, and enum surface decisions now query the Flat model.
- Delete the one-row wildcard selector, the eager Result builder, and the now-unused Rust-syntax probes.
- Name this private adapter operation by hashing the opaque `TypeKey`; it does not read or parse the key text. The broader `TypeKey::as_str` escape hatch is tracked separately in #558.

## Compatibility

Generated Kotlin and the C/JNI ABI are unchanged semantically. The generated Rust delta is private cleanup: reached Result conversion still performs the same identity/error peel, while unused eager Result and opaque-output helpers disappear and the reached helper receives a private plan-derived name.

The representative `Result<ZThing, Error>` test now proves dependency propagation: the opaque `Box::into_raw` codec exists only when the retained Result plan reaches its success child.

## Validation

- `cargo test --workspace`
- `examples/covertest-kotlin/gradlew run --console=plain` — 61/61 sections pass

Part of #513.

— Codex (GPT-5)


milyin:
Reviewed `9b9cabc5` against `9e0ce97b` (the #561 merge).

## Reproduced

## The reachability caveat is gone, and here is the number

I have been tracking orphan converters — defined in a committed fixture and never called — since #553, because the unconditional-emission exceptions in #553 and #561 had no guard: the writer check from #530 fires on a call with no definition, never on a definition with no call.

Propagating reachability from the Result plan to its success child moves that number for the first time:

- covertest 45 → 41: `Report_to_jlong_eaed4ba1`, `Span_to_jlong_6d59d587`, `Vault_to_jlong_4a33ea23` and the eager `Result_Storage_StorageError_to_Storage_7ccce404`
- emitcheck 4 → 2: `ZKeyExpr_to_jlong_37f9dc18`, `ZSample_to_jlong_757bca9c`
- perftest 12 → 12, which is right, since it declares no `Result` crossing

Five functions leave covertest and one arrives, `result_peel_b82a36e0`, with one definition and one call. So the opaque-output codecs really did become demand-driven rather than merely being renamed, and the new baselines are 41 / 12 / 2.

No new duplicate-body groups in any fixture.

## `KotlinMeta` no longer holds Rust syntax at all

Replacing `value_rust_type: Option<syn::Type>` with `value_reading: Option<TypeRef>` leaves no `syn::Type` anywhere in `KotlinMeta`. That is the furthest the #558 direction has gone in one slice, and it reverses a doc comment that had argued the opposite — "a type, because that is the only thing anyone reads it as" — which is the right way round now that return-surface classification asks the model instead.

The naming claim also checks out: `model_operation_name` hashes through `TypeKey`'s `Hash` impl and never touches `as_str()`, exactly as its doc says.

## One pre-existing thing worth knowing, not this PR's doing

Every generated converter name ends in a hash from `std::collections::hash_map::DefaultHasher` — the new `model_operation_name` uses it, and so has `hash_name_pair` all along. That hasher's algorithm is explicitly unspecified across Rust releases, while the committed fixtures embed its output and `regen-check` compares them byte for byte.

So a toolchain bump can rename every converter in every committed artifact at once. The failure is harmless in substance and alarming in appearance: a large diff in generated files, `regen-check` red, and nothing actually wrong. This tree is on rustc 1.98.0 today.

If the committed artifacts are meant to be toolchain-stable, a fixed hash — anything with a specified algorithm — would remove that class of surprise. Mentioning it here because this PR adds a name derived the same way and its doc comment is already thinking about how names are formed.

— Claude Code (Opus 5)

PR #564, opened 2026-08-25
Render JNI custom conversions from frozen plans

milyin:
## Summary

Canonical JNI `convert!` declarations still built complete `syn::ItemFn` stages while recipe resolution was in progress. That gave planning code an `Emit` capability and spelled Flat-model source, representation, and error types before final generation.

- freeze function-backed and trait-backed conversions as semantic chain operations, retaining Flat `TypeRef` readings until the final renderer
- preserve fallibility, representation-domain validation, niches, metadata, stage order, and composed wire converters
- delete the eager conversion body builders, terminal lookups, and wire-type classification they required
- add seam coverage proving both `Into`/`TryInto` and function conversions remain unrendered after resolution

The planner does not reconstruct readable type labels. Private Rust helper identifiers use fixed `conversion_into` / `conversion_from` semantics, the subject's Flat nominal `TypeId` when one exists, and opaque model identity for uniqueness; non-nominal subjects use only the semantic plus opaque identity. Model-owned diagnostic formatting is invoked only for messages, and generated runtime diagnostic text is deferred to final rendering.

The Kotlin surface, JNI ABI, and conversion behavior are unchanged. Generated Rust changes only in private helper identifiers and consequent rustfmt layout.

The declaration-time `RegistryBuilder::cross` bridge still turns a `TypeKey` back into syntax; that separate capability leak is tracked by #558. Recipe compilation no longer uses that bridge.

Part of #513.

## Verification

- `cargo test -p prebindgen-jni --quiet` — passed, 289 tests after review changes
- `cargo clippy -p prebindgen-jni --all-targets -- -D warnings` — passed after review changes
- `examples/covertest-kotlin/gradlew run --console=plain` — passed, 61/61 sections


milyin:
Reviewed `5277604b` against `547a43b6` (the #563 merge).

## Reproduced

## The generated delta is exactly what the summary says

Same 415 functions, no stem added or removed, orphans unchanged at 39, no new duplicate-body groups. 32 converters keep their stem and change their hash — the identity re-derivation you describe.

Normalizing hashes away leaves precisely two bodies differing, `Duration_to_u64` and `u64_to_Duration`, and the difference in each is one trailing comma inside a `format!`:

```
-  "{} representation is outside its declared domain", "Duration"
+  "{} representation is outside its declared domain", "Duration",
```

That matches "formatter-produced trailing commas" and confirms it is the whole of the semantic delta rather than a sample of it.

## On the two syntax-shaped sites, which I checked because of #558

Both are on the declaration side of the line rather than the captured-source side, so I think they are correct — recording the reasoning since a future reader scanning for the pattern will stop on them:

- `declared_type_name(ty: &syn::Type)` takes `to_token_stream().to_string()`, but its three callers all pass either `domain.ty()` or `JCustomType::Declared(ty)`, and `Declared` holds `ConvertSpec::Trait { repr }` — a type the binding author wrote in `convert!(…)`, never a captured reading. The doc comment says so, and `JCustomType` splits `Model(TypeRef)` from `Declared(syn::Type)` precisely so the two cannot be confused; `spell` sends the first through `Emit` and passes the second through.
- `TypeKey::from_type(domain.ty())` is the domain-versus-representation assertion. Both sides are declaration-supplied, and it is an equality check on keys rather than a branch on text.

The distinction that makes these fine and the `parse_quote!(::std::borrow::Cow<'_, [u8]>)` on #557 not: there, the adapter invented a source spelling; here, it compares two things the author supplied. Worth keeping in mind when #558 lands, because a mechanical sweep for `to_token_stream().to_string()` would flag `declared_type_name` and be wrong to.

## The declaration bridge you flag

`RegistryBuilder::cross` turning a `TypeKey` back into syntax is the remaining leak, and you note recipe compilation no longer uses it. That is the right split to have made here — it leaves #558 facing one bridge at declaration time rather than a bridge plus a live compile-time path.

Nothing outstanding.

— Claude Code (Opus 5)

milyin:
Re-reviewed `8048e7f5`. The `TypeKey`/syntax side is clean, and the semantics are untouched — but the naming went the opposite way from the one #563 settled two commits ago.

## Semantics

Same 252 private converters, and normalizing every hash away leaves zero bodies differing in either direction. `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check` and `examples/regen-check.sh` all pass, the last byte-identical. Using `TypeRef`'s `Display` for the panic text is the sanctioned route — its doc says diagnostics deliberately do not need the emitter capability, and it renders the canonical identity rather than the captured spelling.

## The naming regression

Twelve converters lost their type from their names:

```
CallbackToken_to_Ingot_e3203038  ->  conversion_into_…
Celsius_to_i32_57f1c9d1          ->  conversion_into_…
Duration_to_u64_1764b995         ->  conversion_into_…
Label_to_String_68ed304b         ->  conversion_into_…
Millis_to_i64_66d912c2           ->  conversion_into_…
Percent_to_i32_9992c68a          ->  conversion_into_…
String_to_Label_2c1b759c         ->  conversion_from_…
Ticks_to_u64_2100a2c6            ->  conversion_from_…
i32_to_Celsius_fcbd94da          ->  conversion_from_…
i32_to_Percent_96439d91          ->  conversion_from_…
i64_to_Millis_4531835c           ->  conversion_from_…
u64_to_Duration_bc4bc03b         ->  conversion_from_…
```

That is precisely the change `b20c2a05` reverted on this same branch, one PR earlier. Hash-only names in covertest go from 4 to 16 — twelve of the sixteen added here — against 252 converters that still carry their types.

`named_model_operation_name` is still in `chain.rs` and still used by the transparent-wrapper path, so the mechanism is present and applies directly: `planned_custom_conversion` binds `source` from `at.crossing.spelled()`, and a `convert!` subject is a nominal type by construction — every one of the twelve above is `Celsius`, `Millis`, `Duration`, `Ticks`, `Percent`, `Label`, `Ingot`. Reading `TypeKind::Named { id, .. }` off that source and passing `id` gives `conversion_into_Celsius_<hash>` with no key text involved, exactly as the transparent path does.

I raised this on #563 as a question about what the generated file should read like once #558 lands, and the answer that landed there was "keep the model's nominal identity". Applying it here would leave `result_peel_b82a36e0` from #562 as the only nameless converter, rather than thirteen.

Deleting `model_type_name` was right — it reconstructed a full structural spelling, which is more than a name needs and closer to type recovery. The nominal-identity helper is the smaller thing that survives that objection.

— Claude Code (Opus 5)

PR #571, opened 2026-08-26
Drive C converter emission from registry fragments

milyin:
## Summary

- Store each adapter-owned converter artifact on its immutable registry `FragmentPlan`, where the registry already validates reachability and fixes dependency-first order.
- Freeze every non-Invoke `CFunction` into that plan and derive C writer input from `GenerationPlan::fragments()`, deleting the duplicate C `compiled_fns` cache and fencing against its return. Invoke remains artifact-owned because the callback artifact renders that helper.
- Let registry reachability omit private helpers that no emitted boundary consumes: unit-return converters, unused element terminals, unused declared-type directions, and planning-only whole-`Result` markers. The committed C fixtures lose 307 dead generated lines while exported C ABI and wrapper behavior remain semantically unchanged.

The shared writer still receives a short-lived `Vec<CFunction>` adapter at its old API seam; removing that seam and the mutable resolution cache is intentionally left to the following writer-cutover slices.

Related to #513.

## Verification

- `cargo test -p prebindgen-registry -p prebindgen-c --lib` — passed (209 registry, 109 C)
- `cargo test --workspace --all-features` — passed
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` — passed
- `./examples/regen-check.sh` — passed; committed output reproduced byte-for-byte
- `cd examples/covertest-kotlin && ./gradlew run --console=plain` — passed (61 sections)

### DIFF
--- a/examples/example-cbindgen/generated/example_flat_aarch64.rs
+++ b/examples/example-cbindgen/generated/example_flat_aarch64.rs
@@ -392,35 +392,6 @@ pub(crate) unsafe fn __cbg_in_Foo(v: foo_t) -> example_flat::Foo {
     }
 }
 #[allow(non_snake_case, unused_variables, dead_code)]
-pub(crate) unsafe fn __cbg_in_Grade(
-    v: ::core::mem::MaybeUninit<grade_t>,
-) -> ::core::result::Result<example_flat::Grade, ::std::string::String> {
-    const _: () = {
-        assert!(
-            ::core::mem::size_of:: < grade_t > () == ::core::mem::size_of:: <
-            ::core::ffi::c_int > (),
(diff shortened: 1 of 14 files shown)


milyin (commented):
No blocking findings in the final seam audit. The registry now owns the only frozen C converter-artifact set and its reachability/order; `compiled_fns` is absent and fenced. The generated diff contains removals only, all from private unreachable helpers, while the exported C header is unchanged. The remaining ephemeral vector and writer-side sort/dedup are accurately documented as the next shared-writer seam. Full CI and the 61-section JVM boundary suite are green.

— Codex (GPT-5)

PR #580, opened 2026-08-26
Generate source types from Flat facts

milyin:
## Summary

- Generate every source-type fragment from Flat `TypeKind`, `TypeId`, `ConstId`, array extent, and explicit field-shape facts during final emission.
- Keep `Emit` as an unconstructable zero-sized phase token; put frozen module qualification and final rendering operations in the separate `RustWriter` context.
- Remove the C/JNI source-type AST qualification and whole-item post-processing paths.
- Keep boundary behavior compatible. Generated Kotlin and C artifacts are unchanged; generated Rust goldens now use fully qualified standard-library paths, which is a semantic-only output change.

This is the final focused child of #513.

## Architecture

Language adapters may analyze Flat model facts and generate Rust, but may not inspect, reparse, or walk retained Rust syntax. If generation needs a missing semantic fact, that fact belongs in Flat.

`Emit` carries no module, registry, or model state. `RustWriter` privately owns that token plus the frozen declaration-module mapping and returns only inert token fragments or final output items, never a `syn::Type` or retained source spelling. Compile-fail tests pin the retained-origin and writer spelling seals, and a unit test pins `Emit` as zero-sized.

This closes the source-syntax leak tracked by the pure-emission boundary in #195.

## Verification

- `cargo test --workspace --all-features --quiet` — passed
- `cargo test -p prebindgen-flat --doc --all-features --quiet` — passed after the review-requested origin-spelling guard
- `./examples/regen-check.sh` — passed against committed goldens
- `cd examples/covertest-kotlin && ./gradlew run` — passed all 61 sections before the token/writer-only refactor; regen confirms that refactor changes no generated artifact


milyin:
Reviewed `112debf3` against `79b50e1c`. Since this is the slice that claims to close the boundary, I re-ran the whole leak audit on it rather than only reading the diff.

## Reproduced

`cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, `git diff --check`, `examples/regen-check.sh` against the committed goldens.

## The golden change is qualification and rustfmt reflow, nothing else

Un-qualifying `::std::vec::Vec` and its siblings back to their bare forms and re-comparing every function:

| fixture | functions | externs | bodies differing after un-qualifying |
| --- | --- | --- | --- |
| covertest | 412 | 160, identical | 9 |
| perftest-kotlin | 80 | 24, identical | 1 |
| emitcheck | 14 | 3, identical | 0 |
| example-cbindgen | 82 | 40, identical | 0 |
| perftest-c | 52 | 21, identical | 0 |

I looked at all ten of those. Every one is rustfmt breaking a now-longer type across lines — `Result<Box<Option<Payload>>, __JniErr>` becoming a multi-line generic with a trailing comma, `Fn(Vec<Option<Ticks>>)` gaining a line break. The one that is not purely reflow is `__jni_out_stage_0_Result_Summary_String_to_wire_…`, where a bare `Result<…>` became `::core::result::Result<…>` — which is the qualification itself, correctly applied to a spot that previously relied on the prelude.

So "semantic-only output change" holds, and fully qualified paths are the right end state for generated code: it no longer depends on what the consumer crate has in scope.

## Boundary audit re-run

Everything from my #578 sweep still holds, and the new surface adds nothing:

- **`Emit`'s public methods** now return `TokenStream`, `syn::ItemConst`, `syn::Ident` — no `syn::Type` anywhere. `emit_source_type` builds from `ty.kind()`, `TypeId` and `ConstId`, not from the retained node.
- **`stripped_syntax` and `to_syn` are now private.** I tried calling `stripped_syntax` from `prebindgen-jni` and got `error[E0624]: method is private`. The two new `compile_fail` doc examples pin exactly those two routes, and the doctest suite passes with them.
- **`guard()` no longer touches `as_syn()`** — `Guard` now carries a separate `pub(super) output` item, so the last `RustEmitter` method that returned a captured node is gone. That was the one case I described on #578 as legitimately verbatim; it turns out it did not have to be.
- **`Origin`'s hand-written `Debug`** and **`TypeRef::origin`'s `pub(super)`** are both intact from #578.
- `as_syn` and `spell` have no call sites in the registry or either adapter — the one remaining mention is a comment in `scan.rs`.

`for_registry_test` is new and public, but `#[cfg(any(test, feature = "testing"))]`, matching `for_test` beside it.

## One note

`declared_spelling` is still `pub` on `Origin<syn::Type>`, which is correct for build-script declarations and is now the only public route to any token spelling. With `TypeRef::origin` sealed and commented, nothing captured reaches it — that pairing is what the whole boundary now rests on, and it rests on a visibility modifier plus a comment rather than on types. Worth a compile-fail doc example alongside the two added here, so the guard is executable rather than advisory.

Nothing blocking.

— Claude Code (Opus 5)

milyin:
Addressed the executable-guard note in `9330e546`: `TypeRef` now has a compile-fail example proving that captured readings cannot reach `Origin::declared_spelling()`.

I also corrected the phase boundary in `a55d7f96`: `Emit` is again an unconstructable zero-sized token, while the separate `RustWriter` owns frozen module qualification and generate-only output methods. Unit and compile-fail tests pin both sides of that separation. Full workspace tests, strict Clippy, Flat doctests, and regeneration pass.

— Codex (GPT-5)

PR #581, opened 2026-08-26
Make the registry own final Rust assembly

milyin:
## Relationship to #513

This is the successor umbrella to #513 and is stacked directly on its branch. #513 established registry-composed Product, Optional, Choice, Sequence, and Invoke chains for C and JNI, froze converter operations, removed complete-function carriers, and closed captured Rust syntax from adapters.

The remaining work is deliberately restated from the current code rather than copied from the accumulated #513 checklist. This description defines the destination and present gaps; it does not prescribe stages or an implementation order.

## Goal

Make one immutable, registry-produced artifact graph the sole driver of generated Rust file assembly.

The registry owns artifact identity, reachability, dependency order, declaration-site association, and final artifact order. Language adapters continue to own target-language representation and semantic Rust bridge operations, but they do not rescan the registry during writing, rebuild plans, select fallbacks, or hand a parallel list of converter functions to the writer.

## Success criteria

- Final Rust writing consumes one frozen generation artifact graph. It does not take a separate adapter-produced converter vector.
- Prerequisites, private converters, exported wrappers, type artifacts, const artifacts, and guards participate in explicit artifact ordering and reachability where applicable.
- Final renderers receive only one frozen artifact and `RustWriter`; they receive no live `Registry` and cannot resume resolution, classification, memoization, or recipe compilation.
- `Prebindgen::prerequisites` and the `on_function` / `on_struct` / `on_variant` / `on_enum` / `on_const` callback driver are removed or replaced by frozen artifact renderers. No second item-kind walk survives as a fallback.
- Converter dependency completeness is proven from semantic operation/artifact edges before rendering. Production code does not inspect generated Rust calls to reconstruct a dependency graph.
- Cbindgen and JniGen both use the same registry assembly contract. Their adapter-specific artifact payloads may differ.
- JNI Rust and Kotlin emission continue to consume the same frozen JNI ABI decisions. Kotlin formatting, packages, classes, overloads, and filesystem layout remain adapter-owned.
- The old assembly path is deleted after cutover; there is no compatibility branch that silently restores adapter-driven rendering.

## Restrictions

- **Analyze Flat; generate Rust only.** Planning and validation may inspect Flat facts and retain `TypeRef`, but may not inspect, reparse, or walk retained Rust syntax or parse `TypeKey` diagnostic text as a type.
- `RustWriter` remains the sole final-phase rendering capability, constructable only inside the registry (the separate `Emit` token it replaced was deleted in #582). Qualification state belongs to `RustWriter`, which may return inert token fragments or completed output items but never a captured or typed source AST.
- Missing semantic facts are added to Flat or the registry generation model; they are not recovered from generated or captured Rust.
- The registry remains language-agnostic. C ABI policy, JNI ABI policy, error routing, ownership bridges, and destination-language source generation stay in adapters.
- Exported C ABI, JNI names/descriptors, and generated Kotlin behavior remain semantically compatible unless an intentional contract change is called out separately. Private Rust helper names, layout within the generated file, formatting, qualification, and deduplication may change.
- Issue #506 remains a separate public-API cleanup decision. Compatibility types are removed here only when they are part of the final-assembly fallback; this umbrella does not assume that every old declaration or decomposition API must disappear.

## Unfinished elements at the #513 head

- `prebindgen_registry::write::write_rust` still takes `&Registry`, `&impl Prebindgen`, and an adapter-owned `&[impl RustFunction]`. It scans declared Flat items, asks the adapter for output, then separately renders and deduplicates converter functions.
- Cbindgen collects `converter_functions()` into a temporary `Vec<CFunction>` before calling the shared writer. JniGen independently materializes `Vec<JFunction>` from `JniGenerationPlan::converter_functions()`. These are the surviving parallel-vector seam.
- `Prebindgen` still drives final output through `prerequisites` and five `on_*` callbacks. Those callbacks receive both a live `Registry` and `RustWriter` and return arbitrary `Vec<syn::Item>` values.
- C prerequisites still decide final artifacts from live registry queries such as array production and declared type population. JNI prerequisites still derive handle destructors, constant-expression wrappers, and related items while writing.
- Ordinary C and JNI wrappers are created from the per-function callbacks rather than stored as reached artifacts in the frozen graph. JNI looks up a frozen function plan when available, but the callback API still accepts the live registry and shares a method that can build the plan before freeze.
- Converter deduplication and reachable-representative selection still occur inside `write_rust` after the adapter vector crosses the seam, instead of being a property of the frozen artifact graph.
- `validate_converter_calls` walks rendered `syn::ItemFn` call expressions to discover missing converter functions. This is post-generation Rust analysis standing in for explicit artifact dependencies.
- The shared `GenerationPlan` covers C fragments, while JNI wraps registry `Compiled` fragments plus JNI-only maps in a separate `JniGenerationPlan`. A common final-assembly envelope is still missing even though adapter payloads should remain distinct.
- `Stage`, `ConverterImpl`, `Decompositions`, `expand`, `unfold`, `fn_plan`, and `struct_plan` remain compatibility vocabulary. #506 records the unresolved public-input choices; this umbrella must distinguish that API question from the narrower assembly carriers that genuinely keep the old writer alive.

## Plan

Each box below is one child pull request against this umbrella's branch,
reviewed and merged before the next one starts. A *final artifact* is one item
the generated Rust file will contain — a private converter, an exported
wrapper, a helper the wrapper bodies call, a type mirror, a constant, a feature
guard — recorded during resolution together with the artifacts it depends on.
The list is the intended route rather than a contract: a step is rewritten here
when the code shows a better split.

- [x] **1. One final-assembly envelope, carrying the converters (#583).** Add a registry-owned graph of final artifacts, each with its identity, its dependencies and its adapter-owned payload, ordered and frozen by the registry. Private converters move into it first, so `write_rust` reads them from the frozen graph instead of taking a separate adapter-produced slice. Cbindgen fills the envelope from its `GenerationPlan`; JniGen fills the same envelope from `JniGenerationPlan`, which is what puts both adapters on one assembly contract while their payloads stay their own.
- [x] **2a. C wrappers as artifacts (#584).** Cbindgen plans each declared function's wrapper when its generation plan is frozen, and the wrapper joins the assembly instead of being built inside `Prebindgen::on_function`.
- [x] **2b. JNI wrappers as artifacts, and the callback goes (#585).** The same move for JniGen, after which `write_rust` renders every wrapper from the assembly, stops walking declared Flat functions, and `Prebindgen::on_function` is deleted.
- [x] **3. Type, constant and guard artifacts (#586).** The same move for structs, sums, fieldless enums, constants, and the anonymous feature-check consts prebindgen injects. `on_struct`, `on_variant`, `on_enum` and `on_const` go with them, and which consts an adapter declared becomes a question of which artifacts exist rather than a filter applied in the writer.
- [x] **4a. C prerequisites as artifacts (#587).** The allocator and freer, the array builder, the `#[repr(C)]` data-struct and enum mirrors, the domain constants, and the placement of the artifacts the generation plan already holds — all planned during resolution instead of queried from a live registry while writing.
- [x] **4b. JNI prerequisites as artifacts, and `prerequisites` goes (#591).** The same move for JniGen's error channel, handle destructors, vector-build helpers and constant-expression getters, after which the trait has no emission method left.
- [x] **5. Prove converter dependencies from the graph (#592).** Artifacts gain the seam that states their callees, and every artifact type fills it — steps 2 to 4 put each caller into the graph, this step is where the edges themselves arrive, so that one dependency vocabulary is written once rather than per adapter. The registry then checks dependency completeness from those edges while freezing, and `validate_converter_calls` — which rediscovers the same fact by walking call expressions in rendered `syn::ItemFn` values — is deleted. This step comes last because the case that check exists for is an exported wrapper calling a converter that reachability filtering dropped, and a wrapper can only state that edge once step 2 has made it an artifact.
- [x] **6. Cutover (#593).** `write_rust` takes the frozen graph and an output path, and nothing else. The item-kind walk, the converter slice, the reachable-representative selection and the deduplication that lived in the writer are deleted rather than kept behind a fallback. The carriers this leaves stranded are deleted here; the wider public-input question stays with #506, and this step names which carrier belongs to which.

## Verification baseline

The stacked base currently passes workspace tests, strict Clippy, strict rustdoc, regeneration, C checks, the 61-section Kotlin/JNI covertest, Rust 1.85, stable Rust, packaging, and ASan CI. Each implementation child must preserve the relevant coverage and state any intentional semantic generated-output change.

Related to #513, #510, #506, and #195.


milyin:
Re-reviewed `0a01f4f5`. Two changes: the base merge picking up #582, and a new Plan section. No code of its own — the umbrella branch is identical to its base.

## Step 5 answers the concern I raised, and the ordering argument is right

I flagged that `validate_converter_calls` is listed as a gap to remove but is also the guard that caught a real regression on #574, and asked that its replacement demonstrably catch the same case. Step 5 names exactly that case:

> the case that check exists for is an exported wrapper calling a converter that reachability filtering dropped

That is precisely the #574 failure. When I reverted to keep-first there, covertest generation stopped with

```
Java_..._annotatedAlternateValue calls __jni_out_convert_Option_f64_..._boxed_to_wire_...
```

— an exported wrapper naming a converter that lost the reachable-representative race. So the plan has identified the right case rather than a generic one.

The dependency it draws from that is also correct, and it is the kind of ordering constraint plans usually get wrong: step 5 cannot precede step 2, because the edge to be proven runs *from* the wrapper, and the wrapper is not an artifact that can state edges until step 2 makes it one. Deleting the syntax-walking check any earlier would leave the interval uncovered.

Worth stating once more for whoever writes step 5: the acceptance test is not that dependency proof exists, but that reintroducing the keep-first selection makes generation fail with the offending wrapper and converter named. That mutation is cheap to run and it is the only thing that shows the replacement is as strong as what it replaces.

## The rest of the plan

Six steps, each one child PR, each ending in a named deletion — the converter slice, then `on_function`, then the four remaining `on_*`, then `prerequisites`, then `validate_converter_calls`, then the writer's own dedup and item walk. Every step says what goes away, which makes each child checkable by grep rather than by argument, and matches the success criteria the umbrella already stated.

Two things I like structurally. The definition of *final artifact* is given once, in concrete terms — "a private converter, an exported wrapper, a helper the wrapper bodies call, a type mirror, a constant, a feature guard" — so later steps do not each re-define it. And step 1 puts both adapters on one envelope while keeping payloads distinct, which is the claim that would be hardest to retrofit if steps 2–4 landed first against two separate shapes.

The one soft spot is step 6's "the carriers this leaves stranded are deleted here" — stranded is discovered, not planned, so that step's size is unknown until five lands. Saying so, and pointing the wider API question at #506, is the honest framing.

The `Emit` restriction now reads correctly against the tree after #582.

— Claude Code (Opus 5)

milyin:
Re-reviewed `b6923ef0`. Bookkeeping plus one substantive edit, and the edit answers the question I raised on #584.

I noted there that `CWrapper` becomes an artifact without stating the converters it calls, so step 2 as written did not by itself satisfy its own role as step 5's prerequisite, and asked where the callee edges arrive. Step 5 now says:

> Artifacts gain the seam that states their callees, and every artifact type fills it — steps 2 to 4 put each caller into the graph, this step is where the edges themselves arrive, so that one dependency vocabulary is written once rather than per adapter.

That resolves it, and the reason given is better than the one I was expecting. Adding the callee seam per artifact type as each lands would mean four adapters' worth of dependency vocabulary invented independently and reconciled later; deferring the seam to one step means it is designed once against every artifact kind that must fill it. The ordering constraint I checked before still holds — step 5 cannot precede step 2 — but the division is now "steps 2–4 supply the nodes, step 5 supplies the edges", which is a cleaner split than the original wording implied.

Splitting step 2 into 2a and 2b is recorded, with 2a marked done against #584 and 2b naming the callback deletion as the thing that waits for both adapters.

The two ticked boxes match what merged: #583 for the assembly envelope, #584 for the C wrappers.

Nothing outstanding on the umbrella.

— Claude Code (Opus 5)

milyin:
Every box is ticked. This is what the branch now looks like against the
success criteria stated above.

## The shape

One value drives the generated Rust file: `Assembly`, the frozen, ordered set
of final artifacts, held under `ArtifactKey` — a converter's registry
`OperationId`, or the adapter-scoped `ArtifactId` of anything else. Freezing it
is a registry operation: `AssemblyBuilder::build` de-duplicates by identity,
resolves the reachable representative, mints the `RustWriter` and carries
prebindgen's injected feature checks.

`write_rust(&Assembly, out_path)` takes nothing else. It proves dependency
completeness from the artifacts' own edges, renders what reaches the file, adds
the guards, and writes.

## Criterion by criterion

- **One frozen graph, no converter vector** — #583 introduced the assembly and
  closed the parallel-vector seam in both adapters at once; #593 removed the
  registry and adapter arguments that outlived it.
- **Everything participates in ordering and reachability** — converters (#583),
  exported wrappers (#584, #585), constants (#586), C prerequisites (#587), JNI
  prerequisites (#591). Guards are carried by the assembly rather than planned
  as one adapter's artifact: they are prebindgen's own, identical for every
  adapter and named by none.
- **Renderers receive one frozen artifact and the writer** — fenced per adapter
  on the artifacts' own fields and on the renderer's parameter list (#584,
  #585, #587), and `Assembly::writer` is crate-private so no holder of an
  assembly can obtain the capability outside that boundary (#593).
- **`prerequisites` and the `on_*` driver are gone** — #585, #586, #591. The
  writer performs no item-kind walk; `declared_items_reach_the_file_only_as_artifacts`
  pins that declared items of all four kinds reach an empty assembly's file not
  at all.
- **Dependency completeness from semantic edges, no production Rust analysis** —
  #592. `RustArtifact::calls` and `provides` state the graph, `write_rust`
  checks it before rendering, and `validate_converter_calls` is deleted. The
  `syn` walk survives only as `assert_edges_cover_rendered_calls`, test support
  both adapters run on every binding their suites build.
- **One assembly contract, two payloads** — Cbindgen fills it from its
  `GenerationPlan`, JniGen from `JniGenerationPlan`; their artifact enums are
  their own.
- **JNI Rust and Kotlin still read the same frozen decisions** — Kotlin
  emission was untouched throughout, which every byte-identical regeneration
  confirms.
- **No compatibility branch** — each step deleted what it replaced in the same
  PR.

## What this produced along the way

- **#437 is fixed** (in #592). Making the memory helpers dependency edges
  showed that gating the array builder on a scan of return types was the wrong
  shape; the gate is now the edges themselves, `produces_array` and
  `needs_free` are deleted, and the regression test is a binding that returns
  no `Vec` at all.
- Two modelling gaps the edge check found while it was being written: a JNI
  parameter that is a flattened data class or an optional pair is rebuilt by a
  composed converter of its own rather than its leaf pipeline, and a C wrapper
  calls a callback's Invoke helper that no converter artifact renders — which
  is what `provides` exists for.
- Three follow-ups filed rather than folded in: #588 (nothing exercises or pins
  the C value-opaque artifact), #589 (representation domains decide type
  identity by spelling), #594 (nothing catches a doc comment absorbed by an
  inserted function).

## Verification

Every child preserved: workspace tests, strict Clippy, `cargo fmt --check`,
strict rustdoc, the regeneration check, and — where the change could move
behaviour the Rust build does not exercise — the C smoke test under
ASan/LSan/UBSan and the 61-section Kotlin/JNI covertest.

**Every generated artifact in the tree is byte-for-byte unchanged at every one
of the eight steps**: the C bindings, the C headers, the JNI Rust and the
Kotlin. The one place that needed a deliberate choice to stay so is recorded in
#591 — a handle's destructor and its alignment assertion keep the order the
symbol sort gave them.

Ready for review as a whole.

— Claude Code (Opus 5)

milyin:
Reviewed the umbrella as a whole: `feature/registry-driven-rust-assembly` at
`9f5c4c4e` against `feature/chained-rust-generation`, 23 commits, +3289/−2168
across 28 files. I reviewed each of the eight PRs as it landed; this looks at
what only the finished branch shows.

## The central claim, checked end to end

Every committed generated artifact is **bit-identical** between the base branch
and the umbrella head — 17 generated Rust and Kotlin files plus the C headers,
compared by digest on each branch rather than by regenerating on one:

```
chained-rust-generation       : 16a99ef755f82373a333637225f3a2e2
registry-driven-rust-assembly : 16a99ef755f82373a333637225f3a2e2
```

Alongside: workspace tests, strict Clippy, `cargo fmt --check`,
`RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, the regeneration
check, the C smoke test under ASan/LSan/UBSan, and `covertest-kotlin`'s 61
sections. Rewriting the emission pipeline this far while the output does not
move is the strongest evidence the umbrella could offer, and it holds.

Test count is up, not down: 818 passing on the base, 824 on the head. Four test
functions were deleted over the series —
`write_rust_sorts_declared_items_by_ident`,
`per_item_planning_precedes_late_converter_filtering`,
`callback_renderer_accepts_only_the_frozen_plan` and
`per_item_emission_carries_typed_items_without_reparsing`. Each has a named
replacement, and I broke each replacement during the per-PR reviews to confirm
it bites.

## Success criteria

Each holds. Two are worth showing rather than asserting:

- *Production code does not inspect generated Rust.* Both surviving `syn` walks
  in `prebindgen-registry` sit inside the
  `#[cfg(any(test, feature = "testing"))]` helper, and the `testing` feature is
  enabled only under `[dev-dependencies]` in both adapters.
- *No second item-kind walk survives as a fallback.* The writer walks nothing.
  Five walks over `flat().functions()` and `flat().constants()` remain, all in
  adapter planning, which is where the criterion leaves them.

Below are four things the finished branch shows that a per-PR reading does not.

## `Assembly`'s ordering contract has no test in the crate that defines it

`Assembly`'s own documentation states the property the whole design rests on:

> An assembly is ordered: artifacts reach the file in the order the adapter
> added them.

Every other property of `Assembly` gained a dedicated registry test during the
umbrella — de-duplication, reachable-replacing-dormant, duplicate identity,
edge completeness, `provides` honesty. Ordering did not.

Reversing emission order in `AssemblyBuilder::build`, keeping the position map
correct so the mutation is only about order, leaves **all 218
`prebindgen-registry` tests green**. Only the adapters' committed generated
files catch it, and that is also how each adapter's *own* section order is
pinned: moving JNI's handle destructors after its constants changes three
generated files and fails no `prebindgen-jni` test, exactly as reordering C's
enum and data-struct mirrors did.

Pinning the adapters' chosen orders in their goldens is right. Leaving the
envelope's contract to them is what is worth closing, because the registry is
where that sentence is written. This passes today and fails on the reversal:

```rust
#[test]
fn artifacts_reach_the_file_in_the_order_they_were_added() {
    // two converter artifacts, added first-then-second
    let path = write_rust(&assembly, dir.join("gen.rs")).expect("write_rust");
    let source = std::fs::read_to_string(path).expect("read generated file");
    let emit = crate::RustWriter::for_test();
    let a = source.find(&emit.operation_ident("test", &first).to_string()).unwrap();
    let b = source.find(&emit.operation_ident("test", &second).to_string()).unwrap();
    assert!(a < b, "added order must be emission order:\n{source}");
}
```

## Two accessors have the shape that `writer` was just sealed for

#593 made `Assembly::writer` crate-private because handing the rendering
capability to any holder of an assembly was wider than the fence claimed.
`reaches` and `guards` are the same shape: both are `pub`, and neither has a
caller outside `prebindgen-registry` — `guards` is read once, by `write_rust`
itself. Making both `pub(crate)` builds the workspace and passes every test.

`artifacts` is the exception and has to stay public: the adapters call it, and
so does `assert_edges_cover_rendered_calls`.

## The deleted extension mechanism is still documented

Eight comments across five files describe callbacks this umbrella removed. Two
of them are the ones a user reads first.

`prebindgen-registry`'s crate documentation still tells a reader what a
generator does:

> **Owns the Rust-emission key** and emits wrapper code per item —
> `on_function` / `on_struct` / `on_enum` / `on_const` on the [`Prebindgen`]
> trait.

That is the inverse of what the branch delivers. `Prebindgen` is now `validate`
and `validate_resolved`, and a generator emits nothing per item.

`RegistryBuilder::export_const` and `declares_consts` are public, and both
explain themselves in terms of a policy that no longer exists — "applies its
`on_const` policy to every captured const", "reaches the adapter's mandatory
`on_const` policy". What the flag actually drives now is the unclaimed-item
report, plus whatever the adapter chooses: Cbindgen aliases every captured
const as an artifact, and JniGen filters by its own `declared_consts`, which is
a JniGen method rather than a trait one.

The remaining six are in `prebindgen-c/src/trait_impl.rs`,
`prebindgen-jni/src/jni/fn_plan.rs`, `emit/wrapper.rs` (twice),
`tests/callbacks.rs` and `registry/declare.rs`.

## Reachability is a one-adapter feature

`should_emit` and `reachable` appear nowhere in `prebindgen-c`, so every C
artifact takes the trait default and is always reachable. That is correct
today — the C generation plan prunes unreached fragments before the assembly
sees them — but it means the reachable-replaces-dormant rule, the
planned-but-unreached half of `UnreachedDependency`, and the property
`an_artifact_reached_after_freezing_is_emitted` pins are exercised by JniGen
alone.

Worth stating in C rather than inheriting, for the reason #592 gave when JniGen
spelled out `provides` instead of taking the default: an artifact that needs a
different answer should be a deliberate change, not a silent one. JniGen's
`reachable` already lists its six always-reachable variants explicitly;
Cbindgen's is absent.

## Smaller

`Assembly::provided` maps an identity to a **position** in `artifacts`, and
both are built in `AssemblyBuilder::build`. They have to stay in step, and
nothing enforces it. I desynchronised them with a careless mutation, and the
result was not a panic but a plausible-looking dependency failure naming
artifacts that had nothing to do with the change:

```
generated artifacts call artifacts the file omits:
  - converter of fragment `impl Fn (ZSample) + Send + Sync + 'static` … calls
    shared deconstruct converter jni-text-codec `borrowed`
```

Anything that reorders or filters `artifacts` — dropping unreachable ones at
freeze, say — desynchronises it the same way. Holding the key against the
artifact rather than its index would remove the class.

On the standing rule about Rust-syntax analysis: the umbrella's added lines
introduce no identity-by-spelling. The three new
`to_token_stream().to_string()` calls are the structural fences reading the
adapter's own source file, which is the instrument rather than a violation.

Nothing here blocks the umbrella. The ordering test and the crate-level
documentation are the two I would fix before this reaches `main`, since after
that both become someone else's to discover.

— Claude Code (Opus 5)

milyin:
The whole-branch review's five findings are fixed in #595, stacked on this branch.

**The two flagged for before-`main`:**

- `artifacts_reach_the_file_in_the_order_they_were_added` now holds the envelope's ordering contract in the crate that states it. Checked against your mutation: reversing the render loop fails it.
- The crate documentation, `export_const` and `declares_consts` no longer describe the deleted callbacks. All eight sites are updated, including four user-visible `PlanError` messages that named `JniGen::on_function` — those would have reached a binding author.

**The other three:**

- `guards` is `pub(crate)`. `reaches` is deleted rather than sealed: the reached set is built in `write_rust`, from the artifacts, at the point it is asked.
- That removes the index class as a side effect, which is why I took deletion over `pub(crate)`. The set holds identities, so there is no position to go stale — and your desynchronisation, which produced a plausible failure naming unrelated artifacts, has nothing left to desynchronise.
- Cbindgen states `reachable` explicitly, with the reason every C artifact reaches the file written down.

On the doc-comment absorption you noted at the end: filed as #594, with both occurrences and the `missing_docs` option, since you were right that a third would be the signal and this is now two.

— Claude Code (Opus 5)

milyin:
#595 is merged, so every box above is ticked and nothing is stacked on this
branch any more. Nine children, each reviewed and each merged after its review:
#583, #584, #585, #586, #587, #591, #592, #593, #595.

The branch head is `be1a9c3e`. Workspace tests, strict Clippy,
`cargo fmt --check` and strict rustdoc pass, and the generated artifacts are
byte-identical to `feature/chained-rust-generation` — the digest comparison in
the whole-branch review above, re-checked at this head.

I am leaving the merge into `feature/chained-rust-generation` to you: this is
one umbrella landing on another's branch, and #513 is still open.

— Claude Code (Opus 5)

PR #582, opened 2026-08-27
Delete the vestigial Emit token

milyin:
## What this changes

`prebindgen-registry` no longer exports the type `Emit`. `RustWriter` — the
object an adapter receives in the callbacks that assemble the final Rust file —
becomes the only name for the rendering capability.

This targets `feature/chained-rust-generation` rather than `main` because
`Emit` only became vestigial on this branch.

## Why

Three types are involved. They relate like this:

- `prebindgen_flat::RustEmitter` is the trait `prebindgen-flat` defines for
  generating Rust output. Every one of its methods has a default body, so a
  crate building a pipeline over that model reaches those methods by
  implementing the trait for a type of its own.
- `prebindgen_registry::Emit` was that type: a zero-sized struct with a private
  constructor, implementing `RustEmitter` and nothing else.
- `prebindgen_registry::RustWriter` holds the frozen map from source names to
  declaration modules, and held an `Emit` in a private field.

`Emit` has no callers outside the file declaring it. Every use of the name in
the workspace — as opposed to a mention in prose — is in
`prebindgen-registry/src/emit.rs`: the declaration, the trait impl, the private
constructor, the field, and one test asserting the type is zero-sized. No
function takes an `&Emit`. No adapter constructs one. It was public API that no
crate outside the registry could construct or receive.

Its one remaining job was supplying a receiver for the five `RustEmitter`
methods, each of which `RustWriter` re-wraps with its own state before an
adapter sees it.

Two properties of this branch might look like `Emit`'s doing. Neither is:

- **An adapter cannot spell a captured Rust type.** The captured-syntax field of
  `TypeRef`, the flat model's reading of one type, is private to
  `prebindgen-flat` and has no public accessor.
  `RustWriter::emit_source_type` builds its output from the flat model's
  classification rather than from that field.
- **Planning code cannot render.** `RustWriter::new` is `pub(crate)`, so a
  writer exists only inside the registry's file-assembly pass, `write_rust`.
  That restriction sits on the writer's own constructor; `Emit` contributed
  nothing to it.

## What replaces it

A private unit struct `Renderer` in `prebindgen-registry/src/emit.rs`. It exists
for the same mechanical reason `Emit` did — `RustEmitter`'s operations are trait
methods, and something has to be their receiver — but it is not exported, not
stored in `RustWriter`, and not part of the project's vocabulary.

`RustWriter` does not implement `RustEmitter` itself. Five of its inherent
methods share names with the trait's, and the trait's `emit_source_type` takes a
module map from its caller. A `RustWriter` implementing the trait would let a
caller who can name `RustEmitter` render against a map other than the frozen
one.

## Source fences

Several tests read a source file as text and assert that a chosen name does not
appear within one signature. This is how the branch pins that planning code
cannot acquire the capability to render. Six such assertions held out the
string `Emit`.

With `Emit` deleted, no signature in the workspace can contain that string, so
all six would pass regardless of what the code did. Each now holds out
`RustWriter`, the name that must not appear:

- `prebindgen-registry/src/registry/tests.rs` — `RegistryBuilder::convert_with`,
  and the recipe compiler's `Cx` and `Compiler` types
- `prebindgen-jni/src/jni/tests/values.rs` — both JObject input-plan builders
- `prebindgen-jni/src/jni/tests/callbacks.rs` — the callback input planner

The remaining prose references to `Emit` across both adapters, both docs files
and the README are updated to name `RustWriter` or the writer.

## Verification

- Generated artifacts under `examples/` are byte-identical. The build
  regenerates them and `git status` reports no change to that directory.
- `cargo test --workspace` passes: 28 test binaries, no failures.
- `cargo clippy --workspace --all-targets` and `cargo doc --workspace --no-deps`
  report no warnings.
- The compile-fail doctest pinning the private constructor asserts `E0624`
  specifically, so it cannot start passing because of an unrelated compile
  error.

— Claude Code (Opus 5)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ runtime crate (~350 lines), not the generator.
 | `prebindgen` | Base: reads what `#[prebindgen]` captured and hands out `(syn::Item, SourceLocation)` pairs via `Source` | — |
 | `prebindgen-proc-macro` | The `#[prebindgen]` macro itself | `prebindgen` |
 | `prebindgen-flat` | Independent flat parser/model and Rust-rendering protocol | `prebindgen` |
-| `prebindgen-registry` | Collector/resolver; owns the callback-only `Emit` key | `prebindgen-flat`, `prebindgen` |
+| `prebindgen-registry` | Collector/resolver; owns the callback-only `RustWriter` | `prebindgen-flat`, `prebindgen` |
 | `prebindgen-c` | C / cbindgen adapter (`CbindgenBuilder`) — experimental proof of concept | `prebindgen-registry` |
 | `prebindgen-jni` | JNI / Kotlin adapter (`JniGenBuilder`) | `prebindgen-registry`, [`kotlin-codegen`](https://github.com/milyin/kotlin-codegen) |
 | `prebindgen-c-runtime` | Leaf, ~65 lines, no deps — traits the generated C converters call at run time | — |
--- a/docs/crate-split.md
+++ b/docs/crate-split.md
@@ -38,7 +38,7 @@ prebindgen                BASE. Source stream only. No adapters, no registry.
 prebindgen-flat           the flat model + TypeKey + the RustEmitter protocol
   deps: prebindgen
 
-prebindgen-registry       language-agnostic pipeline + Emit key + shared decl vocabulary
+prebindgen-registry       language-agnostic pipeline + RustWriter + shared decl vocabulary
   deps: prebindgen-flat
 
 prebindgen-jni            JNI/Kotlin generator
@@ -88,10 +88,10 @@ The boundary is therefore split into protocol and key:
   methods emit source types from `TypeKind`/identities, constructors from
   `FieldShape`, and private output fragments recorded by Flat. It has no method
   returning a captured spelling or typed source AST.
-- `prebindgen-registry::Emit` is the concrete, zero-sized registry key. Its
-  constructor is registry-private and it carries no qualification or model state.
-- `prebindgen-registry::RustWriter` owns the frozen source-module map and the
-  private `Emit` token. Emission callbacks receive `&RustWriter`; its narrow
+- `prebindgen-registry::RustWriter` is the concrete registry capability. Its
+  constructor is registry-private, it owns the frozen source-module map, and it
+  implements the protocol through a private zero-sized receiver that is unnamed
+  in any public API. Emission callbacks receive `&RustWriter`; its narrow
   surface generates inert fragments and never returns a captured or typed AST.
 
 This preserves the dependency and mental model:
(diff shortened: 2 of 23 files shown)


milyin:
Reviewed `71398d73` against `feature/chained-rust-generation`.

## The vacuity claim is the point, and it is true

The PR's central argument is that six source fences held out the string `Emit`, and deleting `Emit` would make them pass regardless of what the code did. That is exactly the failure mode worth proving rather than asserting, so I proved it: on this head, with `Emit` gone, I restored the old `contains("Emit")` assertion on `convert_with` **and** gave the planner a `RustWriter` parameter — the very thing the fence exists to forbid — and made it compile.

```
test registry::tests::conversion_planning_cannot_obtain... ok
```

The old fence passes while the capability is reachable. So the vacuity was real, not theoretical, and re-pointing the assertions is not cosmetic — it is what keeps six checks alive.

The replacements bite. Adding `Option<&RustWriter>` to `convert_with` and fixing all four call sites so the crate compiles fails `conversion_planning_cannot_obtain_the_writer`; the same shape on `build_jobject_struct_input_plan` fails its fence too. Both needed the compiling variant to be meaningful, since the type system rejects the careless version first.

## The E0624 pin is also non-vacuous

`compile_fail,E0624` on `RustWriter::new` is the right form — an unqualified `compile_fail` would pass on any error, including an unrelated one. I checked it is pinned to the privacy specifically: making the constructor `pub` turns that doctest red.

## Verification

`cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `examples/regen-check.sh` byte-identical. `Emit` has no remaining code use anywhere in the workspace — only `RustEmitter`, which is the flat trait and a different thing.

The reasoning about why `RustWriter` must not implement `RustEmitter` is correct and worth keeping in the file: the trait's `emit_source_type` takes a module map from its caller, so a `RustWriter` implementing it would let anyone who can name the trait render against a map other than the frozen one. `Renderer` being private and unstored is the minimum that satisfies the trait's need for a receiver.

## One correction

The body says "Five such assertions held out the string `Emit`", then its own bullet list enumerates six sites — `convert_with`, `Cx`, `Compiler`, both JObject builders, and the callback planner. The code agrees with the list, not the number: three assertions in `registry/tests.rs`, two in `values.rs`, one in `callbacks.rs`, and the diff removes six `contains("Emit")` and adds six `contains("RustWriter")`.

Nothing is wrong in the change itself — all six were converted. Worth fixing the sentence before merge, since a reader checking the work will count the bullets.

— Claude Code (Opus 5)

milyin:
Fixed. The count is six, as the enumeration said: three assertions in `registry/tests.rs` (`convert_with`, `Cx`, `Compiler`), two in `jni/tests/values.rs`, one in `jni/tests/callbacks.rs`. The diff removes six `contains("Emit")` and adds six `contains("RustWriter")`.

Corrected in both places it appeared — the PR body and the commit message, which carried the same wrong number.

— Claude Code (Opus 5)

PR #584, opened 2026-08-27
Plan the exported C wrappers as artifacts

milyin:
## What this changes

The exported wrapper — the `#[no_mangle] extern "C"` function a declared
`#[prebindgen]` function becomes — was built inside `Prebindgen::on_function`,
which `write_rust` calls once per declared item while it assembles the file.
The wrapper therefore came into existence during writing, and read what it
needed off `CbindgenBuilder` at that moment: the exported symbol, the path it
calls the source function by, whether `.panic()` was declared, and which
parameters name an opaque resource the alias preflight compares.

It is now planned once, immediately after the C generation plan is frozen, and
joins the assembly #583 introduced as a `CFinalArtifact::Wrapper`. The assembly
holds converters first, in the plan's dependency order, then one wrapper per
declared function in source-name order — the order `write_rust` emitted them in
before, which is why the generated file does not move.

## What a planned wrapper holds

`CWrapper` retains its `flat::Function`, its symbol, its call path, its
`.panic()` flag, and one alias slot per parameter — the resource that argument
names and how the call uses it, `None` when the parameter names no single owned
resource. Deciding a slot needs the set of declared opaque types, which is a
binding question, so it is answered while planning.

For its boundary sites it keeps an `Rc` to the frozen `GenerationPlan`. That is
a lookup in an immutable plan, not a resumption of one: `CbindgenBuilder`, the
`Registry` and the compilation cache are all absent from the artifact, which
`a_planned_wrapper_holds_no_declaration_object` now fences.

One check moves rather than staying put. That a `Result`-returning function's
error type was declared with `.error()` was asserted while rendering the error
wiring; it is asserted while planning the wrapper now, with the same message.

## Scope

`CbindgenBuilder::on_function` returns nothing and is now dead weight, but the
trait method stays until JniGen's wrappers follow — deleting it needs both
adapters off it. That is the next PR, and step 2 of #581 is split into these
two.

## Verification

generated artifact in the tree — C bindings, C headers, JNI Rust, Kotlin — is
byte-for-byte unchanged.

`ordinary_wrapper_rendering_cannot_resume_legacy_planning` moves with the
renderer: it now reads `assembly.rs` and `CWrapper::render_fn`, still forbids
the five legacy planning entry points, and requires the frozen-plan reads
(`self.site(`, `failure_route()`, `planned_inputs(`, `.encode(`). The rename
from `emit_planned_inputs` to `planned_inputs` is the only required token that
changed spelling.

Step 2a of #581.

— Claude Code (Opus 5)


milyin:
Reviewed `823caaed` against `feature/registry-driven-rust-assembly`. Step 2a of #581.

## Reproduced

Workspace tests, strict Clippy, `cargo fmt --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `regen-check` byte-identical.

I also ran `examples/smoke-asan.sh`, which passes clean under ASan/LSan/UBSan. That was worth doing rather than skipping: this moves the alias preflight and the error-wiring assertion, and the smoke test's own summary line names "alias preflight" among the things it exercises. Ownership behaviour is unchanged where it is hardest to see from a diff.

## The two moved checks are still live

**The `.error()` assertion.** Deleting it from `CWrapper::new` fails `result_error_not_declared_is_build_error`, so relocating it from render-time to plan-time kept its coverage. It also fires earlier now — that test calls `build_with(..).and_then(write_rust)` and is satisfied by a panic at either point, so the move is invisible to the test but real for a user, who gets the diagnostic before any file is written.

**The no-declaration-object fence.** Adding an `Option<Rc<CbindgenBuilder>>` field to `CWrapper` and supplying it at the constructor so the crate compiles fails the fence with the message it was written for. Parsing the struct's fields with `syn` rather than grepping the file is the right form — it cannot be fooled by the name appearing in a doc comment or a nearby function.

## On the split, and on step 5

Splitting step 2 into 2a and 2b is a good call: `on_function` cannot go until both adapters are off it, and doing C alone keeps this diff at 5 files with a byte-identical result. Leaving the now-dead `CbindgenBuilder::on_function` in place until 2b is the honest way to do that.

One thing to carry into 2b and beyond. `CWrapper` retains its function, symbol, call path, panic flag and alias slots — but not the converters its body calls. Step 5 of #581 rests on "every caller now an artifact that states its callees", and after this PR the C wrapper is an artifact that does not yet state them. Nothing is wrong here — the umbrella's step 5 only needs the edges to exist before `validate_converter_calls` is deleted, not at step 2 — but it means step 2 as currently written does not by itself satisfy its own role as step 5's prerequisite. Worth knowing whether the callee edges arrive in 2b, in step 5, or in a step that has not been written yet, because that is the ordering the umbrella already identified as load-bearing.

Nothing outstanding.

— Claude Code (Opus 5)

PR #585, opened 2026-08-27
Plan the exported JNI externs as artifacts, and delete on_function

milyin:
## What this changes

The JNI extern — the `#[no_mangle]` function a declared `#[prebindgen]`
function is exported as — was built inside `Prebindgen::on_function` while the
writer walked declared items. Rendering it asked live questions in four places:

- the function's frozen plan, looked up through `Declarations::fn_plan`;
- the origin module of what the extern calls, through `fn_module`;
- a `LiveDeliveryContext` used to qualify the accessor calls a `Return`-delivery
  convert walks to reach its single leaf;
- another `LiveDeliveryContext` used to encode a fallible function's `Err` arm.

All four are answered while planning now, and the extern becomes an artifact of
the assembly. With Cbindgen already off `on_function` (#584), the method is
deleted and the writer no longer walks declared Flat functions at all.

## Frozen deliveries

`FrozenDelivery` is not new: callback trampolines have used it since the
callback artifacts landed. It holds, for one set of delivered leaves, the
origin qualification of the accessor calls that reach them and the Flat shape
of any sum among them — the two model facts a delivery needs at rendering time.

Three plans now carry one, each built where its leaves are decided:
`UnfoldOutputPlan` (builder/fold delivery), `ErrorOutputPlan` (the `Err` arm),
and `ValueOutputPlan` (only for a `Return`-delivery convert, which reaches its
leaf itself rather than through the leaf encoder).

`LiveDeliveryContext` is deleted, and with it the two `Declarations`-taking
leaf helpers — `leaf_is_prim` and `leaf_wire` — that only it called. Their
frozen counterparts already existed; the primitive-`jvalue` rule that lived in
the deleted `leaf_is_prim` moves to the frozen one as documentation.

## What a planned extern holds

`JWrapper` retains the `Rc<JniFunctionPlan>` for its function, the model
element it exports, and a `JCallee` — a source path for a function, or a
binding-defined expression for the nullary getter a declared constant is
exported through. Its renderer takes `&self` and the `RustWriter`.

Externs join the assembly at the end of resolution: converters first, in the
plan's dependency order, then one extern per declared function in source-name
order. That is the order `write_rust` emitted them in before, so the generated
file does not move. Constant getters still reach the renderer through
`on_const`, which is step 3 of #581, and they build a `JWrapper` the same way.

## Verification

unchanged.

Two fences follow the JNI renderer, matching the ones #584 added on the C side:
`a_planned_extern_holds_no_declaration_object` parses `JWrapper`'s fields and
rejects `Declarations`, `Registry`, `Compiled` and `RefCell`, and
`extern_rendering_asks_the_registry_nothing` pins the renderer's parameter list
to `&self` plus the writer and rejects the two lookups it used to make.

The registry's own writer tests keep their coverage on a different item kind:
the two that need an adapter-emitted caller (a late plan activated during
per-item emission, and a call to a converter reachability filtering dropped)
now emit that caller from `on_struct`, since `on_function` is gone.
`write_rust_sorts_declared_items_by_ident` asserts no function reaches the file
from an empty assembly, which is the new rule stated as a test.

Step 2b of #581.

— Claude Code (Opus 5)


milyin:
Reviewed `dd69aa6c` against `feature/registry-driven-rust-assembly`. Step 2b of #581.

## Reproduced

Workspace tests, strict Clippy, `cargo fmt --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `regen-check` byte-identical, and the covertest JVM run at 61 sections — worth running here because this moves the `Err`-arm encoding and the `Return`-delivery accessor walk, which are behaviour the Rust build alone would not exercise.

`on_function` occurs nowhere in the workspace, and `write_rust` no longer walks declared Flat functions.

## The re-pointed registry tests keep their teeth

This is the part I most wanted to check, because moving a test's fixture from one callback to another is the easy way to make it stop testing anything. Both survive:

- **`per_item_planning_precedes_late_converter_filtering`** — the property is that an artifact activated *during* per-item emission is still seen as reachable. Snapshotting `reachable()` before the per-item walk instead of after fails it. My first attempt at that mutation placed the snapshot after the walk and the test passed, which is the correct answer for a mutation that does not actually change the ordering; moving it genuinely earlier turns the test red.
- **`a_call_to_a_filtered_converter_is_a_writer_error`** — disabling `validate_converter_calls` fails it.

So re-homing both on `on_struct` changed the fixture, not the coverage.

## The two new JNI fences bite

- **`a_planned_extern_holds_no_declaration_object`** — adding an `Option<Rc<Declarations>>` field to `JWrapper`, with the constructor fixed so the crate builds, fails with "a planned extern retains Declarations, so rendering could resume planning".
- **`extern_rendering_asks_the_registry_nothing`** — this one is stronger than its C counterpart. It `assert_eq!`s the renderer's whole parameter list against `&self,emit:&prebindgen_registry::RustWriter`, so *any* added parameter fails by construction rather than only a named type. And the body half bites: planting `fn_module(` in the renderer fails with "the extern renderer resumed planning through fn_module(".

Pinning the exact parameter list is the right instrument for a renderer whose contract is "receives one frozen artifact and the writer, nothing else" — a forbidden-substring list would have to guess which live object a future author reaches for.

## Note

`LiveDeliveryContext` and its two `Declarations`-taking helpers are deleted, and the primitive-`jvalue` rule that lived in `leaf_is_prim` is preserved as documentation on the frozen counterpart rather than dropped. That is the kind of thing that goes missing silently in a deletion, so it is good it is called out in the description and true in the diff.

Nothing outstanding.

— Claude Code (Opus 5)

PR #586, opened 2026-08-27
Plan constants as artifacts, and delete the item walk

milyin:
## What this changes

After #585 the writer still walked the model four times — declared structs, the
two enum shapes, declared constants — calling back into the adapter for each
item it found. Both adapters returned nothing for the three type kinds. Only
constants produced anything, and both adapters had a policy for them.

Constants become artifacts, and the walk goes.

- **Cbindgen** plans one `CFinalArtifact::Const` per captured constant, in
  source-name order. It renders the alias to the source item that `on_const`
  rendered: the initializer is never copied, because it may name source-crate
  internals, and cbindgen cannot evaluate a path initializer, so an aliased
  constant does not become a `#define`.
- **JniGen** plans one `JFinalArtifact::Const` per *declared* constant, carrying
  both halves it emitted before — the alias, and the nullary getter extern the
  Kotlin `val` is initialized from. The getter is a `JWrapper`, the artifact
  #585 made of every other extern, built with the constant's path as its callee.

Which constants reach the file is now a property of which artifacts an adapter
planned. JniGen filters by its own declarations; Cbindgen has no constant
declaration mechanism and plans them all. The registry-side `declared.consts`
gate the writer applied is deleted along with the walk that applied it, as is
`Registry::declared` — nothing else read it — and the name-sorting helper.

One check moves, as the `.error()` assertion did in #584: that a declared
constant's type is not an opaque handle is asserted while planning the artifact
rather than while rendering it. `handle_const_rejected` covers either timing,
since it drives `build_with(..).and_then(write_rust)`.

## What is left of `Prebindgen`

`prerequisites`, `validate`, `validate_resolved`, `source_module`. No method on
it is per `#[prebindgen]` item any more, and `write_rust` reads the model for
exactly one thing: prebindgen's own anonymous feature-check consts.

## Guards stay in the writer

They are prebindgen's injected feature checks, not adapter output: identical for
every adapter, named by none — so an adapter cannot declare one — and last in
the file by construction. Planning them into one adapter's assembly would move
registry knowledge into adapters and buy nothing. This is the "where applicable"
in #581's success criterion about guards.

## Verification

unchanged: constants land where the walk put them, because the artifacts are
added to the assembly after the wrappers.

Two writer tests move with what they test:

- The caller that races reachability filtering (in both
  `per_item_planning_precedes_late_converter_filtering` and
  `a_call_to_a_filtered_converter_is_a_writer_error`) now comes from
  `prerequisites` — the last adapter output still produced while the file is
  written. The first test's position assertion flips accordingly, since
  prerequisites precede the assembly; what it pins is unchanged, that a plan
  marked reachable during writing survives the filter.
- `write_rust_sorts_declared_items_by_ident` becomes
  `declared_items_reach_the_file_only_as_artifacts`: same eight declared items
  of all four kinds, an empty assembly, and an assertion that none of them
  reaches the file. The old test pinned the walk's sort order; there is no walk
  to sort, and this is the rule that replaced it.

Step 3 of #581.

— Claude Code (Opus 5)


milyin:
Reviewed `c11f9a06` against `feature/registry-driven-rust-assembly`. Step 3 of #581.

## Reproduced

Workspace tests, strict Clippy, `cargo fmt --check`, `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --no-deps`, `regen-check` byte-identical, covertest at 61 sections.

All four remaining `on_*` methods occur nowhere in the workspace. `Prebindgen` is down to `prerequisites`, `validate`, `validate_resolved` and `source_module` — none of them per-item.

## The twice-moved writer tests still hold

These two have now been re-pointed twice — `on_function` to `on_struct` in #585, `on_struct` to `prerequisites` here — which is the pattern where a test quietly stops testing. Both still fail when the property breaks:

- `per_item_planning_precedes_late_converter_filtering` — snapshotting `reachable()` before `prerequisites` runs turns it red. The position assertion flipping (prerequisites precede the assembly) is a change in where the caller lands, not in what is pinned.
- `a_call_to_a_filtered_converter_is_a_writer_error` — disabling `validate_converter_calls` turns it red.

## The replacement test is the stronger one

`write_rust_sorts_declared_items_by_ident` pinned the order of a walk that no longer exists, so replacing rather than deleting it was the right call, and `declared_items_reach_the_file_only_as_artifacts` pins the rule that took its place. It is not vacuous: reintroducing a walk that emits declared structs verbatim fails it with

```
`AStruct` reached the file from an empty assembly
```

Eight declared items across all four kinds against an empty assembly is the right fixture — it would catch a walk over any of them, not just the kind that happened to regress.

## The moved handle-const assertion is covered

Same check as on #584's `.error()` move. Neutering `reject_handle_key`'s `is_handle` computation fails `handle_const_rejected`, so moving the assertion from render-time to plan-time kept it live. Worth noting my first mutation neutered `reject_handle_const` instead — the wrapper, not the assertion — and the test passed, correctly, because the real check is one call deeper.

## On leaving guards in the writer

The argument is sound and worth keeping: prebindgen's injected feature checks are identical across adapters, named by none, and last by construction, so planning them into an adapter's assembly would move registry knowledge outward for nothing. The umbrella's success criterion says "where applicable" about guards, and this is the case that phrase was reserving.

The `declared.consts` gate and `Registry::declared` going with the walk is the right kind of deletion — the filter existed only to serve the walk, so which constants reach the file becomes a property of which artifacts exist, exactly as step 3 said it would.

Nothing outstanding.

— Claude Code (Opus 5)

PR #587, opened 2026-08-27
Plan the C prerequisites as artifacts

milyin:
## What this changes

`Prebindgen::prerequisites` is the last callback the writer makes into an
adapter. Cbindgen answered it by building nine sections while the file was
written, five of them from live registry queries:

| section | what it asked the registry |
|---|---|
| allocator, C-string block, universal freer | whether the layer hands `char*` or array memory to C |
| array builder | whether a `Vec<T>` return hands out a block |
| opaque handles, value-opaques, tagged unions, callbacks | the frozen generation plan, filtered by artifact kind |
| data-struct mirrors | which declared data structs cross, and each field's wire |
| enum mirrors | which declared enums cross, and their values |
| domain constants | how deep the declared conversions nest optionals |

All nine become artifacts of the assembly, planned when the generation plan is
frozen and placed ahead of the converters in exactly the order they were
emitted in — which is why the generated file does not move.

## The five artifact kinds

- **`CMemory`** carries the memory helpers as one artifact, because they are
  one concern: memory handed to C. It is planned only when that happens, so the
  panic for a missing `.free_memory_function` fires while planning now.
- **`CPlanned`** places one artifact the generation plan already holds —
  handle, value-opaque, tagged union, callback — keeping the plan's order
  within each kind. The payload stays the plan's; the artifact is the identity
  to look it up by, plus an `Rc` to the plan.
- **`CDataStruct`** freezes its fields in wire form. Deciding a field's wire
  needs the binding's declarations, so it is decided while planning.
- **`CEnum`** freezes the model's own values instead, and the writer spells
  them. That is deliberate: a C enum mirror re-states each discriminant **as
  written**, so `= 0x07` stays `0x07` and a `const`- or `cfg`-driven expression
  survives. Resolving each to a number while planning would narrow the domain
  to what `i64` and a literal can express.
- **`CDomainConstant`** is one reserved representation value.

## Scope

`Prebindgen::prerequisites` stays until JniGen's prerequisites follow — its
error channel, handle destructors, vector-build helpers and constant-expression
getters. That is the next PR, and it is what removes the trait's last emission
method. Step 4 of #581 is split into these two, as step 2 was.

## Verification

unchanged.

`render_artifacts` is deleted with the callback that called it, and the fence
that pinned its signature —
`callback_renderer_accepts_only_the_frozen_plan` — becomes
`planned_artifacts_reach_the_file_through_the_frozen_plan`, which parses
`CPlanned`'s fields and asserts they are a frozen plan and an identity, with no
`CbindgenBuilder`, `Registry`, `Compiled` or `RefCell` among them. Pinning the
struct rather than a function type is the same instrument #584 and #585 use for
the wrapper and the extern.

Step 4a of #581.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-c/src/assembly.rs
+++ b/prebindgen-c/src/assembly.rs
@@ -24,6 +24,18 @@ pub(crate) enum CFinalArtifact {
     Wrapper(Box<CWrapper>),
     /// One captured constant, re-stated as an alias to its source.
     Const(Box<CConst>),
+    /// The memory helpers the generated layer hands `char*` and array blocks
+    /// to C through.
+    Memory(Box<CMemory>),
+    /// One `#[repr(C)]` mirror of a declared data struct.
+    DataStruct(Box<CDataStruct>),
+    /// One `#[repr(C)]` mirror of a declared fieldless enum.
+    Enum(Box<CEnum>),
+    /// One reserved representation value a generated sum-type ABI uses.
+    DomainConstant(Box<CDomainConstant>),
+    /// One artifact the registry generation plan already holds: an opaque
+    /// handle, a value-opaque, a tagged union, or a callback.
+    Planned(Box<CPlanned>),
 }
 
 impl RustArtifact for CFinalArtifact {
@@ -32,6 +44,11 @@ impl RustArtifact for CFinalArtifact {
             Self::Converter(converter) => converter.key(),
             Self::Wrapper(wrapper) => wrapper.key(),
             Self::Const(constant) => constant.key(),
+            Self::Memory(memory) => memory.key(),
+            Self::DataStruct(item) => item.key(),
+            Self::Enum(item) => item.key(),
+            Self::DomainConstant(item) => item.key(),
+            Self::Planned(item) => item.key(),
         }
     }
 
@@ -40,6 +57,11 @@ impl RustArtifact for CFinalArtifact {
             Self::Converter(converter) => converter.render(emit),
             Self::Wrapper(wrapper) => wrapper.render(emit),
             Self::Const(constant) => constant.render(emit),
+            Self::Memory(memory) => memory.render(emit),
+            Self::DataStruct(item) => item.render(emit),
+            Self::Enum(item) => item.render(emit),
+            Self::DomainConstant(item) => item.render(emit),
+            Self::Planned(item) => item.render(emit),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


milyin:
Reviewed `0b3e7639` against `feature/registry-driven-rust-assembly`. Step 4a of #581.

## Reproduced

generated files are unchanged, the C smoke test is clean under
ASan/LSan/UBSan, and `covertest-kotlin` runs its 61 sections.

The five artifact kinds are the right cut, and `CEnum` retaining
`EnumValue` rather than a resolved number is the right call for the reason
the doc comment gives.

Four things below. The first is a hole in the new test, the second is a gap in
what the change is checked by, the third is a lost diagnostic, and the fourth
is older than this PR and is flagged because `CDomainConstant` now carries it
one layer further.

## The new test has one escape

`planned_artifacts_reach_the_file_through_the_frozen_plan` does bite: giving
`CPlanned` a `RefCell` field fails it with

```
a placed artifact retains RefCell, so rendering could resume planning
```

But one of the four forbidden strings is `"Registry,"`, with the comma, and
the field list is compared after whitespace is stripped. A `Registry` field
written last without a trailing comma therefore passes:

```rust
pub(crate) struct CPlanned {
    generation: Rc<GenerationPlan<crate::compile::CRepresentation>>,
    id: prebindgen_registry::generation::ArtifactId,
    registry: *const Registry
}
```

That is the one field the test most wants to forbid, and the test is green on
it. `cargo fmt` would restore the comma, but a test that only holds because a
separate gate runs is not holding. Stripping `prebindgen_registry` from the
field string and then matching bare `Registry` closes the escape, and keeps
the reason the comma was there in the first place.

## The value-opaque placement is not checked by anything

Reordering the enum mirrors ahead of the data-struct mirrors moves all five
committed generated files by 100 lines, so the regeneration check does pin
placement. It does not pin every section. `example-cbindgen` plans no
value-opaque at all — instrumenting the `planned` closure reports
`c-opaque-handle 1`, `c-value-opaque 0`, `c-tagged-union 2`, `c-callback 4` —
so swapping the value-opaque placement past the data-struct mirrors leaves
every generated file byte-identical and all 110 `prebindgen-c` tests green.

Of the eight sections this PR places, the value-opaque is the one nothing
pins. Declaring one value-opaque in the example would pin it, and would also
give the `c-value-opaque` rendering path its first end-to-end exercise.

## A duplicate identity now drops silently

`AssemblyBuilder::artifact` replaces a stored artifact only when the stored one
is unreachable, so two reachable artifacts sharing a key keep the first and
discard the second. Forcing `CDataStruct::key` to a constant shows what a
reader would see:

```
error[E0425]: cannot find type `foo_t` in this scope
    --> examples/example-cbindgen/generated/example_flat_x86_64.rs:1472:16
```

Before this PR the same collision produced a duplicate definition naming
`foo_t` twice. Both are errors in the generated crate rather than a prebindgen
diagnostic, so nothing is newly broken — but a missing type is the harder of
the two to trace back to two declared structs whose short names mangle alike.

A narrow guard is available and costs nothing. Instrumenting the two examples
to report every reachable duplicate gives only `ArtifactKey::Operation` —
converters legitimately reached from several sites — and never
`ArtifactKey::Artifact`. Asserting on a repeated reachable
`ArtifactKey::Artifact` is accordingly green across the workspace tests, both
examples and the regeneration check:

```rust
if let ArtifactKey::Artifact(id) = artifact.key() {
    assert!(
        !(self.artifacts[position].reachable() && artifact.reachable()),
        "two reachable artifacts share the identity {id}"
    );
}
```

That is `AssemblyBuilder`'s to add rather than this PR's, since every adapter
that names its own artifacts has the same exposure. Raising it here because
this PR is what gives the C adapter five kinds that name themselves.

## The converter filter now ends in a wildcard

`CbindgenBuilder::converters` matched all three variants before and now reads
`_ => None`. With eight variants the wildcard is the shorter code, but the
match is a filter over artifact kinds, so the next kind that should be visible
to a converter caller will be filtered out with nothing to say so. Listing the
five new variants next to `Wrapper` and `Const` keeps the compiler as the
reminder.

## `RepresentationDomain` decides meaning by spelling

This is not introduced here. `CDomainConstant` stores `ty: syn::Type` taken
from `domain.ty()`, so the spelling now survives into a planned artifact,
which is why it is worth naming now.

`RepresentationDomain` keeps its scalar as a `syn::Type` and recovers the kind
by matching the rendered token string. It does so twice: in `contains_expr`,
to decide whether a bounds check needs a NaN test, and in `niche_values`, to
pick the integer or float candidate generator, ending `_ => unreachable!()`.
The kind is not missing. `DomainScalar` produces it as `ScalarValue::$v` in
the same macro arm that produces the spelling, and throws it away.

A third site is in JniGen, which compares the domain's spelling against the
representation's, and it is the one that shows the consequence is not
theoretical. Spelling one integer `::core::primitive::u64` instead of `u64` —
the same type — fails the build with

```
representation u64
```

So three sites decide equality of meaning by equality of text, and the one
that reports it is in the adapter this PR does not touch. `DomainScalar` is
publicly re-exported from `prebindgen_registry` and is not sealed, so a
downstream implementation choosing any other spelling of the same scalar
reaches the `unreachable!()`.

The fix is to carry the tag `DomainScalar` already computes rather than
re-deriving it from text. Flat's `ScalarKind` is not a drop-in — it has
`Bool`, `Isize` and `Usize`, and lacks `I128` and `U128` — so this needs
either a domain-side kind or a widened `ScalarKind`. Either way both string
matches and the `unreachable!()` go.

Worth its own issue rather than this PR. Nothing here blocks #587.

— Claude Code (Opus 5)

milyin:
Fixed the two that belong here, and filed the other two.

**The fence's comma escape** — fixed in `1c679880`. The fence now matches the bare `Registry`, as the two sibling fences already do; `prebindgen_registry` cannot collide with it because the crate path is lowercase, which is why the comma was never needed. Checked against the exact escape from the review: a trailing `registry: *const Registry` with no comma now fails with "a placed artifact retains Registry, so rendering could resume planning".

**The wildcard** — fixed in the same commit. `converter_functions` lists the seven non-converter kinds again, so the next kind that should be visible there is a compile error rather than a silent filter.

**The duplicate-identity guard** — agreed it is `AssemblyBuilder`'s rather than this PR's, since every adapter naming its own artifacts has the same exposure. It goes into step 4b, which is the next registry-touching change, with the assertion restricted to a repeated reachable `ArtifactKey::Artifact` exactly as written in the review.

**The two things that are not fixes** are filed rather than folded in:

- #588 — nothing exercises or pins the C value-opaque artifact. It needs a type and a crossing added to `example-flat` and `example-cbindgen`, and the flat crate deliberately has no `#[repr(C)]`, so it is `opaque_data_struct`/`opaque_owned_struct` rather than `repr_c_struct`. That is an addition to the example surface, not a correction to this change.
- #589 — representation domains decide type identity by spelling, in three places, one of which ends in an `unreachable!()` a downstream `DomainScalar` can reach. It records that the tag already exists at the point the spelling is produced, and that `ScalarKind` is not a drop-in replacement.

— Claude Code (Opus 5)

PR #591, opened 2026-08-27
Plan the JNI prerequisites as artifacts, and delete prerequisites

milyin:
## What this changes

JniGen answered `Prebindgen::prerequisites` with four groups built while the
file was written:

- the **prelude** every extern body calls by bare name — the `__JniErr` alias,
  the `OwnedObject` carrier borrowed-handle plans return, and the two
  error-channel functions;
- one **destructor** per opaque handle, gated on the handle being referenced by
  a scanned `#[prebindgen]` function;
- one **`…VecNew/Push/Free` trio** per flattenable element type a `&[T]`/`Vec<T>`
  parameter takes;
- one nullary **getter** per binding-defined constant expression.

All four become artifacts, planned while the generation plan is frozen and
placed ahead of the converters in the order they were emitted in.

`JHandleDestructor` retains the handle's source type as a reading — the writer
spells it, since only the writer knows how to qualify it — plus the exported
symbol, which is both its identity and its position. `JVecBuild` retains the
frozen element flatten plan and the three symbols, mangled while planning. The
constant-expression getters are `JWrapper`s, like every other extern, and the
check that a constant expression's type is not an opaque handle moves to
planning with them.

One subtlety: `plan_vec_build_helpers` reads the planning store directly rather
than going through `collect_vec_build_helpers`, because it runs *during*
`JniGenerationPlan::freeze`, before the frozen plan is installed on the
declarations. The frozen reader is untouched — Kotlin emission uses it, and
routing the planner through it produced Kotlin missing every `Vec` helper.

## `Prebindgen` has no emission method left

With both adapters off it, `prerequisites` is deleted. Nothing an adapter
implements returns Rust items for the writer to place any more, which
`the_adapter_emits_no_items_and_the_writer_reparses_none` states as a test.
What remains on the trait is `validate`, `validate_resolved` and
`source_module`.

`write_rust` still takes `&impl Prebindgen`, for `source_module` alone. Taking
the frozen graph and nothing else is step 6.

## The duplicate-identity guard from #587's review

`AssemblyBuilder::artifact` now refuses two *reachable* artifacts under one
identity. Keeping either would silently drop an item the file needs, and the
reader would meet it as a missing name in the generated crate rather than as a
diagnostic here. Converters are exempt, since one registry operation is
legitimately reached from several sites — that is what the de-duplication
exists for. `two_reachable_artifacts_may_not_share_an_identity` covers it.

## Verification

unchanged.

Keeping the JNI Rust identical needed one deliberate choice. A handle's
destructor and its alignment assertion used to be two entries in one
symbol-sorted list, `Java_…_freePtr` before `Java_…_freePtr__align_assert`, so
the artifact renders them in that order — the assertion second. An anonymous
const asserts from wherever it stands, so this is placement, not meaning.

The registry's two writer tests move to the assembly, since there is no
callback left to emit their caller from. Both now state their property
directly:

- `an_artifact_reached_after_freezing_is_emitted` — an artifact frozen dormant
  and reached afterwards is still emitted. That is not hypothetical: JniGen
  shares one reachability cell between every clone of a plan and marks it when
  a parent that calls it is compiled, which is why the writer asks each
  artifact whether it is reachable as it renders.
- `a_call_to_a_filtered_converter_is_a_writer_error` — the caller is now an
  artifact of the test assembly rather than callback output, and the assertion
  on the reported `(caller, missing converter)` pair is unchanged.

Step 4b of #581.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/emit/vec_build.rs
+++ b/prebindgen-jni/src/jni/emit/vec_build.rs
@@ -163,7 +163,7 @@ pub(crate) fn vec_build_elem(
 /// takes as a `&[T]`/`Vec<T>` input — the set the synthetic `…VecNew/Push/Free`
 /// externs are emitted for (once per type, shared across all such functions).
 /// Deduped by [`TypeKey`] and sorted for deterministic output (mirrors
-/// [`build_handle_destructor_items`]).
+/// [`crate::jni::plan_handle_destructors`]).
 ///
 /// The key is the **canonical** element's, since that is what
 /// [`vec_build_elem`] answers with — so `Vec<Payload>` and `Vec<Box<Payload>>`
@@ -262,7 +262,7 @@ fn vec_helper_symbol(ext: &Declarations, base: &str, suffix: &str) -> String {
 
 /// One `#[no_mangle] extern "C"` `…VecNew/Push/Free` trio per flattenable
 /// element type used as a `&[T]`/`Vec<T>` input — the Rust half of the
-/// build-the-Vec-incrementally path. Modeled on [`build_handle_destructor_items`]
+/// build-the-Vec-incrementally path. Modeled on [`crate::jni::plan_handle_destructors`]
 /// (deterministic symbol sort, emitted only for element types a scanned function
 /// actually takes by slice/Vec).
 ///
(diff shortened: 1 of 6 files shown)


milyin:
Reviewed `70b8f445` against `feature/registry-driven-rust-assembly`. Step 4b of #581.

## Reproduced

generated files are unchanged, the C smoke test is clean under
ASan/LSan/UBSan, and `covertest-kotlin` runs 61 sections whose names match the
base branch's exactly.

`Prebindgen` returning no items at all is the milestone this PR is for, and
`the_adapter_emits_no_items_and_the_writer_reparses_none` counting to zero is
the right way to hold it.

## Both moved registry tests still bite

These are the two I flagged on #586 as the ones to watch: they had been
re-pointed twice already, and this PR rewrites rather than re-points them,
which was the right call once no callback was left to host them.

Filtering unreachable artifacts out inside `AssemblyBuilder::build` — the
natural way to break the property — fails both:

```
test write::tests::an_artifact_reached_after_freezing_is_emitted ... FAILED
test write::tests::a_call_to_a_filtered_converter_is_a_writer_error ... FAILED
```

## The vec-helper phase dependency is fenced, not just documented

The PR body flags that `plan_vec_build_helpers` reads the planning store
directly because it runs before `JniGenerationPlan::freeze` installs the frozen
plan. That is worth more than a comment, because `freeze` does not merely
finish after the planner — it `take`s the same store, so a later reordering
leaves the planner reading an empty map rather than an absent one.

Emptying the store before the planner reads it does get caught, loudly:
`cross_artifact_flatten_vec_callback_builder_agree`,
`native_symbols_are_jni_escaped` and `slice_input_builds_vec_handle` all fail,
and the regeneration check moves two `generated_bindings.rs` files. So the
comment marks a hazard that is already held by tests.

## The reachability test is weaker than its own doc comment

`an_artifact_reached_after_freezing_is_emitted` pins that reachability is not
snapshotted when the assembly is frozen. It does not pin what its doc comment
claims:

> The writer asks each artifact whether it is reachable while it renders, which
> is what makes that legal.

Reading reachability once, just before the render loop, instead of alongside
each `render` call:

```rust
let snapshot: Vec<bool> = assembly.artifacts().map(|a| a.reachable()).collect();
```

leaves every workspace test green and the regeneration check clean. The old
test caught that, because it activated the plan from inside `prerequisites`,
which ran during `write_rust`.

Nothing marks an artifact reachable during rendering any more, so the weaker
property may be all that is needed — but then the doc comment should say what
the test holds. If the render-time read is meant to stay load-bearing, an
artifact whose `render` marks a later one reachable would pin it.

## The guard's exemption is the untested half

`two_reachable_artifacts_may_not_share_an_identity` does bite — neutering the
assertion turns it red. It covers only the prohibition. The exemption that the
doc comment states, that one registry operation may legitimately arrive twice
from several sites, has no test.

Widening the assertion from `ArtifactKey::Artifact` to every key leaves all 216
`prebindgen-registry` unit tests green. Only the examples catch it, at build
time:

```
two reachable artifacts share the identity Operation(OperationId { owner:
ModelArtifact { carrier: TypeKey("bool"), artifact: ArtifactId { kind:
"c-terminal-input", name: "bool" }, direction: Construct }, role: Converter })
```

That makes the exemption the half that erodes silently, since a unit-test run
would not notice its loss. Adding a second reachable twin to
`registry_operations_are_deduplicated_before_rendering` covers it where the
de-duplication rule is already exercised.

## The handle-destructor documentation moved to the wrong function

`plan_handle_destructors` lost its doc comment, and `render_prelude` gained it.
The two blocks are now fused with no blank line between them, so rustdoc
renders `render_prelude` — the error-channel prelude — under several paragraphs
about the `Java_<package>_<class>_<mangled-freePtr>` symbol scheme and its
Kotlin counterpart, and then explains `__JniErr`.

The paragraph that matters most is the last of the misplaced ones. It is the
reason a destructor is planned only for a handle a scanned function references,
which is what keeps a `#[cfg]`'d-out handle from emitting a destructor for a
type not in scope. That paragraph was edited in this diff, so the move looks
accidental rather than intended. It now sits above a function it does not
describe, and the function it describes has nothing.

`cargo doc -D warnings` passes either way, since both intra-doc links still
resolve.

## Two small ones

`JFinalArtifact::reachable` now reads `_ => true`. That is the wildcard removed
from `CFinalArtifact`'s converter filter one PR ago, for the reason that applies
here too: the next artifact kind whose reachability is conditional gets `true`
by default and nothing says so. Spelling out the other six costs a line each.

`plan_vec_build_helpers` sorts twice — by element key, then by
`JVecBuild::sort_key` — and only the second survives, except where two sort keys
tie. Dropping the first would say that the symbol is what orders the file.

Nothing here blocks the PR. The reachability doc comment and the misplaced
documentation are the two worth fixing before it lands.

— Claude Code (Opus 5)

milyin:
All five fixed in `33dfe120`.

**The reachability doc comment** — the test now says what it holds: reachability is read when the file is written, not snapshotted when the assembly is frozen, and *not* pinned to any point during writing, since nothing marks an artifact reachable while another renders. The paragraph about JniGen's shared cells stays, because that is why the weaker property is the one that matters. I did not add an artifact whose `render` reaches a later one: no adapter does that, and inventing the capability in a test would pin a contract nothing depends on.

**The guard's exemption** — a second reachable twin of one registry operation now joins `registry_operations_are_deduplicated_before_rendering`, where the de-duplication rule is already exercised. Checked the mutation you named: widening the assertion from `ArtifactKey::Artifact` to every key fails that test now instead of only the examples.

**The misplaced documentation** — `plan_handle_destructors` has its docs back, `render_prelude` keeps only its own. You were right that it was accidental: the block survived the deletion of `prerequisites` and fused onto the next function.

**The wildcard** and **the double sort** — both as described. `plan_vec_build_helpers` sorts once, by symbol, which is what ordered the file all along.

— Claude Code (Opus 5)

PR #592, opened 2026-08-27
Prove the file complete from artifact edges

### LINKED ISSUE #437: cbindgen: a Vec callback argument builds an array but the array helper is not emitted

milyin:
## What this changes

`validate_converter_calls` walked every rendered `syn::ItemFn` looking for
calls to converters that reachability filtering had dropped. It was the last
place production code read generated Rust to learn something about the
generation, and by construction it could only run once everything had been
rendered.

Artifacts state their own edges now:

- **`RustArtifact::calls`** returns the artifacts this one's body calls.
  `write_rust` checks, before rendering anything, that every callee of a
  reachable artifact itself reaches the file. There is no default: an artifact
  that calls nothing returns an empty vector, and one that gains a call has to
  answer again — an unanswered edge would be a silently weaker check rather
  than a compile error.
- **`RustArtifact::provides`** is the other half. One artifact may render an
  item belonging to another identity: the C callback artifact renders the
  Invoke helper of a converter operation whose own fragment deliberately
  carries no artifact. It defaults to the artifact's own key.

`WriteError::UnrenderedConverterCalls` becomes `UnreachedDependency` and
reports semantic identities rather than Rust symbols, since the check runs
before any name is allocated. Unreachable artifacts are no longer rendered for
their names alone; they stay in the assembly so a *planned but unreached*
dependency can be told from an *unplanned* one.

## Following what the renderer calls

The edges name what each renderer actually invokes, which is not always the
carrier nearest to hand. Two cases found by the check itself while this was
being written:

- A JNI parameter that is a flattened data class or an allocation-free optional
  pair is rebuilt by **one composed converter of its own**, not through its
  leaf pipeline. Stating the pipeline — the obvious field — named a converter
  the body never calls, and the check reported it against five covertest
  wrappers.
- A C wrapper taking a callback calls that callback's Invoke helper, which no
  converter artifact renders. That is what `provides` exists for; before it,
  the check reported two perftest wrappers.

Both were real gaps in the model of who calls whom, surfaced by the new check
rather than by inspection.

## Proving the edges complete

An edge check is only worth what its edges are worth: an artifact that
under-reports weakens the check without failing anything. So the syn walk does
not disappear — it moves to where it belongs, as evidence rather than as the
mechanism.

`prebindgen_registry::write::assert_edges_cover_rendered_calls` is test support
(like `RustWriter::for_test`, and gated the same way). It renders every
artifact of an assembly and refuses a call to any converter the artifact did
not declare. Both adapters run it on **every binding their test suites build** —
it sits in `Cbindgen::write_rust` and `JniGen::write_rust` under `#[cfg(test)]`,
which covers the C suite's `write` helper and the ~120 JNI call sites at once.

Checked that it bites: clearing `CWrapper::calls` fails 30+ C tests, and
clearing `JWrapper::calls` fails 40+ JNI tests, each naming the caller, the
callee and the undeclared call.

## Verification

unchanged.

`prebindgen-c` gains a dev-dependency on the registry's `testing` feature,
which is what makes the test-support helper and `RustWriter::for_registry_test`
available to its suite. `prebindgen-jni` already depended on it.

## Fixes #437

The array builder was gated on `produces_array`, a scan of declared functions'
return types — the gate that issue's "Why" section names. Deriving the runtime
artifacts from the dependency edges instead deletes that scan, so a `Vec`
delivered to a C callback plans the builder because the callback's own encode
declares it.

`a_run_of_converted_optionals_uses_the_declared_wire` is the regression test:
its binding returns no `Vec` at all, and it asserts the builder is emitted.

Fixes #437.

Step 5 of #581.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-c/Cargo.toml
+++ b/prebindgen-c/Cargo.toml
@@ -17,3 +17,9 @@ prebindgen-registry = { workspace = true }
 syn = { workspace = true }
 quote = { workspace = true }
 proc-macro2 = { workspace = true }
+
+[dev-dependencies]
+# The registry's test support: `RustWriter::for_registry_test` and
+# `assert_edges_cover_rendered_calls`, which every binding these tests build
+# runs through.
+prebindgen-registry = { workspace = true, features = ["testing"] }
--- a/prebindgen-c/src/assembly.rs
+++ b/prebindgen-c/src/assembly.rs
@@ -52,6 +52,26 @@ impl RustArtifact for CFinalArtifact {
         }
     }
 
+    fn provides(&self) -> Vec<ArtifactKey> {
+        match self {
+            Self::Planned(item) => item.provides(),
+            _ => vec![self.key()],
+        }
+    }
+
+    fn calls(&self) -> Vec<ArtifactKey> {
+        match self {
+            Self::Converter(converter) => converter.calls(),
+            Self::Wrapper(wrapper) => wrapper.calls(),
+            Self::Const(constant) => constant.calls(),
+            Self::Memory(memory) => memory.calls(),
+            Self::DataStruct(item) => item.calls(),
+            Self::Enum(item) => item.calls(),
+            Self::DomainConstant(item) => item.calls(),
+            Self::Planned(item) => item.calls(),
+        }
+    }
+
     fn render(&self, emit: &prebindgen_registry::RustWriter) -> Vec<syn::Item> {
         match self {
             Self::Converter(converter) => converter.render(emit),
@@ -118,6 +138,20 @@ impl CWrapper {
                 .map(|param| decls.alias_slot_of(&param.ty))
                 .collect(),
         }
+    }
+
+    /// One boundary site of this wrapper's own function, when the function has
+    /// one in that role. A void return and an infallible function have no
+    /// return or error site.
+    fn site_if_planned(
+        &self,
+        role: prebindgen_registry::recipe::Role,
+    ) -> Option<&prebindgen_registry::generation::SitePlan<crate::compile::CRepresentation>> {
+        let id = prebindgen_registry::generation::SiteId::new(prebindgen_registry::recipe::Site {
+            owner: self.function.name.clone(),
+            role,
+        });
+        self.generation.site(&id)
     }
 
     /// One boundary site of this wrapper's own function.
@@ -549,6 +583,25 @@ impl CWrapper {
 }
 
 impl RustArtifact for CWrapper {
+    fn calls(&self) -> Vec<ArtifactKey> {
+        // One site per parameter, plus the return and the error channel when
+        // the function has them — the same sites the body decodes and encodes
+        // through.
+        let mut calls = Vec::new();
+        let roles = (0..self.function.params.len())
+            .map(|index| prebindgen_registry::recipe::Role::Param { index })
+            .chain([
+                prebindgen_registry::recipe::Role::Return,
+                prebindgen_registry::recipe::Role::Error,
+            ]);
+        for role in roles {
+            if let Some(site) = self.site_if_planned(role) {
+                site.abi().payload().calls(&mut calls);
+            }
+        }
+        calls
+    }
+
     fn key(&self) -> ArtifactKey {
         ArtifactKey::Artifact(
             prebindgen_registry::generation::ArtifactId::new("c-wrapper", self.symbol.to_string())
@@ -590,6 +643,10 @@ impl CConst {
 }
 
 impl RustArtifact for CConst {
+    fn calls(&self) -> Vec<ArtifactKey> {
+        Vec::new()
+    }
+
     fn key(&self) -> ArtifactKey {
         ArtifactKey::Artifact(
             prebindgen_registry::generation::ArtifactId::new(
@@ -645,7 +702,27 @@ impl CPlanned {
     }
 }
 
+impl CPlanned {
+    /// The plan's own artifact description.
+    fn payload(&self) -> &crate::chain::CArtifact {
+        self.generation
+            .artifact(&self.id)
+            .unwrap_or_else(|| panic!("the C generation plan lost artifact {}", self.id))
+            .payload()
+    }
+}
+
 impl RustArtifact for CPlanned {
+    fn provides(&self) -> Vec<ArtifactKey> {
+        let mut provided = vec![self.key()];
+        provided.extend(self.payload().provides());
+        provided
+    }
+
+    fn calls(&self) -> Vec<ArtifactKey> {
+        self.payload().calls()
+    }
+
     fn key(&self) -> ArtifactKey {
         ArtifactKey::Artifact(self.id.clone())
     }
@@ -695,6 +772,10 @@ impl CMemory {
 }
 
 impl RustArtifact for CMemory {
+    fn calls(&self) -> Vec<ArtifactKey> {
+        Vec::new()
+    }
+
     fn key(&self) -> ArtifactKey {
         artifact_id("c-runtime", "memory")
     }
@@ -816,6 +897,10 @@ impl CDataStruct {
 }
 
 impl RustArtifact for CDataStruct {
+    fn calls(&self) -> Vec<ArtifactKey> {
+        Vec::new()
+    }
+
     fn key(&self) -> ArtifactKey {
         artifact_id("c-data-struct", self.c_struct.to_string())
     }
@@ -877,6 +962,10 @@ impl CEnum {
 }
 
 impl RustArtifact for CEnum {
+    fn calls(&self) -> Vec<ArtifactKey> {
+        Vec::new()
+    }
+
     fn key(&self) -> ArtifactKey {
         artifact_id("c-enum", self.c_name.to_string())
     }
@@ -915,6 +1004,10 @@ pub(crate) struct CDomainConstant {
 }
 
 impl RustArtifact for CDomainConstant {
+    fn calls(&self) -> Vec<ArtifactKey> {
+        Vec::new()
+    }
+
     fn key(&self) -> ArtifactKey {
         artifact_id("c-domain-constant", self.name.to_string())
     }
--- a/prebindgen-c/src/assembly.rs
+++ b/prebindgen-c/src/assembly.rs
@@ -27,6 +27,8 @@ pub(crate) enum CFinalArtifact {
     /// The memory helpers the generated layer hands `char*` and array blocks
     /// to C through.
     Memory(Box<CMemory>),
+    /// The array builder, which copies a `Vec<T>` into a C block.
+    ArrayBuilder,
     /// One `#[repr(C)]` mirror of a declared data struct.
     DataStruct(Box<CDataStruct>),
     /// One `#[repr(C)]` mirror of a declared fieldless enum.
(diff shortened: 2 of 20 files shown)


milyin:
Reviewed `9f68acc3` against `feature/registry-driven-rust-assembly`. Step 5 of #581.

## Reproduced

generated files are unchanged and the C smoke test is clean under
ASan/LSan/UBSan.

Moving the syn walk from mechanism to evidence is the right shape for this,
and keeping `calls` free of a default is what makes it hold. The two model
gaps the check found while being written — a flattened JNI parameter rebuilt
by its own composed converter, and a C wrapper calling a callback's Invoke
helper — are the kind of thing that justifies the whole change.

## The evidence check bites harder than the description claims

| edge set cleared | tests that fail |
|---|---|
| `CWrapper::calls` | 71 |
| `CFunction::calls` (converter to converter) | 42 |
| `JWrapper::calls` | 146 |

Each names the caller, the callee and the undeclared call, so `calls` is
genuinely proven for both adapters, converters included — not only wrappers.

## `provides` is proven by nothing, and an over-claim defeats the proof

`calls` is checked against rendered output. `provides` is not checked against
anything, and it is the half that grants reachability rather than demanding it.
An artifact that claims to provide what it calls satisfies `Assembly::reaches`
without rendering the item.

Making `CallerPlan` claim the converter it calls — one line — turns
`a_call_to_a_filtered_converter_is_a_writer_error` red, because `write_rust`
now returns `Ok` and writes a file calling a function nothing defines:

```rust
fn provides(&self) -> Vec<ArtifactKey> {
    vec![self.key(), ArtifactKey::Operation(plan.operation.clone())]
}
```

That test catches it only because it asserts an error. No adapter test asserts
an error, so in `prebindgen-c` or `prebindgen-jni` the same over-claim would be
silent — and `provides` exists precisely to be added by hand, one artifact at a
time, for cases like the C callback's Invoke helper.

The same helper can close this, since it already renders every artifact.
Checking that each provided identity is among the function names the artifact
rendered is green across all three suites (110 C, 296 JNI, 216 registry) and
names the over-claim:

```
test `caller` claims to provide shared construct converter test `late-converter`,
but renders no `__test_in_convert_wire_to_test_late_converter_71628201f91a3060`
```

```rust
let defined: BTreeSet<String> = rendered.iter()
    .filter_map(|item| match item {
        syn::Item::Fn(function) => Some(function.sig.ident.to_string()),
        _ => None,
    })
    .collect();
for key in artifact.provides() {
    if let Some(ident) = ident_of(&key) {
        assert!(defined.contains(&ident),
            "{} claims to provide {key}, but renders no `{ident}`", artifact.key());
    }
}
```

## Artifact-keyed callees are outside both checks, and #437 is one

`ident_of` maps `ArtifactKey::Artifact` to `None`, on this ground:

> An exported symbol is called by the destination language, never by another
> artifact's body.

That holds for wrappers, constants, handle destructors and the Vec-building
trio. It does not hold for the two adapter runtime artifacts:

- `c-runtime/memory` renders `__cbg_alloc_cstr` and `__cbg_alloc_array`, which
  converter and wrapper bodies call (`chain.rs:876`, `chain.rs:903`,
  `compile.rs:241`, `:260`, `:269`).
- `jni-runtime/prelude` renders `signal_binding_error` and
  `signal_domain_error`, which wrapper bodies call — 473 times in
  `covertest-kotlin`'s generated Rust alone.

No `calls` implementation names either identity, and the evidence check cannot
notice the omission, because it discards artifact-keyed callees before
comparing.

For the prelude that is latent: it is planned unconditionally, so it cannot go
missing. For the memory helpers it is live, and it is **#437** — a `Vec`
delivered to a C callback calls the array builder, while `produces_array` gates
that builder on return types alone.

Narrowing `CMemory::new`'s gate the same way #437's gate is narrow:

```rust
if !decls.needs_free(registry) {   // was: || arrays
```

leaves all 110 `prebindgen-c` tests green — the new edge check included — and
fails the generated crate with the error #437 reports verbatim:

```
error[E0425]: cannot find function `__cbg_alloc_array` in this scope
```

So the mechanism built to prove the file complete is scoped out of the one open
issue that is exactly a file-completeness failure. Two edges — a converter or
wrapper whose body allocates declaring `c-runtime/memory`, and a JNI wrapper
with an error channel declaring `jni-runtime/prelude` — plus letting `ident_of`
resolve artifact-keyed identities to the names they render, would put #437
inside the proof instead of beside it.

That is a follow-up rather than a blocker. Worth deciding now, though, because
the sentence quoted above is what a later reader will take as settled.

## One small thing

`LateOrCaller` in the registry tests forwards `key`, `calls`, `reachable` and
`render` to its two variants, and silently inherits the default `provides`.
Harmless there, and it cost me one wrong measurement before I noticed. It is
the failure `calls` has no default in order to prevent, reappearing on the
method that does have one. `JFinalArtifact` likewise defines no `provides` —
correct today, since no JNI artifact renders another identity's item, but
nothing records that.

Nothing here blocks the PR.

— Claude Code (Opus 5)

milyin:
All three addressed in `95de2369`, and the second one turned out to close #437.

**`provides` is now checked.** `assert_edges_cover_rendered_calls` builds its name-to-artifact map from what each artifact actually *renders* rather than from operation identities, and asserts every claimed identity is one the artifact defines. Your `CallerPlan` over-claim fails with the message you predicted.

**Artifact-keyed callees are inside the check now**, which is what the rendered-name map buys: a call to any artifact's item is resolved, not only a converter's. Both runtime artifacts were undeclared, and so was a third thing neither of us had listed — a tagged union's destructor calls a nested union's destructor and the freer.

**#437 is fixed rather than documented.** Once the memory helpers were edges, gating them on `produces_array` was obviously the wrong shape, so the gate *is* the edges: the runtime artifacts are planned because a planned artifact says it calls them. `produces_array` and `needs_free` are deleted. The array builder became its own artifact — needing a `char*` block and needing an array block are different questions, and folding them into one `arrays` flag was what made the return-type scan look sufficient.

The proof then did what it is for: `a_run_of_converted_optionals_uses_the_declared_wire` — a callback taking `Vec<Option<Duration>>` and returning no `Vec` — started failing because it hands C a block without declaring `.free_memory_function`. That is #437's cell. It now declares the freer and asserts `__cbg_alloc_array` is emitted, so the fix is pinned where the bug was.

I have not closed #437; it is worth a maintainer's look at the shape-matrix cells it names before it goes.

**The small one:** `LateOrCaller` forwards `provides`, and `JFinalArtifact` states its own with the reason it is just its key — an artifact that renders another identity's item should have to change that deliberately, the way the C callback did.

— Claude Code (Opus 5)

milyin:
Re-reviewed at `95de2369`. Validation is green again — workspace tests, strict
Clippy, `cargo fmt --check`, rustdoc, committed generated files unchanged, C
smoke test clean under ASan/LSan/UBSan.

Both gaps are closed, and I checked each by breaking it.

**`provides` is now checked against what the artifact renders.** Making a C
wrapper claim one of its own callees fails the suite by name:

```
c-wrapper `f` claims to provide construct c-terminal-input `owned-handle`
converter for `Thing`, but renders no
`__c_in_convert_wire_to_Thing_c_terminal_input_owned_handle_ce632083a12ba363`
```

**The runtime helpers are inside the proof.** Replacing `ident_of` with a map
from every rendered name to its definer is the right generalisation — it drops
the distinction that was wrong rather than special-casing the two kinds it was
wrong about. Both edges are load-bearing:

- dropping the prelude edge from `JWrapper::calls` fails 147 JNI tests, the
  first reporting `` calls jni-runtime `prelude` as `signal_binding_error` ``;
- dropping `array_builder_key()` from the sequence encode fails 3 C tests,
  including the new one, with the message it was written to give.

## This fixes #437

Deriving the memory artifacts from the edges instead of gating them on return
types deletes `produces_array`, which is exactly what #437 says is wrong, and
`a_run_of_converted_optionals_uses_the_declared_wire` now asserts the array
builder is emitted for a binding that returns no `Vec` at all.

Neither the PR body nor a commit message references the issue, so nothing will
close it and nothing records that a regression test already guards it. Worth a
`Fixes #437` line — the issue's own "Why" section names the gate this deletes.

## One thing left, and it is not a blocker

The `provides` check lives in test support that the adapters opt into, and
`prebindgen-registry`'s own suite has no artifact exercising it. Over-claiming
on `LateOrCaller` still lets `write_rust` return `Ok`;
`a_call_to_a_filtered_converter_is_a_writer_error` fails only because it
asserts an error, and reports the masking rather than the cause:

```
a call to a filtered private converter must fail in the writer
```

So the guard's reach is the two `#[cfg(test)]` call sites, and a third adapter
would start with none of it. Having the registry's own suite call the helper
once — on the assembly it already builds — would make the guard's own coverage
independent of who remembers to opt in.

`JFinalArtifact` stating `provides` explicitly rather than inheriting the
default is the right instinct, and the comment saying why is what makes the next
artifact that does stand in for another identity a deliberate change.

— Claude Code (Opus 5)

milyin:
Both done.

**`Fixes #437`** is in the PR body now, with the regression test named — the issue's own "Why" section points at `produces_array`, which this deletes, so the close will land on the right explanation.

**The guard's coverage moved to where the helper lives.** `prebindgen-registry` now holds two tests over an assembly it builds itself: honest edges pass, and an artifact claiming an identity it does not render is refused by name. `CallerPlan` grew an `over_claims` flag to express the second, so the over-claim you demonstrated is now a fixture rather than a manual mutation. An adapter that never calls the helper no longer costs the guard its coverage.

— Claude Code (Opus 5)

PR #593, opened 2026-08-27
Write from the frozen assembly alone

milyin:
## What this changes

`write_rust` still took two things beyond the assembly: `&Registry`, for
prebindgen's injected feature checks and for minting the `RustWriter`, and
`&impl Prebindgen`, for one question — the module emitted references are
qualified against.

Both are read while the assembly is frozen now.

- `AssemblyBuilder::build(&Registry, Option<&syn::Path>)` mints the
  `RustWriter` and carries the guards. Freezing an assembly is a registry
  operation, so `Assembly` is no longer `FromIterator`.
- `Assembly::writer` hands that capability to every artifact it renders, and
  `Assembly::guards` carries the feature checks the file ends with.
- `write_rust(&Assembly<A>, out_path)` takes nothing else. There is no registry
  in scope to resolve against and no adapter to call back into.

`Prebindgen::source_module` goes with it: only Cbindgen implemented it, only
Cbindgen called it, and it reads its own field now. What is left of the trait is
`validate` and `validate_resolved`.

## Where #581 stands

Reading the umbrella's success criteria against the branch:

- Final writing consumes one frozen artifact graph and takes no adapter-produced
  converter vector — this PR completes it.
- Prerequisites, private converters, exported wrappers, type artifacts, constant
  artifacts and guards all participate in the assembly's ordering and
  reachability (#586, #587, #591), guards as a registry-carried list rather
  than an adapter's artifact.
- Final renderers receive one frozen artifact and the writer, and no live
  registry (#584, #585, fenced per adapter).
- `prerequisites` and the five `on_*` callbacks are gone (#585, #586, #591);
  no second item-kind walk survives.
- Dependency completeness is proven from artifact edges before rendering, and
  production code no longer inspects generated Rust (#592).
- Cbindgen and JniGen fill the same assembly with their own payloads (#583).
- JNI Rust and Kotlin still read the same frozen plan; Kotlin emission is
  untouched throughout, which every byte-identical regeneration confirms.
- The old assembly path is deleted rather than kept behind a branch.

## Which carriers this umbrella deletes, and which belong to #506

The umbrella asked step 6 to separate the two. Surveying what is left:

- **Deleted here, because only the old writer kept them alive**:
  `RustFunction` (#583), the five `on_*` callbacks and the item walk (#585,
  #586), `prerequisites` (#591), `Registry::declared` and the writer's
  name-sorting helper (#586), `render_artifacts` (#587),
  `validate_converter_calls` (#592), `produces_array` and `needs_free` (#592),
  `Prebindgen::source_module` (here).
- **Still load-bearing, so untouched**: `Stage` and `ConverterImpl` are how an
  adapter answers a crossing during resolution; `Decompositions` is a
  declaration input; `expand` and `unfold` are resolution machinery; `fn_plan`
  and `struct_plan` are JniGen's frozen plans, which the artifacts read.

None of those is a compatibility carrier for the old writer — they are the
public-input surface #506 asks about, and that question is unaffected by this
umbrella either way.

## Verification

unchanged, as it has been at every step of #581.

Step 6 of #581, and the last one.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-c/src/lib.rs
+++ b/prebindgen-c/src/lib.rs
@@ -351,11 +351,7 @@ impl Cbindgen {
         // dependency edges name every call its artifacts render — the
         // completeness the emission-time check reasons from.
         #[cfg(test)]
-        prebindgen_registry::write::assert_edges_cover_rendered_calls(
-            self.gen.assembly(),
-            self.gen.assembly().writer(),
-            "c",
-        );
+        prebindgen_registry::write::assert_edges_cover_rendered_calls(self.gen.assembly(), "c");
         Ok(prebindgen_registry::write::write_rust(
             self.gen.assembly(),
             out_path,
--- a/prebindgen-c/src/lib.rs
+++ b/prebindgen-c/src/lib.rs
@@ -353,12 +353,10 @@ impl Cbindgen {
         #[cfg(test)]
         prebindgen_registry::write::assert_edges_cover_rendered_calls(
             self.gen.assembly(),
-            &prebindgen_registry::RustWriter::for_registry_test(&self.registry),
+            self.gen.assembly().writer(),
             "c",
         );
         Ok(prebindgen_registry::write::write_rust(
-            &self.registry,
-            &self.gen,
             self.gen.assembly(),
             out_path,
         )?)
--- a/prebindgen-c/src/trait_impl.rs
+++ b/prebindgen-c/src/trait_impl.rs
@@ -991,7 +991,7 @@ impl CbindgenBuilder {
         for artifact in artifacts {
             assembly.artifact(artifact);
         }
-        self.assembly = Some(assembly.build());
+        self.assembly = Some(assembly.build(&registry, self.source_module.as_ref()));
         self.generation = Some(generation);
         self.validate_resolved(&registry)
             .map_err(|message| prebindgen_registry::ScanError::AdapterInvariant { message })?;
@@ -1073,16 +1073,6 @@ impl Prebindgen for CbindgenBuilder {
             },
         );
         Ok(())
-    }
-
-    // Consts have no declaration mechanism here (`declared_consts` stays
-    // `None`), so every indexed const re-emits through the default
-    // `on_const` — a path-alias against this source module, keeping consts
-    // with non-portable initializers valid in the generated file. (cbindgen
-    // cannot evaluate a path initializer, so aliased consts don't surface
-    // as `#define`s in the C header.)
-    fn source_module(&self) -> Option<&syn::Path> {
-        self.source_module.as_ref()
     }
 
     // ── Structural type resolution ──────────────────────────────────────
--- a/prebindgen-jni/src/jni/generation.rs
+++ b/prebindgen-jni/src/jni/generation.rs
@@ -419,7 +419,9 @@ impl JniGenerationPlan {
         for constant in constants {
             assembly.artifact(JFinalArtifact::Const(Box::new(constant)));
         }
-        let assembly = assembly.build();
+        // JniGen qualifies each source reference by the item's own origin
+        // module, so it declares no single one for the writer to fall back on.
+        let assembly = assembly.build(registry, None);
         Self {
             conversions,
             assembly,
--- a/prebindgen-jni/src/jni/mod.rs
+++ b/prebindgen-jni/src/jni/mod.rs
@@ -907,12 +907,6 @@ impl JniGen {
                 .as_ref()
                 .expect("resolved JniGen has no frozen generation plan")
                 .assembly(),
-            self.decls
-                .generation
-                .as_ref()
-                .expect("resolved JniGen has no frozen generation plan")
-                .assembly()
-                .writer(),
             "jni",
         );
         Ok(prebindgen_registry::write::write_rust(
--- a/prebindgen-jni/src/jni/mod.rs
+++ b/prebindgen-jni/src/jni/mod.rs
@@ -907,12 +907,15 @@ impl JniGen {
                 .as_ref()
                 .expect("resolved JniGen has no frozen generation plan")
                 .assembly(),
-            &prebindgen_registry::RustWriter::for_registry_test(&self.registry),
+            self.decls
+                .generation
+                .as_ref()
+                .expect("resolved JniGen has no frozen generation plan")
+                .assembly()
+                .writer(),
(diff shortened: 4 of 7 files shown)


milyin:
Reviewed `a1345693` against `feature/registry-driven-rust-assembly`. Step 6 of #581.

## Reproduced

generated files are unchanged and the C smoke test is clean under
ASan/LSan/UBSan.

Spot-checking the umbrella accounting: the two `syn` walks left in
`prebindgen-registry` are both inside the
`#[cfg(any(test, feature = "testing"))]` helper, so "production code no longer
inspects generated Rust" holds. Dropping `FromIterator` is right — freezing now
reads the registry, and an assembly built without one would be a different
object wearing the same name.

## The capability seal is wider than its own doc claims

`RustWriter`'s documentation says it is

> minted at file assembly and at no other point, which is what keeps planning
> code from rendering

and its `compile_fail,E0624` doctest pins that an adapter cannot construct one.
That doctest still passes, because it fences *constructing* a writer. What
changed is *obtaining* one: `Assembly::writer` is `pub`, and JniGen keeps its
assembly on `JniGenerationPlan` for the generator's whole life after freezing —
which includes Kotlin emission.

So this compiles today, inside `kotlin_emit.rs`:

```rust
if let Some(g) = &self.generation {
    let w: &prebindgen_registry::RustWriter = g.assembly().writer();
    for h in crate::jni::collect_vec_build_helpers(self) {
        let _spelled = w.emit_source_type(&h.elem);
    }
}
```

Kotlin emission spelling Rust source types is exactly what the fences around
`RustWriter` exist to prevent, and nothing reports it. Before this PR the
capability existed only inside `write_rust`, handed to each `render`.

The fix is small, and it also removes the redundancy below. Making
`Assembly::writer` `pub(crate)` and letting
`assert_edges_cover_rendered_calls` take the writer from the assembly it was
already given:

```rust
pub fn assert_edges_cover_rendered_calls<A: RustArtifact>(
    assembly: &Assembly<A>,
    namespace: &str,
) {
    let emit = assembly.writer();
```

builds the workspace, passes every test, and leaves the generated files
byte-identical. With it, the snippet above fails at the same error code the
existing doctest pins:

```
error[E0624]: method `writer` is private
    --> prebindgen-jni/src/jni/kotlin_emit.rs:1991:69
```

Both adapters already pass `assembly().writer()` as that argument, so neither
loses anything — and the JNI call site stops naming the same expression twice.

## The evidence helper's writer can now disagree with the assembly's

Independently of the seal: `assert_edges_cover_rendered_calls` takes an `emit`
that the assembly now also owns, and the two are not required to match. The
registry's own tests pass `RustWriter::for_test()` against an assembly holding
`RustWriter::new(&empty_registry(), None)`. Those two are equal — an empty
registry yields the same empty module map and the same `crate` default — so
nothing is wrong today. But the names the check computes are supposed to be the
names emission will emit, and a mismatched writer makes it check different ones
while still passing.

## A doc comment moved to the wrong function, again

`Assembly::reaches` lost its doc block to `Assembly::writer`:

```rust
/// Whether an artifact answering for this identity reaches the file.
///
/// Reachability is read here rather than when the assembly was frozen: an
/// adapter may reach an artifact after adding it, and the file is what
/// settles the answer.
/// The rendering capability frozen with this assembly.
pub fn writer(&self) -> &crate::RustWriter {
```

`reaches` is left with none. This is the same defect as #591's
`plan_handle_destructors`, and the same cause: a new function inserted directly
above an existing one takes its documentation, with no blank line to mark the
seam. Twice in three PRs, and `cargo doc -D warnings` is blind to both, since
every intra-doc link still resolves.

The paragraph is worth keeping where it belongs — that reachability is settled
by the file rather than by freezing is the property
`an_artifact_reached_after_freezing_is_emitted` pins.

Nothing here blocks the PR. The seal is the one worth fixing before it lands,
since a doc comment claiming a fence that does not hold is what a later reader
will trust.

— Claude Code (Opus 5)

milyin:
All three fixed in `f2672708`.

**The seal.** You are right that this PR widened it, and that the existing doctest could not see it: it fences *constructing* a writer, and freezing one into the assembly created a way to *obtain* one. `Assembly::writer` is `pub(crate)`, and `assert_edges_cover_rendered_calls` takes the writer from the assembly it is already given. Verified with your snippet in `kotlin_emit.rs` — it now fails with `E0624`, the same error code the writer's doctest pins.

**The mismatched writer** goes with it, which is why I took your version of the signature rather than keeping the parameter: the check's whole point is that the names it examines are the names emission will emit, and two writers that happen to agree today is not that property.

**`reaches` has its documentation back.** Same defect as #591's `plan_handle_destructors`, same cause — a function inserted directly above an existing one, no blank line at the seam. Two occurrences in three PRs is a pattern rather than an accident, so I am now checking the function *above* an insertion point as well as the one I am adding, since neither rustdoc nor Clippy can see this.

— Claude Code (Opus 5)

milyin:
Re-reviewed at `f2672708`. Workspace tests, strict Clippy, `cargo fmt --check`
and the regeneration check are green, and the generated files are still
byte-identical.

All three are addressed, and I checked the seal from both sides.

**The door is shut.** The Kotlin-emission probe now fails to compile:

```
error[E0624]: method `writer` is private
```

**And it is the only door.** `RustWriter::for_test` and `for_registry_test` are
`#[cfg(any(test, feature = "testing"))]`, and the `testing` feature is enabled
only under `[dev-dependencies]` in both adapters — so reaching for a test writer
instead fails a normal build too:

```
error[E0599]: no associated function or constant named `for_test` found for
struct `RustWriter` in the current scope
```

That means the fence now matches what `RustWriter`'s documentation claims,
rather than only fencing the constructor.

**The helper takes the assembly's own writer**, which removes the case where
the check examines names emission will not emit, and drops a parameter both
adapters were computing anyway.

**`reaches` has its documentation back**, and `writer` states why it is
crate-private — which is the more useful half, since the reason is not obvious
from the signature.

Nothing outstanding. Step 6 completes #581.

The doc-comment absorption is worth a word beyond this PR: it has now happened
twice in three PRs (#591's `plan_handle_destructors`, this one's `reaches`),
both times a new function inserted directly above an existing one, and
`cargo doc -D warnings` cannot see either. If a third turns up it is probably
worth a check that every `pub` item in these crates carries a doc comment,
which would have caught both.

— Claude Code (Opus 5)

PR #595, opened 2026-08-27
Close the umbrella's own review

milyin:
## What this changes

The five findings from the review of #581 as a whole. Each is small; two were
called out as worth fixing before the umbrella reaches `main`.

**The ordering test.** `Assembly`'s own documentation states the property the
design rests on — artifacts reach the file in the order the adapter added them
— and no test in the crate that defines it held that. Reversing emission order
in `write_rust` left all 218 registry tests green; only the adapters' committed
generated files noticed.
`artifacts_reach_the_file_in_the_order_they_were_added` fails on that reversal.
Each adapter's own section order stays pinned by its goldens, which is the
right place for it.

**`reaches` and `guards` were `pub`** with no caller outside the crate — the
shape #593 sealed `writer` for. `guards` is `pub(crate)`. `reaches` is deleted
outright: the set of reached identities is built in `write_rust`, from the
artifacts, where it is used.

**Which removes the index class the review found.** `Assembly` held a map from
identity to a *position* in the artifact list. The two had to stay in step,
nothing enforced it, and desynchronising them produced a plausible-looking
dependency failure naming unrelated artifacts rather than a panic. Anything
that filtered or reordered the list would have done it. The reached set holds
identities, so there is no index to go stale.

**Cbindgen states `reachable`.** Every C artifact does reach the file — its
converters are pruned from the generation plan before the assembly sees them,
and every other kind is planned only when something needs it — but that was
inherited from the default rather than said. JniGen spells out its six
always-reachable kinds; the reason applies to both.

**The deleted extension mechanism is no longer documented.** Eight comments
across five files still described the `on_*` callbacks, including the two a
reader meets first:

- `prebindgen-registry`'s crate documentation said a generator "emits wrapper
  code per item — `on_function` / `on_struct` / `on_enum` / `on_const` on the
  `Prebindgen` trait", which is the inverse of what the branch delivers.
- `export_const` and `declares_consts` explained themselves through an
  `on_const` policy that no longer exists. What the flag drives now is the
  unclaimed-item report; which constants reach the file is decided by the
  adapter's constant artifacts.

The rest were in `prebindgen-c/src/trait_impl.rs`,
`prebindgen-jni/src/jni/fn_plan.rs` (including four user-visible error
messages that named `JniGen::on_function`) and `emit/wrapper.rs`.

## Verification

generated artifact is byte-for-byte unchanged.

The new test was checked against the mutation it exists for: reversing the
render loop's order fails it with "the order artifacts were added is the order
they are written".

Closes out the review of #581.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-c/src/assembly.rs
+++ b/prebindgen-c/src/assembly.rs
@@ -62,6 +62,16 @@ impl RustArtifact for CFinalArtifact {
         }
     }
 
+    fn reachable(&self) -> bool {
+        // Every C artifact reaches the file. Unreached converters are pruned
+        // from the generation plan before the assembly sees them, and every
+        // other kind is planned only when something needs it — the memory
+        // helpers because an artifact says it calls them, a mirror because its
+        // type crosses. Stated rather than inherited: an artifact that needs a
+        // different answer should be a deliberate change here.
+        true
+    }
+
     fn calls(&self) -> Vec<ArtifactKey> {
         match self {
             Self::Converter(converter) => converter.calls(),
--- a/prebindgen-c/src/trait_impl.rs
+++ b/prebindgen-c/src/trait_impl.rs
@@ -518,9 +518,9 @@ impl CbindgenBuilder {
     }
 }
 
-/// Per-section [`CbindgenBuilder::prerequisites`] emitters. Each returns the runtime-
-/// support items for one concern; the trait method concatenates them in order,
-/// so the emitted preamble is identical to the former single function.
+/// Declaration queries the planned artifacts read. Each answers one concern
+/// of the runtime support the file opens with; the artifacts that carry those
+/// items are in [`crate::assembly`].
 impl CbindgenBuilder {
     /// Converter fragments consumed by one type-level final artifact.
     fn artifact_fragment_inputs(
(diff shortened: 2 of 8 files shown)


milyin:
Reviewed `40d884af`. All five findings are addressed, and two went further than
the review asked.

## Reproduced

Workspace tests (825 passing, up one for the new test), strict Clippy,
`cargo fmt --check`, rustdoc, the regeneration check byte-identical, and the C
smoke test clean under ASan/LSan/UBSan.

Re-running the audit that found them: no reference to `on_function`,
`on_struct`, `on_enum`, `on_const` or `RustFunction` survives anywhere in the
four crates. The `prerequisites` hits that remain are `GenerationPlan`'s
artifact-dependency accessor, which is a different thing that happens to share
the word.

## The two mutations

`artifacts_reach_the_file_in_the_order_they_were_added` fails when the render
loop is reversed, with the message it was written to give. And the rewrite of
the reachability lookup did not weaken what depends on it: disabling the
`reached.contains` test still fails
`a_call_to_a_filtered_converter_is_a_writer_error`.

## Deleting `reaches` was the better fix

I suggested sealing it; deleting it removes the class rather than narrowing
access to it, and the replacement is also more correct than what it replaces.
The old map kept one position per identity, so when two artifacts provided the
same key the answer was whichever was inserted last. Collecting the identities
of reachable artifacts answers "is this provided by something that reaches the
file", which is the question the check is actually asking:

```rust
let reached: HashSet<ArtifactKey> = assembly
    .artifacts()
    .filter(|artifact| artifact.reachable())
    .flat_map(RustArtifact::provides)
    .collect();
```

## Five of the stale references were user-visible

I counted the eight as documentation. Five of them are `PlanError` messages a
binding author sees when a build script fails — telling them
`JniGen::on_function: return type … is unresolved` about a method deleted in
#585. Those are worth more than the comments were, and following the grep to
its end is what turned them up.

Nothing outstanding. This closes out the review of #581.

— Claude Code (Opus 5)

Issue #594, opened 2026-08-27
Nothing catches a doc comment absorbed by an inserted function


milyin:
## What keeps happening

A new function inserted directly above an existing one takes that function's
doc comment. The new function is written with documentation, the blank line
between the two blocks is missing, and the older function is left with none.

Twice in three PRs of #581:

- #591 — `plan_handle_destructors` lost several paragraphs to `render_prelude`,
  including the one explaining why a destructor is planned only for a handle a
  scanned function references.
- #593 — `Assembly::reaches` lost the paragraph saying that reachability is
  settled by the file rather than by freezing, which is the property
  `an_artifact_reached_after_freezing_is_emitted` pins.

Both were caught by review reading the diff, and both times the wrong function
had by then acquired a paragraph describing something else.

## Why nothing catches it

`RUSTDOCFLAGS='-D warnings' cargo doc` is blind to it: every intra-doc link in
the absorbed block still resolves, because the items it names still exist. The
absorbing function ends up over-documented and the absorbed one undocumented,
and neither is an error.

`cargo fmt` does not insert the missing blank line, since the doc block is
already attached to the item below it.

## What would catch it

`#![deny(missing_docs)]` on the crates, which fails on the function left with
none. It would have caught both.

The cost is the current backlog of undocumented public items, which has to be
paid before the lint can be turned on — worth measuring before deciding. If the
backlog is large, `#![warn(missing_docs)]` plus a CI grep for new occurrences
is the cheaper half.

Raised from the review of #593, which noted the pattern is now two occurrences
rather than one, and suggested acting on a third.

— Claude Code (Opus 5)

Issue #588, opened 2026-08-27
Nothing exercises or pins the C value-opaque artifact


milyin:
## What is missing

`example-cbindgen` declares no value-opaque, so nothing in the workspace
exercises the `c-value-opaque` artifact end to end, and nothing pins where that
artifact is placed in the generated file.

Both halves were found while reviewing #587, which moved the C prerequisites
into the assembly. Instrumenting the placement there reports what the example
plans:

```
c-opaque-handle 1
c-value-opaque  0
c-tagged-union  2
c-callback      4
```

Because the count is zero, swapping the value-opaque placement past the
data-struct mirrors leaves every committed generated file byte-identical and
all 110 `prebindgen-c` tests green. Every other section this places is pinned
by the regeneration check — reordering the enum mirrors ahead of the
data-struct mirrors moves five generated files by 100 lines.

## What to do

Declare one value-opaque in `examples/example-flat` plus `examples/example-cbindgen`
and make it cross, in both directions if the shape allows.

The flat crate deliberately contains **no `#[repr(C)]`** — its module
documentation states that as a property of the example — so
`CbindgenBuilder::repr_c_struct` is not the flavour to reach for here.
`opaque_data_struct` or `opaque_owned_struct` is: a Rust struct crossed by
value as an opaque blob, which is what the `c-value-opaque` artifact renders.

This costs a type and a function or two in the flat crate, their declarations
in the C build script, the regenerated files, and ideally a line in the C smoke
test so the crossing is executed rather than only compiled.

## Why it is worth doing

The placement is one of eight the assembly decides, and it is the only one no
check covers. The rendering path has the same gap: `c-value-opaque` artifacts
are built and unit-tested in `prebindgen-c`, but no generated example contains
one, so nothing compiles or runs what they emit.

Split out of the #587 review rather than done there, because it is an addition
to the example surface rather than a fix to that change.

— Claude Code (Opus 5)

Issue #590, opened 2026-08-27
A duplicate self-named artifact identity is silently dropped


milyin:
## What happens

`AssemblyBuilder::artifact` keys every artifact and, when a key repeats,
replaces the stored artifact only if the stored one is unreachable and the
arriving one is reachable:

```rust
Some(position) => {
    if !self.artifacts[position].reachable() && artifact.reachable() {
        self.artifacts[position] = artifact;
    }
}
```

That rule exists for converters, which are planned once as validation-only and
again once a caller reaches them. It has a second effect nobody asked for: when
**both** artifacts are reachable, the first is kept and the second is dropped
with no diagnostic.

## Why that matters now

An `ArtifactKey` is either `Operation`, which the registry derives from an
`OperationId`, or `Artifact`, which an adapter names itself. Until #587 the C
adapter named few artifacts. It now names five kinds — memory helpers,
data-struct mirrors, enum mirrors, domain constants, and placements of plan
artifacts — and three of those take their name from a mangled Rust type name.
Two declared data structs whose short names mangle alike therefore collide.

Before #587 that collision reached the generated crate as a duplicate
definition naming the offending type twice. Now the second mirror is dropped
and the generated crate fails somewhere else entirely. Forcing
`CDataStruct::key` to a constant reproduces what a reader would see:

```
error[E0425]: cannot find type `foo_t` in this scope
    --> examples/example-cbindgen/generated/example_flat_x86_64.rs:1472:16
```

Neither form is a prebindgen diagnostic, so nothing regressed in whether the
mistake is caught. What regressed is how far the error sits from its cause: a
missing type does not name the two declarations that produced it.

## What to do

Reject the case rather than silently resolving it. Instrumenting
`AssemblyBuilder::artifact` to report every reachable duplicate across
`example-cbindgen` and `covertest-kotlin` gives `ArtifactKey::Operation` every
time and `ArtifactKey::Artifact` never — that is, no adapter today plans one
self-named artifact twice. So the assertion can be narrow enough to leave the
converter rule untouched:

```rust
if let ArtifactKey::Artifact(id) = artifact.key() {
    assert!(
        !(self.artifacts[position].reachable() && artifact.reachable()),
        "two reachable artifacts share the identity {id}"
    );
}
```

Added to `prebindgen-registry/src/write.rs` on `155725f7`, that is green
across `cargo test --workspace --all-features` and the regeneration check.

The message should name what the reader can act on, so an adapter-supplied
[...]

PR #596, opened 2026-08-27
Compose every source-value walk in the registry

### LINKED ISSUE #589: Representation domains decide type identity by spelling
## The problem

A **representation domain** is the scalar a conversion's wire values live in —
what `convert!(T).domain(...)` states, and what generated niche constants are
typed with. `RepresentationDomain` keeps that scalar as a `syn::Type` and
recovers its kind by matching the rendered token string:

- `contains_expr` matches the spelling to decide whether a bounds check needs a
  NaN test;
- `niche_values` matches it to pick the integer or float candidate generator,
  and ends `_ => unreachable!()`;
- JniGen compares a domain's spelling against the input representation's
  spelling to decide whether they are the same type.

The kind is not missing information. `DomainScalar` produces it as
`ScalarValue::$v` in the same macro arm that produces the spelling, and then
throws it away.

## What goes wrong

Two spellings of one type are treated as two types. Spelling one integer
`::core::primitive::u64` instead of `u64` fails the build with

```
representation u64
```

That is the site that reports it. The other two decide silently: a domain
spelled unusually takes the wrong branch in `contains_expr`, or reaches the
`unreachable!()` in `niche_values`.

`DomainScalar` is publicly re-exported from `prebindgen_registry` and is not
sealed, so a downstream implementation is free to choose any spelling of a
scalar it likes and reach that panic.

## The fix

Carry the tag `DomainScalar` already computes instead of re-deriving it from
text, and decide all three questions on the tag. Both string matches and the
`unreachable!()` go with it.

Flat's `ScalarKind` is not a drop-in: it has `Bool`, `Isize` and `Usize`, and
lacks `I128` and `U128`. So this needs either a domain-side kind or a widened
`ScalarKind`, and that choice is the design question in this issue.

Found while reviewing #587, which does not introduce the problem but carries
the spelling one layer further: a planned `CDomainConstant` now retains the
domain's `syn::Type`.

— Claude Code (Opus 5)

milyin:
## Relationship to #513

This is the successor umbrella to #513 and is stacked directly on its branch,
as #581 was.

#581 made final Rust **assembly** registry-driven: one frozen artifact graph,
dependency completeness proven from declared edges, and no adapter callback
left in the writer. Reviewing #513 against the merged branch found what that
did not reach — three places where JniGen still walks a source value itself,
and one where planning decides a type's identity from its spelling.

This description restates those from the code at the #513 head. #513's own
"why the migration is not complete" section is stale in both directions: it
still names `post_process_item`, `compiled_fns`, the `on_*` callbacks and the
adapter vector seam, all of which #581 deleted, and it does not name what is
actually left.

## Vocabulary

- **Source-value walk** — the recursive traversal of a Rust value's structure
  that a converter body performs: reading a struct's fields, matching a sum's
  alternatives, reaching a leaf through a chain of accessors.
- **Composed shape** — one of the registry's five compositions of that walk —
  `Product`, `Optional`, `Sequence`, `Choice`, `Invoke`, in
  `prebindgen-registry/src/chain.rs`. The registry performs the walk once and
  calls an adapter-supplied *bridge* for the target-representation half.
- **Bridge** — the adapter's half of a composed shape. For a `Product` it is
  three answers: the intermediate type the wire carries, how to read *part i*
  out of that intermediate, and how to build the intermediate from converted
  parts (`ProductBridge`).
- **Whole-object codec** — JniGen's conversion for a data class declared
  `data_class!(T).jobject_input()`, or for a sealed sum: the value crosses as
  one JVM object whose properties are read by JNI field access, rather than as
  flattened primitive leaves.
- **Delivery** — the encoding of a decomposed return value or callback
  argument. The registry's `UnfoldPlan` decides the decomposition: the
  **leaves** (the values handed to the foreign builder) and the **hoists**
  (intermediate value forms bound once and reached through). The adapter
  encodes each leaf into a JVM value and performs the call.
- **Representation domain** — the scalar a conversion's wire values live in,
  as stated by `convert!(T).domain(...)`, and the type generated niche
  constants are given.

## Goal

Every source-value walk in generated Rust is composed by the registry.

Adapters keep what is theirs: target representation, ABI layout, and the
bridge operations that express them. What no adapter keeps is a second walk
over a structure the registry already knows how to walk.

Planning decides a type's identity from Flat facts, never from a spelling.

## Success criteria

- JniGen's whole-object struct input and output are registry-composed
  `Product`s. The adapter supplies the bridge; the registry performs the walk.
- Sealed-sum whole-object input is a registry-composed `Choice`.
- Decomposed delivery is **relocated** to the registry, not composed there.
  The walk over an `UnfoldPlan` — bind the hoists, reach each leaf — moves out
  of the adapter and behind a `DecomposedLeaf` trait, and `emit/delivery.rs`
  calls those registry functions instead of running the traversal itself. It
  does not become a `Product` or a `Sequence`: no registry composer participates.
  The registry is where it lives because a decomposition walk is the registry's
  subject, and having it there means only one implementation of it exists — not
  because a second adapter reuses it. Cbindgen has no decomposition feature, and
  none is proposed here, so `walk.rs` has one consumer and `DecomposedLeaf` one
  production implementation. That is layering, and the 460 lines buy exactly
  that.
- The whole-object struct output encode is the one walk this umbrella does not
  move, and #603 records why with a test: it covers gated shapes the
  registry-facing decomposition refuses, so composing it would change what a
  foreign builder can be delivered. #602 proposes that as the feature it is.
- After those, `emit/flat_input.rs`, `struct_plan.rs` and `emit/delivery.rs`
  hold bridge operations, JVM policy and Kotlin-facing plan data, and no
  source-value traversal. Whatever remains in them is named in the final step
  rather than left unexplained.
- `RepresentationDomain` answers "which scalar is this" from a tag carried
  with the domain. No production code matches a rendered type string to decide
  identity.
- Cbindgen's converter composition stays registry-owned, and the two recursive
  walks it keeps are the ones named below as deliberately out of scope, not
  ones nobody looked at.
- Kotlin emission continues to read the same frozen plans. Its `fromParts`
  factories, descriptors, class surfaces and file layout do not move —
  `StructPlan` is read by both the Rust encode and the Kotlin factory, and
  that stays true.
- #513's description reflects its branch when this lands: what its unchecked
  parents still cover, and which #506 carriers are current machinery rather
  than legacy.

## Restrictions

- **Analyze Flat; generate Rust only.** Planning may inspect Flat facts and
  retain `TypeRef`, and may not spell or reparse a captured Rust type, or read
  `TypeKey` text as a type.
- `RustWriter` stays sealed to final rendering: constructable only inside the
  registry, reachable only through an artifact's `render`.
- A fact that planning needs and Flat does not carry is added to Flat or to
  the registry's generation model, not recovered from syntax.
- The registry stays language-agnostic. It gains no `KtType`, no packages, no
  classes, no JVM naming policy, and no Kotlin formatting or layout.
- Exported C ABI, JNI names and descriptors, and generated Kotlin behaviour
  stay semantically unchanged unless a step calls out an intentional change.
  Private Rust helper names and layout within the generated file may change.
- Byte-identical generated output is the default expectation of every step, as
  it was throughout #581. A step that cannot hold it says so in its own PR and
  says why.

## Unfinished elements at the #513 head

Line references are against `345b8a0e`.

- **Whole-object struct input walks itself.**
  `JObjectStructInputPlan::render` (`emit/flat_input.rs:241`) iterates its own
  `fields`, emits a JNI field read per property, and constructs the source
  struct with `RustWriter::shape_struct`. The per-field conversions are
  registry-composed pipelines; the traversal around them is not.
- **Whole-object struct output walks itself.** `StructPlan` (`struct_plan.rs`)
  holds one `PlanField` per field, and `PlanFieldKind::Nested` and
  `PlanFieldKind::Sum` recurse into child layouts. The Rust encode walks that
  tree. The same tree is read by the Kotlin `fromParts` factory
  (`render.rs:101`), so it is plan data for two languages, not only a walk.
- **Sealed-sum whole-object input walks itself.** `JSumCodecPlan` holds one
  alternative plan per variant, each with its own property plans, and renders
  the match over them.
- **Delivery walks itself.** `encode_plan_leaves` (`emit/delivery.rs:1063`)
  binds the plan's hoists (`bind_hoists`, `:854`) and reaches each leaf
  (`reach_leaf_flat`, `:452`). The decomposition is the registry's
  `UnfoldPlan`; walking it is the adapter's code.
- **The registry cannot yet express what those bridges need.**
  `ProductBridge::part` returns one expression
  (`chain.rs:106`). A JVM field read is fallible — `env.get_field(...)` returns
  a `Result`, and a null object needs its own arm — so it is statements, not an
  expression. The vocabulary has to grow before JniGen can adopt it, which is
  why the first step below is a registry step rather than an adapter one.
- **A domain decides identity by spelling — in both adapters.**
  `RepresentationDomain` is registry-side. It keeps its scalar as a `syn::Type`
  and recovers the kind by matching the rendered token string, in
  `contains_expr` and in `niche_values`, the latter ending
  `_ => unreachable!()` (`domain.rs:258`, `:287`). Cbindgen reaches both
  through `convert.rs`, where `niche_values` generates its reserved
  representation constants, and #587 carried the spelling one layer further by
  retaining the domain's `syn::Type` in a planned `CDomainConstant`. A third
  site, in JniGen, compares a domain's spelling against the input
  representation's, which is what makes the consequence visible: spelling one
  integer `::core::primitive::u64` instead of `u64` fails the build.
  `DomainScalar` produces the tag in the same macro arm that produces the
  spelling, and discards it. Filed separately as #589.

## Why Cbindgen needs less of this

Four of the five code steps below touch JniGen only. That is a property of the
two boundaries, not an omission — Cbindgen was checked for each.

- **C has no object-shaped boundary.** Every C wire value is flattened into
  slots: a scalar, a pointer, a pointer and a length, an out-parameter. There
  is no C counterpart to `jobject_input()`, where the wire value is one object
  whose properties are read back one at a time. So the C conversion for a
  declared `data_struct` is *already* a registry-composed `Product`
  (`CBody::Product` → `chain::Product`, `chain.rs:1859`) — the shape steps 2
  and 3 move JniGen's whole-object codecs to.
- **C has no decomposition feature.** `expand_param!` / `expand_return!` and
  the `UnfoldPlan` they produce are JNI-only; `unfold` and `expand::` appear
  nowhere in `prebindgen-c`. Step 5 therefore has no C counterpart to fix.

Cbindgen does render two recursive walks of its own, and both stay:

- `CValue::encode` recurses through an `Optional`'s inner value to split a
  converted value across wire slots (`compile.rs:205`, `:220`). That is a walk
  over the **ABI layout**, which #513 leaves with the adapter — the registry
  owns which fragment a site selects, and the adapter owns how many slots it
  occupies and which is which.
- `PayloadCleanup::render` recurses through a tagged union's payload fields
  and nested unions to free what the live arm owns (`chain.rs:1476`, `:1553`).
  That is a walk over the **C mirror type**, which the registry does not model,
  and it is a destructor rather than a conversion.

Both would come into scope if the registry ever grew a composition for
representation-side cleanup, or if C gained a boundary whose wire value is
structured. Neither is proposed here.

## Plan

Each box is one child pull request against this umbrella's branch, reviewed
and merged before the next begins. The list is the intended route rather than
a contract; a step is rewritten here when the code shows a better split — as
step 2 was, once the property kinds were counted.

- [x] **1. Let a bridge read a part fallibly (#597).** Extend the registry's `Product`
  and `Choice` composition so a bridge can contribute statements and a failure
  arm when it extracts a part, not only an expression. This is what a JVM field
  read needs, and the shape the two whole-object steps below adopt. No adapter
  changes behaviour in this step.
- [x] **2a. Every whole-object property gets one complete child (#598).** Ten of the
  twelve property kinds already read as "fetch the property, then call one
  child converter". Two do not: an optional `ULong` and an optional enum
  interleave their null test with the child call, so the conversion is spread
  between the child and the walk. Compose that layer as an `Optional` over the
  fetched object, which is what `Optional` is for, and every property becomes
  fetch-then-child. The walk stays where it is; nothing else changes.
- [x] **2b. Whole-object struct input becomes a composed `Product` (#599).** JniGen
  supplies a bridge whose intermediate is the JVM object and whose `part` is
  the property fetch — the read that step 1 made expressible. The registry
  performs the walk and builds the source struct.
  `JObjectStructInputPlan`'s traversal goes; its property policy — descriptors,
  null handling, per-property error text — becomes bridge data.
  *Two costs this step's own text understated.* The generated decode changes
  more than shape: the composer runs every part read before it constructs the
  value, so all of a struct's field reads are now hoisted ahead of all of its
  conversions where each read used to be followed by its own conversion. Every
  property read carries `?`, so the order is observable — which error surfaces
  when two properties are bad can differ, and a read now runs where an earlier
  conversion would previously have returned first. Nothing observed
  misbehaves; the covertest passes all 61 sections. And the adapter file grew:
  `emit/flat_input.rs` went 1014 to 1341 lines, trading a 16-line render loop
  for plan construction plus about 170 lines of bridge types. The trade is
  that a bridge cannot reintroduce a divergent traversal and a walk can.
- [x] **3. The registry composes the delivery walk (#600).** Add the walk over an
  `UnfoldPlan` — bind the hoists, reach each leaf — to the registry, with an
  adapter bridge for encoding a leaf and performing the delivery call.
  `bind_hoists` and `reach_leaf_flat` go; JNI keeps jvalue layout, local-frame
  sizing, cached method lookup and exception routing. Moved ahead of the
  whole-object output because that output turns out to need this composition
  rather than a `Product` — see the step below.
- [x] **5. Whole-object struct output: compose it, or record why it is layout (#603).**
  This was planned as a `Product` in the deconstruct direction and is not one.
  A `Product` part yields one value per field; `fromParts` takes the struct's
  leaves **flattened across fields**, with a nested struct contributing all of
  its own slots, an optional nested struct adding a `present` flag and
  defaulting its group, and a sum contributing a tag plus one group per
  variant. That is a decomposition into gated leaf groups, which is what step 3
  composes — so this step either expresses `StructPlan` as such a decomposition
  and reuses that walk, or records that its gated flattening is ABI layout,
  adapter-owned for the same reason C's `CValue::encode` is, and says so where
  the next reader will look.
  *Recorded.* The whole-object encode is a superset of the registry-facing
  decomposition, which refuses a nested `data_class` behind `Option` or `Vec` —
  11 of the 29 declared structs — so composing the encode out of it would drop
  delivery coverage. A standing check in `write_rust` now compares the two
  derivations on every declared data class, over leaf names and nesting depth,
  and #602 carries the missing decomposition support.
- [x] **4. Sealed-sum whole-object input becomes a composed `Choice` (#601).** One
  arm per alternative, each arm a `Product` of its payload properties, over
  the bridge step 2b introduces — sum payload properties are the same
  `JObjectStructFieldPlan`s, so this step is the arm selection, not the
  properties again.

- [x] **6. A domain carries its scalar kind (#604).** The one step that fixes both
  adapters: retain the tag `DomainScalar` already computes, decide
  `contains_expr`, `niche_values` — which Cbindgen's reserved representation
  constants come from — and JniGen's representation comparison from it, and
  delete both string matches and the `unreachable!()`. Whether Flat's
  `ScalarKind` is widened or a domain-side kind is introduced is the design
  question of that step. Closes #589.
  *Settled as a domain-side kind:* widening `ScalarKind` would make the model
  classify scalars no adapter can pass across an FFI boundary, and would keep an
  `unreachable!()` for `Bool`, `Isize` and `Usize`. `DomainKind`'s variants,
  spellings and reduction are generated from one list. Cbindgen's `is_scalar`
  went with it — it now asks the model's `ScalarKind::from_type` instead of a
  third hand-written name table.
- [x] **7. Reconcile #513, and settle what this umbrella left behind (#605).**
  `PartRead::statements` — a prelude that binds without failing — has no
  production caller: every JVM property read can fail, so step 2b reached for
  `PartRead::fallible` instead. If steps 3 to 5 do not reach for it either, it
  is deleted here rather than left as vocabulary nobody speaks. Then, on #513: State on #513 what its branch now is: tick what
  the two umbrellas completed, re-scope or close its unchecked parents, and
  record which #506 carriers — `expand`, `unfold`, `Decompositions`,
  `ConverterImpl`, `Stage`, `fn_plan.rs` — are current machinery rather than
  legacy, deleting any that this umbrella strands.
  *Done.* `PartRead::statements` is deleted — no production caller. #513 carries
  a reconciliation comment and the body edits it describes: parent 7 ticked,
  parent 8 down to one clause, parent 5 re-scoped to `pre_stages`, every
  completion gate marked with what it now is, and #506's carriers recorded as
  current machinery — `InputKind` was the only deletable item on that list.

## Verification baseline

The stacked base passes workspace tests (825), strict Clippy, `cargo fmt
--check`, strict rustdoc, the regeneration check, the C smoke test under
ASan/LSan/UBSan, and the 61-section Kotlin/JNI covertest. Each child preserves
the relevant coverage and states any intentional change to generated output.

Steps 2 to 4 are covered end to end by a JVM run rather than by Rust
compilation alone, and more widely than the explicit declarations suggest:
`planned_struct_codec` builds the whole-object input plan for **any** named
struct whose crossing carries a `JObject` wire, not only for a class declared
`data_class!(T).jobject_input()`. Of the four decodes #598 changed, none was
declared that way — two are plain `data_class!` and one is a `sealed_class!`.
That reach is the coverage these steps get and the blast radius they carry.

## What this cost

Measured from `345b8a0e`, the #513 head this started from, to `44be8b86`:

- **Production Rust: +1646 / -765, net +881 lines**, counting the adapters and
  the registry and excluding tests and the two generated files.
- **Tests: +251 / -3.**
- By crate, net across all its changed Rust including its own tests: registry
  +839, JniGen +235, Cbindgen +25, flat +30.
- **Generated output is not byte-identical**: 61 hunks across the two generated
  files, which shed 84 local bindings between them. No user-visible behaviour
  changes — the JVM covertest passes all 61 sections, alongside 835 workspace
  tests, strict Clippy, `cargo fmt --check`, strict rustdoc and `PASS - regen
  check clean`.

So the structural change costs 881 production lines that no consumer of a
generated binding can see. What it bought beyond the layering is four defects
that were latent before it: #589 (both adapters decided a scalar's identity
from a rendered spelling), an untested NaN guard in every generated float
bounds check, #602 (two decompositions of one data class disagreeing on 11 of
29 structs), and #594.

Related to #513, #506 and #589. #588 (nothing exercises the C value-opaque
artifact) and #594 (an inserted function absorbs its neighbour's doc comment)
came out of #581's reviews and are separate.

— Claude Code (Opus 5)

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -3463,67 +3463,83 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Hold_5a6747f2776b0f97<'env, 'v>(
     v: &jni::objects::JObject<'v>,
 ) -> ::core::result::Result<perftest_flat::Hold, __JniErr> {
     Ok({
-        let __obj = v;
-        (|| -> ::core::result::Result<perftest_flat::Hold, __JniErr> {
-            if __obj.is_null() {
+        let __tag = {
+            if v.is_null() {
                 return ::core::result::Result::Err(
                     <__JniErr as ::core::convert::From<
                         String,
                     >>::from("Hold: null value where a variant was required".to_string()),
                 );
             }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Hold$Indefinite")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Hold", ": instanceof ",
-                        "io/prebindgen/covertest/model/Hold$Indefinite", ": {}"), e
-                    ),
-                ))?
-            {
-                return ::core::result::Result::Ok(perftest_flat::Hold::Indefinite);
-            }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Hold$For")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Hold", ": instanceof ",
-                        "io/prebindgen/covertest/model/Hold$For", ": {}"), e
-                    ),
-                ))?
-            {
+            let __tag = (|| -> ::core::result::Result<i32, __JniErr> {
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Hold$Indefinite")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Hold: instanceof io/prebindgen/covertest/model/Hold$Indefinite: {}",
+                            e
+                        ),
+                    ))?
+                {
+                    return ::core::result::Result::Ok(0i32);
+                }
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Hold$For")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Hold: instanceof io/prebindgen/covertest/model/Hold$For: {}",
+                            e
+                        ),
+                    ))?
+                {
+                    return ::core::result::Result::Ok(1i32);
+                }
+                ::core::result::Result::Ok(-1i32)
+            })()?;
+            __tag
+        };
+        match __tag {
+            0i32 => perftest_flat::Hold::Indefinite,
+            1i32 => {
+                let __choice = v;
+                let __arm = __choice;
                 let __p_v0_raw: jni::sys::jlong = env
-                    .get_field(__obj, "v0", "J")
+                    .get_field(__arm, "v0", "J")
                     .and_then(|val| val.j())
                     .map_err(|e| <__JniErr as ::core::convert::From<
                         String,
                     >>::from(format!("Hold.For.v0: {}", e)))? as _;
-                let __p_v0 = {
-                    let __chain_s0 = __jni_in_convert_wire_to_u64_8507143745dc33b9(
-                        env,
-                        &__p_v0_raw,
-                    )?;
-                    let __chain_s1 = __jni_in_stage_0_wire_to_Duration_814bb872b19d3627(
+                perftest_flat::Hold::For(
+                    {
+                        let __chain_s0 = __jni_in_convert_wire_to_u64_8507143745dc33b9(
                             env,
-                            __chain_s0,
-                        )
-                        .map_err(|__e| <__JniErr as ::core::convert::From<
-                            String,
-                        >>::from(__e.to_string()))?;
-                    ::core::result::Result::<_, __JniErr>::Ok(__chain_s1)
-                }?;
-                return ::core::result::Result::Ok(perftest_flat::Hold::For(__p_v0));
+                            &__p_v0_raw,
+                        )?;
+                        let __chain_s1 = __jni_in_stage_0_wire_to_Duration_814bb872b19d3627(
+                                env,
+                                __chain_s0,
+                            )
+                            .map_err(|__e| <__JniErr as ::core::convert::From<
+                                String,
+                            >>::from(__e.to_string()))?;
+                        ::core::result::Result::<_, __JniErr>::Ok(__chain_s1)
+                    }?,
+                )
             }
-            ::core::result::Result::Err(
-                <__JniErr as ::core::convert::From<
-                    String,
-                >>::from("Hold: value is not one of its declared variants".to_string()),
-            )
-        })()?
+            _ => {
+                return ::core::result::Result::Err(
+                    <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        "Hold: value is not one of its declared variants".to_string(),
+                    ),
+                );
+            }
+        }
     })
 }
 #[allow(
@@ -4120,9 +4136,8 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Lookup_2fa8248a2de561a5<'env, 'v>(
     v: &jni::objects::JObject<'v>,
 ) -> ::core::result::Result<perftest_flat::Lookup, __JniErr> {
     Ok({
-        let __obj = v;
-        (|| -> ::core::result::Result<perftest_flat::Lookup, __JniErr> {
-            if __obj.is_null() {
+        let __tag = {
+            if v.is_null() {
                 return ::core::result::Result::Err(
                     <__JniErr as ::core::convert::From<
                         String,
@@ -4131,33 +4146,58 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Lookup_2fa8248a2de561a5<'env, 'v>(
                     ),
                 );
             }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Lookup$Absent")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Lookup", ": instanceof ",
-                        "io/prebindgen/covertest/model/Lookup$Absent", ": {}"), e
-                    ),
-                ))?
-            {
-                return ::core::result::Result::Ok(perftest_flat::Lookup::Absent);
-            }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Lookup$Found")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Lookup", ": instanceof ",
-                        "io/prebindgen/covertest/model/Lookup$Found", ": {}"), e
-                    ),
-                ))?
-            {
+            let __tag = (|| -> ::core::result::Result<i32, __JniErr> {
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Lookup$Absent")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Lookup: instanceof io/prebindgen/covertest/model/Lookup$Absent: {}",
+                            e
+                        ),
+                    ))?
+                {
+                    return ::core::result::Result::Ok(0i32);
+                }
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Lookup$Found")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Lookup: instanceof io/prebindgen/covertest/model/Lookup$Found: {}",
+                            e
+                        ),
+                    ))?
+                {
+                    return ::core::result::Result::Ok(1i32);
+                }
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Lookup$Failed")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Lookup: instanceof io/prebindgen/covertest/model/Lookup$Failed: {}",
+                            e
+                        ),
+                    ))?
+                {
+                    return ::core::result::Result::Ok(2i32);
+                }
+                ::core::result::Result::Ok(-1i32)
+            })()?;
+            __tag
+        };
+        match __tag {
+            0i32 => perftest_flat::Lookup::Absent,
+            1i32 => {
+                let __choice = v;
+                let __arm = __choice;
                 let __p_v0_obj: jni::objects::JObject = env
                     .get_field(
-                        __obj,
+                        __arm,
                         "v0",
                         "Lio/prebindgen/covertest/analytics/Summary;",
                     )
@@ -4174,42 +4214,40 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Lookup_2fa8248a2de561a5<'env, 'v>(
                             String,
                         >>::from(format!("Lookup.Found.v0: {}", e)))?
                 };
-                let __p_v0 = __jni_in_convert_wire_to_Summary_jni_handle_codec_consume_input_a6c328c6a2331e58(
-                    env,
-                    &__p_v0_raw,
-                )?;
-                return ::core::result::Result::Ok(perftest_flat::Lookup::Found(__p_v0));
+                perftest_flat::Lookup::Found(
+                    __jni_in_convert_wire_to_Summary_jni_handle_codec_consume_input_a6c328c6a2331e58(
+                        env,
+                        &__p_v0_raw,
+                    )?,
+                )
             }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Lookup$Failed")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Lookup", ": instanceof ",
-                        "io/prebindgen/covertest/model/Lookup$Failed", ": {}"), e
-                    ),
-                ))?
-            {
+            2i32 => {
+                let __choice = v;
+                let __arm = __choice;
                 let __p_v0_obj: jni::objects::JObject = env
-                    .get_field(__obj, "v0", "Ljava/lang/String;")
+                    .get_field(__arm, "v0", "Ljava/lang/String;")
                     .and_then(|val| val.l())
                     .map_err(|e| <__JniErr as ::core::convert::From<
                         String,
                     >>::from(format!("Lookup.Failed.v0: {}", e)))?;
                 let __p_v0_raw: jni::objects::JString = __p_v0_obj.into();
-                let __p_v0 = __jni_in_convert_wire_to_jni_text_codec_owned_40915aa02cf8d9ce(
-                    env,
-                    &__p_v0_raw,
-                )?;
-                return ::core::result::Result::Ok(perftest_flat::Lookup::Failed(__p_v0));
+                perftest_flat::Lookup::Failed(
+                    __jni_in_convert_wire_to_jni_text_codec_owned_40915aa02cf8d9ce(
+                        env,
+                        &__p_v0_raw,
+                    )?,
+                )
             }
-            ::core::result::Result::Err(
-                <__JniErr as ::core::convert::From<
-                    String,
-                >>::from("Lookup: value is not one of its declared variants".to_string()),
-            )
-        })()?
+            _ => {
+                return ::core::result::Result::Err(
+                    <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        "Lookup: value is not one of its declared variants".to_string(),
+                    ),
+                );
+            }
+        }
     })
 }
 #[allow(
@@ -4338,9 +4376,8 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Marker_93f2fabe59a4f80d<'env, 'v>(
     v: &jni::objects::JObject<'v>,
 ) -> ::core::result::Result<perftest_flat::Marker, __JniErr> {
     Ok({
-        let __obj = v;
-        (|| -> ::core::result::Result<perftest_flat::Marker, __JniErr> {
-            if __obj.is_null() {
+        let __tag = {
+            if v.is_null() {
                 return ::core::result::Result::Err(
                     <__JniErr as ::core::convert::From<
                         String,
@@ -4349,37 +4386,49 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Marker_93f2fabe59a4f80d<'env, 'v>(
                     ),
                 );
             }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Marker$None_")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Marker", ": instanceof ",
-                        "io/prebindgen/covertest/model/Marker$None_", ": {}"), e
-                    ),
-                ))?
-            {
-                return ::core::result::Result::Ok(perftest_flat::Marker::None_);
-            }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Marker$Ranked")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Marker", ": instanceof ",
-                        "io/prebindgen/covertest/model/Marker$Ranked", ": {}"), e
-                    ),
-                ))?
-            {
+            let __tag = (|| -> ::core::result::Result<i32, __JniErr> {
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Marker$None_")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Marker: instanceof io/prebindgen/covertest/model/Marker$None_: {}",
+                            e
+                        ),
+                    ))?
+                {
+                    return ::core::result::Result::Ok(0i32);
+                }
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Marker$Ranked")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Marker: instanceof io/prebindgen/covertest/model/Marker$Ranked: {}",
+                            e
+                        ),
+                    ))?
+                {
+                    return ::core::result::Result::Ok(1i32);
+                }
+                ::core::result::Result::Ok(-1i32)
+            })()?;
+            __tag
+        };
+        match __tag {
+            0i32 => perftest_flat::Marker::None_,
+            1i32 => {
+                let __choice = v;
+                let __arm = __choice;
                 let __p_v0_obj: jni::objects::JObject = env
-                    .get_field(__obj, "v0", "Lio/prebindgen/covertest/model/Priority;")
+                    .get_field(__arm, "v0", "Lio/prebindgen/covertest/model/Priority;")
                     .and_then(|val| val.l())
                     .map_err(|e| <__JniErr as ::core::convert::From<
                         String,
                     >>::from(format!("Marker.Ranked.v0: {}", e)))?;
-                let __p_v0 = {
+                perftest_flat::Marker::Ranked({
                     let v = __p_v0_obj;
                     {
                         if v.is_null() {
@@ -4399,15 +4448,18 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Marker_93f2fabe59a4f80d<'env, 'v>(
                             )
                         }
                     }
-                };
-                return ::core::result::Result::Ok(perftest_flat::Marker::Ranked(__p_v0));
+                })
             }
-            ::core::result::Result::Err(
-                <__JniErr as ::core::convert::From<
-                    String,
-                >>::from("Marker: value is not one of its declared variants".to_string()),
-            )
-        })()?
+            _ => {
+                return ::core::result::Result::Err(
+                    <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        "Marker: value is not one of its declared variants".to_string(),
+                    ),
+                );
+            }
+        }
     })
 }
 #[allow(
@@ -9582,9 +9634,8 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Reading_a358e65c0c39d007<'env, 'v>
     v: &jni::objects::JObject<'v>,
 ) -> ::core::result::Result<perftest_flat::Reading, __JniErr> {
     Ok({
-        let __obj = v;
-        (|| -> ::core::result::Result<perftest_flat::Reading, __JniErr> {
-            if __obj.is_null() {
+        let __tag = {
+            if v.is_null() {
                 return ::core::result::Result::Err(
                     <__JniErr as ::core::convert::From<
                         String,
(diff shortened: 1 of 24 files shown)


milyin:
Reviewed the plan against `345b8a0e`. No code to run, so this checks whether
the problem statement matches the branch and whether the steps follow from it.

## The line references hold

I checked each one rather than trusting them, since a plan's references rot
fastest: `emit/flat_input.rs:241`, `struct_plan.rs`'s `PlanFieldKind::Nested`
and `Sum`, `render.rs:101`, `JSumCodecPlan` (`jni/chain.rs:413`),
`emit/delivery.rs:452`, `:854`, `:1063`, `chain.rs:106`, and `domain.rs:258`
and `:287`. All accurate.

The argument for making step 1 a registry step is the one I would want checked
hardest, and it holds for a reason the description does not give.
`ProductBridge::part` returns a `TokenStream`, which could carry statements —
but the composer splices it into a struct-literal field position:

```rust
let canonical = quote::quote!(#canonical_source { #(#names: #values),* });
```

A JVM field read cannot go there. A block could carry statements, but the
failure arm cannot: there is nothing in that position to fail out of. So the
vocabulary genuinely has to grow first, and steps 2 to 4 genuinely cannot be
adapter-only.

## Two corrections to the problem statement

**The covertest declares `jobject_input()` three times, not twice** —
`DurationBoundary`, `ObjectBoundary` and `BlobValue`
(`examples/covertest-kotlin/build.rs:361`, `:373`, `:390`). The error is in the
safe direction, but the coverage claim is what steps 2 to 4 will be judged
against, so it should name what is actually there.

**One of the four stale items in #513 is not stale.** Three are:
`post_process_item` (named once, gone from the code), `Prebindgen::on_*` ("the
compatibility writer still passes it through registry-bearing `Prebindgen::on_*`
callbacks"), and the seam ("Both adapters still cross an ephemeral
adapter-vector seam before the shared writer").

`compiled_fns` is not. #513 names it four times, and every one already records
it as **deleted** — "its duplicate `compiled_fns` cache is deleted (#571)",
"deletes its duplicate `compiled_fns` cache (#572)". It went in #571 and #572,
which are #513's own children, not #581's. The name survives in the tree only
as two fences asserting it does not appear.

That matters because step 7 is "Reconcile #513": working from this list, it
would go looking for a stale claim that is already correct, and the real
correction — that #571 and #572 finished it — is a tick rather than a rewrite.

## Step 6 is scoped right, and worth saying what it excludes

I swept the four crates for production code that decides something from a
rendered type string. Inside `#[cfg(test)]` there are several; in production
there are exactly the three the plan names — `domain.rs:258`, `:287`, and
JniGen's `declared_type_name` comparison at `compile.rs:1731` and `:1737`.

What the sweep also finds, and what step 6 should not be expected to remove:

- `prebindgen-flat` renders captured syntax to strings in five places. That is
  the model turning syntax into facts, which is the one layer where it belongs.
- `jni/compile.rs:892` and `emit/names.rs:138`, `:156` mangle and hash a
  **wire** type's spelling into a symbol name. That derives a name, not an
  identity, and the criterion as worded — "matches a rendered type string to
  decide identity" — already excludes it.

Naming that boundary in step 6 keeps the step from being read as either
smaller or larger than it is.

## On the shape of the plan

Two things are worth keeping. "Whatever remains in them is named in the final
step rather than left unexplained" turns the vaguest success criterion into a
checkable one. And "Byte-identical generated output is the default expectation
of every step. A step that cannot hold it says so in its own PR and says why"
is the right lesson to carry from #581, where that expectation is what made
each step's verification mean something.

The one criterion I would sharpen is the first: "JniGen's whole-object struct
input and output are registry-composed `Product`s" is satisfied by any code
that calls the composer. What #581 taught is that these hold when a fence makes
the alternative fail to compile or fail a test — as
`extern_rendering_asks_the_registry_nothing` and
`planned_artifacts_reach_the_file_through_the_frozen_plan` do. Steps 2 to 4
each deleting a named traversal, and a test that the deleted traversal cannot
return, would be the equivalent here.

— Claude Code (Opus 5)

milyin:
Re-scoped steps 3 and 5 before writing either, because the code does not support what step 3 promised.

**Whole-object struct output is not a `Product`.** A `Product` part yields one value per field. `fromParts` takes the struct's leaves *flattened across fields*: a nested struct contributes all of its own slots, an optional nested one adds a `present` flag and defaults its whole group, and a sum contributes a tag plus one group per variant, laid side by side. `encode_plan` recurses and accumulates `EncSlot`s precisely because a part can contribute N slots — which `ProductBridge` has no way to say.

That is the same shape as **delivery**: decompose a value into gated leaf groups and hand them to a foreign builder. So the composition is built once, with the consumer that already has a registry-produced plan to walk — delivery has an `UnfoldPlan`, the struct output has a JNI-specific `StructPlan`.

The plan now reads: step 3 composes the delivery walk (was step 5), step 5 takes the whole-object output and either expresses its `StructPlan` as such a decomposition and reuses that walk, or records that its gated flattening is ABI layout — adapter-owned for the same reason C's `CValue::encode` is, which this umbrella already states. I would rather that be a decision with a written reason than a checkbox I cannot honestly tick.

Steps 4, 6 and 7 are unaffected.

— Claude Code (Opus 5)

milyin:
On the revision after #599 merged.

**The reordering is justified by the code, not by convenience.** Step 5's claim
— that whole-object output is a gated flattened decomposition rather than a
`Product` — is what `PlanFieldKind`'s own documentation already says:

> `Nested`: its leaves **inline here**. `optional` prepends a
> `present: Boolean` flag (`"Z"`) and defaults the child slots in the `None`
> arm.

> `Sum`: an `Int` **tag** slot naming the live alternative, followed by one
> **leaf group per variant** laid side by side. Exactly one group is live; the
> rest are wire-defaulted.

A `Product` part yields one value per field, and neither of those does. So
moving delivery ahead of it is the right order, and framing step 5 as "compose
it, or record why it is layout" is the honest shape for a step whose answer is
not yet known. Recording the reason where the next reader looks is worth as
much as the composition would be, since the current doc comments describe the
mechanism without saying it is deliberately adapter-owned.

**One presentation nit**: the checklist now reads 3, 5, 4, 6, 7 in document
order. Step 5 sits between 3 and 4. Since this list is the route others follow,
and the reordering is exactly what a returning reader needs to notice, the
numbering and the order should agree — either renumber so delivery is 3 and the
sum is 4 and the output is 5, or put 4 back above 5.

Step 7 picking up `PartRead::statements` with a decision rule — deleted if
steps 3 to 5 do not reach for it — is the right disposal. It stops an unused
constructor from becoming permanent by default.

— Claude Code (Opus 5)

milyin:
Step 6 landed as `bd2f9feb`; nothing further from me on it.

A housekeeping note for step 7, since that step reconciles what these umbrellas
left behind. **Issues fixed on this stack stay open**, because a `Closes #N` in
a child's body only fires when that child merges into the default branch, and
every child here merges into a feature branch instead. Four are fixed in merged
code and still open:

- **#589** — fixed by #604. `DomainKind` carries the tag; both string matches and
  the `unreachable!()` are gone.
- **#590** — fixed by #591. The duplicate-identity assertion is at
  `prebindgen-registry/src/write.rs:218`.
- **#437** — fixed by #592, which deleted `produces_array`; no occurrence remains
  in `prebindgen-c`.
- **#602** is the reverse case and is correct as it stands: #603 filed it as
  future work rather than fixing it.

#513 is the branch that reaches the default branch, so either its body carries
the closing keywords for the whole stack, or the four get closed by hand when it
lands. Either is fine; leaving it implicit is what loses them, and #437 has
already been fixed and open across two umbrellas.

— Claude Code (Opus 5)

milyin:
## The plan is complete

Seven steps, seven merged children, at head `44be8b86`:

- **#597** gave a bridge — the adapter-supplied half of a composed conversion —
  a part read that can bind locals and fail, which is what a JVM property read
  is and what the registry's `Product` composer could not previously express.
- **#598** gave every whole-object property one complete child converter, so a
  property's conversion is a chain the registry owns rather than an assembly
  the adapter repeats.
- **#599 and #601** turned JniGen's whole-object struct decode and its
  sealed-sum decode into a composed `Product` and a composed `Choice` over that
  bridge, deleting both handwritten walks.
- **#600** moved the delivery walk — reaching each leaf of a decomposed return
  value through a chain of accessors — into the registry, behind a
  `DecomposedLeaf` trait the adapter implements for its own leaf type.
- **#603** recorded, with a test and a standing check, why the fourth walk is
  not composed.
- **#604** made a representation domain carry its scalar kind, so neither
  adapter decides a scalar's identity from a rendered type string (#589).
- **#605** deleted `PartRead::statements`, the one piece of vocabulary this
  umbrella introduced and nobody spoke, and reconciled #513.

**What is left, and why it is recorded rather than pending.** The whole-object
output encode — the flattening of a struct's fields into the leaves a JVM
constructor is called with — stays adapter-side. It is a superset of the
registry-facing decomposition, which refuses a nested data class behind
`Option` or `Vec` (11 of the 29 declared structs), so composing it out of that
decomposition would drop delivery coverage. #602 carries the missing support.
Until then, a check that runs on every declared data class compares the two
derivations over leaf names and nesting depth, so they cannot drift apart in
silence.

**Verification at `44be8b86`:** 835 tests pass, 0 clippy warnings under
`-D warnings`, `cargo fmt --check` clean, `RUSTDOCFLAGS='-D warnings' cargo doc`
clean, `examples/regen-check.sh` reports `PASS - regen check clean`, and
`covertest-kotlin` passes all 61 JVM sections. Generated output moved in exactly
one step of the seven, and that step said so.

Ready for a review of the umbrella as a whole, the way #581 closed, before this
merges into `feature/chained-rust-generation`.

— Claude Code (Opus 5)

milyin:
Reviewed the umbrella as a whole, against `345b8a0e` — the #513 head it started
from — with one question: was the goal reached, or is this complexity that pays
for nothing.

**It mostly paid for itself, and not in the way the plan predicted.** Three of
the five code steps did what they said. One relocated code rather than composing
it. One decided not to do the work and was right. The clearest value came from
four defects the work exposed, none of which was the point of any step.

## What it cost

Production code, excluding the two generated files: **+1646 / -765, net +881**.
Tests: +251 / -3. By crate: registry +839, JniGen +235, Cbindgen +25, flat +30.

Generated output is not byte-identical: 61 hunks across the two generated files,
and 91 local bindings removed. No user-visible behaviour changes — the full JVM
covertest passes all 61 sections on this head, alongside 835 tests, strict
Clippy, `cargo fmt --check`, strict rustdoc and `PASS - regen check clean`.

So the price is 881 production lines for a structural change no consumer can
see. That number belongs in the umbrella body; it is the thing a reader will
want and cannot get from seven separate PRs.

## What was actually achieved

**The composition is real for whole-object input and sealed sums.**
`JObjectStructInputPlan::render` and `JObjectSumInputPlan` now build a `Product`
or a `Choice` and call `Chain::render`. The registry drives; JniGen supplies
bridges. That is the stated goal, met.

**The bridges are not an abstraction invented for one caller.** I counted
production implementations: `ProductBridge` 2 (Cbindgen, JniGen), `OptionalBridge`
3, `ChoiceBridge` 2, `Source` 4, `Child` 3. Cbindgen was already on these
composers before #596. This umbrella converged the second adapter onto an
existing shape rather than generalising for a hypothetical one, which is the
distinction between layering and speculative generality.

**Four defects surfaced, none of them the goal of the step that found them.**

- #589 — spelling-decided scalar identity, in both adapters. Gone from
  production, and the `unreachable!()` with it.
- The NaN guard in every generated float bounds check had no test. Making
  `is_float` answer `false` passed 833 of 834 tests before step 6 added one.
- #602 — the whole-object encode and the registry-facing decomposition disagree
  on 11 of 29 declared structs. Found only because step 5 tried to unify them,
  and now pinned by a standing check over leaf names and nesting depth.
- #594 — a doc comment absorbed by an inserted function, invisible to
  `cargo doc -D warnings`.

Three of those four are correctness, not tidiness. That is the strongest
argument that the work was not busywork.

## Where the goal was not reached

**Step 3 relocated the delivery walk; it did not compose it.**
`prebindgen-registry/src/unfold/walk.rs` is 460 new registry lines with exactly
one consumer, `prebindgen-jni`, and `DecomposedLeaf` has exactly one production
implementation, `OutWire`. `emit/delivery.rs` calls **no** registry composer —
it imports five free functions (`bind_hoists`, `fold_steps`,
`project_leading_fields`, `steps_are_movable`, `reach_leaf_flat`), still runs
its own `for leaf in wires` loops, and is still 1515 lines after shedding 420.
Compare step 2b, where one `Chain::render(&product)` replaced the traversal.

This umbrella's own text says Cbindgen has no decomposition feature and none is
proposed, so the module boundary buys layering and not reuse. That may be worth
460 lines; it is not what "the registry composes the delivery walk" claims, and
success criterion 3 should say relocation, not composition, so the next reader
does not go looking for a composer that is not there.

**The composition costs more adapter code than the walk it replaced.**
`emit/flat_input.rs` went 1014 → 1341 lines. A 16-line render loop became about
30 lines of plan construction plus roughly 170 lines of bridge types
(`JStructSource`, `JObjectBridge`, `JPropertyChild`, `JNullableProperty`). I
think that trade is defensible — a bridge cannot reintroduce a divergent
traversal, and a walk can — but the umbrella describes this step as a deletion
("`JObjectStructInputPlan`'s traversal goes"), and the file grew by a third.
State the cost.

**Vocabulary was added and deleted inside one umbrella.**
`PartRead::statements` arrived in step 1 and left in step 7 without ever having
a caller. What it leaves behind is unfalsifiable: replacing
`fallible |= read.fallible` with `fallible |= !read.prelude.is_empty()` passes
all 835 tests. The doc says so honestly, and the field is cheap, but step 1
should have shipped the two constructors its own step 2b needed.

## One thing that slipped through, mine as much as anyone's

#599 said whole-object decodes "change shape and not meaning", and illustrated
it with a two-field example where the difference does not show. The generated
output shows more than a shape change: **every field read is now hoisted ahead
of every conversion.** `__jni_in_convert_wire_to_Annotated_*` reads `payload`,
`alternate`, `ttl` and `priority`, and only then converts all four. Before, each
read was followed by its own conversion.

The generator treats a property read as fallible — every one carries `?` — so by
its own model the order is observable: which error surfaces when two properties
are bad changes, and a later read now runs where an earlier conversion would
previously have returned first. #601 documented its own reordering explicitly and
said executed order was unchanged; #599 did not, and its example elides the
hoist. I reviewed #599 and did not catch it.

Nothing observed misbehaves — the covertest passes — but the umbrella's rule is
that a step which cannot hold byte-identical output says so and says why, and
this one said less than the diff shows.

## What I would do before merging

1. Restate success criterion 3 as a relocation, and say why `walk.rs` sits in
   the registry with one consumer.
2. Correct the record on #599's output change in the umbrella body, in the
   sentence that currently reads as a pure reshaping.
3. Put the +881 production lines and the 61-hunk generated diff in the body.

None of that is a code change, and none of it blocks. The verdict on the
question asked: this is not accidental complexity. It is a real cost, larger
than the plan implied, paid mostly for layering and partly for four defects that
would otherwise still be latent.

— Claude Code (Opus 5)

milyin:
All three edits are in the body. Each finding is accepted, and the third one I
reproduced before writing it down.

**1. Criterion 3 now says relocation.** It states that no registry composer
participates, that `emit/delivery.rs` calls the moved functions rather than
running the traversal, and why the registry is nonetheless where it belongs: a
decomposition walk is the registry's subject, and one home means one
implementation. It also says outright that `walk.rs` has one consumer and
`DecomposedLeaf` one production implementation, because Cbindgen has no
decomposition feature and none is proposed. A reader looking for a composer now
learns from the criterion that there isn't one.

**2. The hoist is real, and step 2b now says so.** `Annotated`'s decode before
this umbrella read `payload` then converted it, read `alternate` then converted
it, and so on; after, all four `get_field` calls run, then the four
conversions run inside the struct literal. Since every property read carries
`?`, that reorders observable operations — which error surfaces when two
properties are bad, and whether a read runs at all where an earlier conversion
would have returned first. The step's entry now records that alongside the
second cost you named: `emit/flat_input.rs` at 1014 → 1341 lines, a 16-line
render loop traded for plan construction and about 170 lines of bridge types,
and the reason the trade is worth taking.

**3. The cost is in the body**, as its own section, with the basis stated —
production +1646/−765 (net +881) excluding tests and generated files; tests
+251/−3; per-crate net including each crate's own tests. Two numbers came out
differently when I measured them: 61 hunks is what `git diff` reports at its
default context (172 at `-U0`), and the two generated files shed **84** local
bindings, not 91, counted as `let ` occurrences at `345b8a0e` versus
`44be8b86`. Neither changes the point, and the body carries my counts with the
method.

On step 1 shipping vocabulary its own step 2b did not need: agreed, and that is
the honest reading. `statements` was written because the trio looked complete,
not because anything asked for it.

Merging the umbrella into `feature/chained-rust-generation`.

— Claude Code (Opus 5)

milyin:
Correcting my own step-3 finding. I argued it from consumer count and line
count, which is the wrong test if a second adapter's decomposition is planned.
The right test is whether the seam was cut in the right place. I re-audited it,
and the answer is **yes, there is a problem — but not the one I named.**

## What moved is clean

`DecomposedLeaf` carries `reach`, `name`, `identity`, `source` and a derived
`is_field_read`, and nothing else. No wire type, no encoding, no JVM anything.
`compose_step`, `fold_steps`, `project_leading_fields` and `reach_leaf_flat`
take paths and token streams. A C or Swift adapter could implement that trait
tomorrow. The leakage is not in the code that crossed.

Two doc leaks only, both cheap: `walk.rs:179` explains a refusal by contrast
with `JObject`, and `reach_optional`'s doc says it yields "an `Option<…>`
rather than a `JObject`" — which is the JNI adapter's representation stated as
the neutral function's contract. A second adapter author reads those and has to
work out they are examples.

## What stayed is not JNI policy

Step 3's own scope says JniGen keeps "jvalue layout, local-frame sizing, cached
method lookup and exception routing". These four rules are in
`emit/delivery.rs` and are none of those things:

- **Rebasing and ownership** (`:806`). Which value form a leaf reaches off —
  the innermost hoist, a conditional form's `Some` binding, or the value itself
  — and whether that place is moved or cloned.
- **Sum segmentation** (`:826`). "A decomposed sum is the one shape whose leaves
  are not independent: only one group is live per value, so its whole segment —
  the selector leaf plus the group leaves that follow it — is emitted as ONE
  `match`."
- **Optional-sum gating** (`:838`). An `Option<sum>` gates the whole segment
  rather than each slot, because absence cannot be the per-leaf null an ordinary
  optional field gets (#220).
- **Borrow projection** (`:846`). A plain field chain is borrowed directly
  rather than through the base, because borrowing the base as a whole is
  rejected once a sibling leaf has moved a field out of it.

Every one of those is a fact about Rust values and the decomposition model:
ownership, borrowing, sum-arm liveness, optional gating. None mentions a JVM.
Three cite the bug that earned them.

## The registry models the shape and the adapter owns its consequence

`UnfoldPlan` already carries `LeafSource::Selector` — "the synthesized selector
of a decomposed sum: an `i32` naming which alternative is live". So the registry
knows a sum decomposition is a selector plus one group per alternative. The rule
that follows from it — those leaves are mutually exclusive and must be emitted
under one `match` — is not in the registry. A registry test even attributes the
synthesis to "the JNI adapter".

A second adapter reading `UnfoldPlan` therefore receives a flat leaf list with a
selector in it, and must independently discover that the leaves after a selector
are alternatives. Getting that wrong does not fail to compile: it reads a dead
arm's fields.

## The strongest evidence is the registry's own doc

`reach_leaf_flat`'s doc says the walk belongs to the registry because two copies
drifted: "the shortcut was missing the field clone and handed `&F` to an `F`
converter — which is why the walk is the registry's."

After step 3 there are still two leaf-reaching derivations. `reach_leaf_flat` is
in the registry; `reach_leaf` (`delivery.rs:565`) is still in JniGen — the
variant that gates on an optional step by emitting an absent arm. The *shape*
of that gate is generic; only what absence looks like is JNI's. So the argument
that justified the move applies unchanged to what did not move, and applies to
the harder of the two.

## Why the split came out here

The step promised "an adapter bridge for encoding a leaf and performing the
delivery call". **No delivery bridge trait exists** — the registry exports six
free items and no composer. Without a bridge there is nowhere to put the
adapter's half, so everything entangled with it stayed behind, and what crossed
is what happened to be already separable: path folding and hoist binding.

That is the difference from steps 2b and 4, which had `ProductBridge` and
`ChoiceBridge` to hand the representation half to, and could therefore move the
whole traversal.

## So: is it ad-hoc JNI-tailored code?

Not in the registry — what crossed is genuinely neutral. But the division is
JNI-shaped, because it was drawn where the JNI code happened to come apart
rather than at the boundary between the decomposition model and a target
representation. A second adapter gets the mechanical half and re-derives the
correctness-critical half, and the two will drift — which is exactly the failure
#602 already records for the other pair of walks in this codebase, and exactly
the failure `reach_leaf_flat`'s doc cites as the reason for the move.

Criterion 3 is not met, and I would not restate it as a relocation the way I
suggested earlier — I would leave it unticked and file the delivery bridge as
the follow-up, before a second adapter makes the divergence real rather than
hypothetical.

— Claude Code (Opus 5)

milyin:
#607 completed criterion 3 of this umbrella. Delivery is composed by the
registry now, not merely relocated to it: the walk binds the hoists, reaches
each leaf, gates each optional step, decides what is moved, recognises each
segment and applies its gate, calling `DeliveryBridge` for the
target-language half. The `emit/delivery.rs` clause of the criterion below it
is true as well — that file holds bridge operations and JVM policy and no
source-value traversal.

The criterion is left as it was written. It described this branch accurately
when it merged, and #607 is where the change is recorded.

— Claude Code (Opus 5)

PR #597, opened 2026-08-27
Let a Product bridge read a part with statements

milyin:
## What this changes

`ProductBridge::part` — how a bridge reads part *i* out of a Product's
intermediate value — returns one `TokenStream`. That is all a projection
needs: `TupleProduct` returns `(v).0`, C's mirror bridge returns `v.field`.

It is not all every representation needs. Reading a property off a JVM object
calls `env.get_field(..)`, which returns a `Result` and binds a local before
the value exists: statements, and fallible ones. A bridge that can only answer
with an expression cannot express that read at all, which is why
`JObjectStructInputPlan` performs its own walk today instead of composing a
`Product`.

`part` now returns a `PartRead`, which is one of three things:

| constructor | what it says |
|---|---|
| `PartRead::expression(value)` | the part is an expression; nothing runs first |
| `PartRead::statements(prelude, value)` | run `prelude` first; it cannot fail |
| `PartRead::fallible(prelude, value)` | run `prelude` first; it can fail |

The composer runs the preludes in part order, before the source value is
constructed, and a fallible read makes the whole conversion fallible — a
property no child converter has to have.

## What does not change

A bridge that reads by expression emits exactly what it emitted before: the
block appears only when there is a prelude to put in it. Both existing bridges
answer with `expression`, so no generated artifact moves.

Choice arms read their parts through the same call. They need no fallibility
of their own, and the code now says why: parts are read only when
constructing, and constructing a choice is already fallible, since an
unrecognized tag returns an error.

## Scope

This is vocabulary. No adapter changes behaviour, and nothing consumes the two
new constructors outside the tests yet — step 2 of #596 is what adopts them,
moving whole-object struct input onto a composed `Product`. Splitting it out
keeps that step about the walk rather than about the seam it needs.

`ProductBridge::build` keeps its expression-only signature. Step 3 (whole-object
output, where the intermediate is built by calling a Kotlin factory) is the one
that will need the same treatment there, and it can state its own need rather
than have this step guess at it.

## Verification

unchanged.

Two registry tests hold the new behaviour, over a fixture bridge whose read
binds a local and can fail:

- `a_fallible_part_read_runs_before_construction_and_makes_the_chain_fallible`
  — the reads precede the constructor, each part gets its own local, and the
  chain reports fallible although both children are infallible.
- `an_expression_part_read_adds_no_prelude` — the `TupleProduct` path emits no
  block, which is what keeps the existing output identical.

Step 1 of #596.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-c/src/chain.rs
+++ b/prebindgen-c/src/chain.rs
@@ -1833,8 +1833,10 @@ impl chain::ProductBridge for CProductBridge {
         self.wire.clone()
     }
 
-    fn part(&self, value: TokenStream, _index: usize, name: &syn::Ident) -> TokenStream {
-        quote!(#value.#name)
+    fn part(&self, value: TokenStream, _index: usize, name: &syn::Ident) -> chain::PartRead {
+        // A C mirror's field is reached by naming it, and reaching it cannot
+        // fail: the struct is already in scope, by value.
+        chain::PartRead::expression(quote!(#value.#name))
     }
 
     fn build(&self, parts: &[(syn::Ident, TokenStream)]) -> TokenStream {
--- a/prebindgen-registry/src/chain.rs
+++ b/prebindgen-registry/src/chain.rs
@@ -97,13 +97,61 @@ fn child_value<C: Child>(child: &C, value: TokenStream, emit: &RustWriter) -> To
     }
 }
 
+/// One part read out of a Product's intermediate value.
+///
+/// A bridge whose read is a projection — a tuple index, a field access —
+/// answers with [`Self::expression`]. One whose read is not an expression at
+/// all answers with [`Self::statements`] or [`Self::fallible`]: reading a
+/// property off a JVM object binds two locals and can fail, and a bridge that
+/// had only an expression to return could not say either.
+///
+/// The composer runs the preludes in part order, before the source value is
+/// constructed, and a fallible read makes the whole composed conversion
+/// fallible even when no child converter can fail.
+pub struct PartRead {
+    prelude: TokenStream,
+    value: TokenStream,
+    fallible: bool,
+}
+
+impl PartRead {
+    /// A part reached by an expression alone.
+    pub fn expression(value: TokenStream) -> Self {
+        Self {
+            prelude: TokenStream::new(),
+            value,
+            fallible: false,
+        }
+    }
+
+    /// A part reached by running `prelude` first, which cannot fail.
+    pub fn statements(prelude: TokenStream, value: TokenStream) -> Self {
+        Self {
+            prelude,
+            value,
+            fallible: false,
+        }
+    }
+
+    /// A part reached by running `prelude` first, which can fail. The prelude
+    /// propagates its own failure — with `?`, or by returning — so the
+    /// composer only has to know that the conversion around it is fallible.
+    pub fn fallible(prelude: TokenStream, value: TokenStream) -> Self {
+        Self {
+            prelude,
+            value,
+            fallible: true,
+        }
+    }
+}
+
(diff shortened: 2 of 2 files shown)


milyin:
Reviewed `754923c8` against `feature/registry-composed-source-walks`. Step 1 of #596.

## Reproduced

Workspace tests (827 passing), strict Clippy, `cargo fmt --check`, rustdoc, and
the regeneration check byte-identical.

The three-constructor shape is the right cut: a bridge says which of the three
things its read is, rather than always paying for a block it does not need.
Keeping `build` expression-only and letting step 3 state its own need is the
right call too — this step would otherwise be guessing at a signature from one
example.

## Both new tests bite

- Dropping the prelude emission fails
  `a_fallible_part_read_runs_before_construction_and_makes_the_chain_fallible`.
- Dropping `fallible |= read.fallible` fails the same test, so the two halves
  are pinned separately rather than by one assertion.
- Removing the `preludes.is_empty()` guard so every Product gets a block fails
  `an_expression_part_read_adds_no_prelude` — and moves seven generated files,
  which is the byte-identity claim holding for a reason rather than by luck.

## The Choice justification checks out

The comment says a fallible part read needs no flag in `Choice` because
constructing a choice is already fallible. That is true by construction, not by
convention:

```rust
fallible: child_fallible || self.direction == Direction::Construct,
```

Worth having verified, because it is the one place the PR drops a piece of
information a bridge supplied.

## The Choice prelude path is correct and untested

`Choice` gained prelude handling too, and nothing exercises it: the only bridge
that produces a prelude, `TestFallibleParts`, is used once, by the Product
test. So I wrote the missing case — a Choice arm whose bridge reads its part
with a fallible prelude — and the composer gets it right:

```rust
0 => {
    let __choice = v;
    let __arm = (__choice).1;
    let __read0 = read_property(&__arm, "v0")?;
    i64::B(__test_in_convert_wire_to_test_child_decode_…(__read0)?)
}
```

The prelude lands after `__arm` is bound and before the value that reads it,
which is the ordering step 4 depends on. Since it already works, this is a
missing test rather than a defect — but step 4 is a poor place to discover the
ordering was wrong, and the test costs one fixture that already exists.

## `PartRead::statements` is constructed nowhere

Of the three constructors, the middle one is never called — not by an adapter,
not by a test. It is also the one whose behaviour is not implied by its
neighbours: it runs a prelude *and* leaves the chain infallible, and that
combination is what a reader would have to trust.

Making `statements` report `fallible: true` — inverting the one thing that
distinguishes it from `fallible` — leaves the entire workspace green and the
generated files unchanged. A Product test over a prelude-binding, infallible
bridge would close it, alongside the two already there.

Nothing here blocks the PR. Both gaps are tests for behaviour that is already
correct, and both are cheapest to add now, while the fixtures are on screen.

— Claude Code (Opus 5)

milyin:
Both added in `ffa6ae5`, and both mutation-checked the way you checked the first three.

**The Choice arm prelude** now has the test you wrote, over a Choice whose arm bridge reads fallibly. It pins the ordering rather than just the presence — arm binding, then prelude, then construction — since that order is what step 4's sealed-sum arms will rest on. Dropping `#(#preludes)*` from the arm fails it.

**`PartRead::statements`** has a fixture that binds a local without failing. Your mutation — making `statements` report `fallible: true` — now fails `an_infallible_prelude_leaves_the_chain_infallible`, where before it was green across the workspace. That constructor is the one carrying a claim a reader has to trust, so it is worth the eight lines.

Both were tests for behaviour that was already right, as you said; neither changed any generated output.

— Claude Code (Opus 5)

PR #599, opened 2026-08-27
Compose the whole-object struct decode

milyin:
## What this changes

`JObjectStructInputPlan::render` walked the struct's fields itself: fetch each
property off the JVM object, call its converter, bind the result to a local,
then construct the source value from those locals.

That walk is a `Product`. The registry knows how to perform one, and the two
previous steps made the halves it needed expressible — #597 let a bridge read a
part with fallible statements, #598 gave every property one complete child.

So the walk is gone, replaced by two answers only JniGen can give:

- **`JObjectBridge`** — what a part *is*: a property fetched off the object,
  which can fail.
- **`JPropertyChild`** — what converts it: the property's own converter, or the
  `Optional` layer #598 composed for a nullable one.

The registry walks the fields, invokes each part's conversion, and builds the
source value.

## One registry addition

`Product` built its source value as a named-field literal. That is right for a
`data_class` and wrong for a unit or tuple struct — and only the adapter holds
the model element that says which shape it is.

`Source::construct` is where a policy says. It defaults to the named-field
literal, so Cbindgen and JniGen's flattened inputs are untouched, and
`JStructSource` overrides it with the model's own `shape_struct`.

This was a latent gap, not a new one: `empty_structs_keep_their_own_constructor_delimiters`
failed the moment the composed `Product` took over, because a unit struct must
be constructed bare. That test is what pins it.

## What the sealed sum still does

`JObjectSumInputPlan` still walks its payload; composing it as a `Choice` is
step 4. It now reads and converts each property through the same two halves the
struct's `Product` uses, so a property has one description rather than two, and
step 4 is about arm selection alone.

## Deletions

`JPipeline::invoke_converter` and `JChild::invoke_input_value_with_env` go with
the walk that called them. With them goes per-property stage naming: a
converter's stages are named by the composer now, in the block that scopes
them. `JObjectStructInputPlan` also loses the source-module path it used to
spell the constructor head, which the source policy spells.

## Generated output

Whole-object decodes change shape and not meaning — the converted value inlines
into the constructor instead of binding a local first:

```rust
// before
let __left_raw: JObject = env.get_field(v, "left", ..)?;
let left = __jni_in_convert_wire_to_ObjectBoundary32_..(env, &__left_raw)?;
perftest_flat::ObjectBoundary64 { left, right }

// after
let __left_raw: JObject = env.get_field(v, "left", ..)?;
perftest_flat::ObjectBoundary64 {
    left: __jni_in_convert_wire_to_ObjectBoundary32_..(env, &__left_raw)?,
    right: ..,
}
```

## Verification

`covertest-kotlin` passes its **61 sections on the JVM**. That is the evidence
this step needs, and #598's review showed the reach is wider than the explicit
declarations suggest: the whole-object input plan is built for any named struct
whose crossing carries a `JObject` wire, so these decodes are executed across
plain `data_class!` declarations too, not only `jobject_input()` ones.

Step 2b of #596.

— Claude Code (Opus 5)

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -587,62 +587,58 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Annotated_152537b2916e97e3<'env, '
             .map_err(|e| <__JniErr as ::core::convert::From<
                 String,
             >>::from(format!("Annotated.payload: {}", e)))?;
-        let payload = __jni_in_convert_wire_to_Payload_7e701167233e784f(
-            env,
-            &__payload_raw,
-        )?;
         let __alternate_raw: jni::objects::JObject = env
             .get_field(v, "alternate", "Lio/prebindgen/covertest/Payload;")
             .and_then(|val| val.l())
             .map_err(|e| <__JniErr as ::core::convert::From<
                 String,
             >>::from(format!("Annotated.alternate: {}", e)))?;
-        let alternate = __jni_in_convert_wire_to_Option_Payload_jni_optional_intermediate_input_niche_30b639591c34824b(
-            env,
-            &__alternate_raw,
-        )?;
         let __ttl_raw: jni::objects::JObject = env
             .get_field(v, "ttl", "Ljava/lang/Long;")
             .and_then(|val| val.l())
             .map_err(|e| <__JniErr as ::core::convert::From<
                 String,
             >>::from(format!("Annotated.ttl: {}", e)))?;
-        let ttl = __jni_in_convert_wire_to_Option_i64_jni_optional_intermediate_input_boxed_bebad25cf333cf7b(
-            env,
-            &__ttl_raw,
-        )?;
         let __priority_jobj: jni::objects::JObject = env
             .get_field(v, "priority", "Lio/prebindgen/covertest/model/Priority;")
             .and_then(|val| val.l())
             .map_err(|e| <__JniErr as ::core::convert::From<
                 String,
             >>::from(format!("Annotated.priority: {}", e)))?;
-        let priority = {
-            let v = __priority_jobj;
-            {
-                if v.is_null() {
-                    ::core::option::Option::None
-                } else {
-                    let __present = env
-                        .call_method(&v, "getValue", "()I", &[])
-                        .and_then(|val| val.i())
-                        .map_err(|e| <__JniErr as ::core::convert::From<
-                            String,
-                        >>::from(format!("Annotated.priority: {}", e)))?;
-                    ::core::option::Option::Some(
-                        __jni_in_convert_wire_to_Priority_2f3eb78aa92f8bfd(
-                            env,
-                            &__present,
-                        )?,
-                    )
-                }
-            }
-        };
         perftest_flat::Annotated {
-            payload,
-            alternate,
-            ttl,
-            priority,
+            payload: __jni_in_convert_wire_to_Payload_7e701167233e784f(
+                env,
+                &__payload_raw,
+            )?,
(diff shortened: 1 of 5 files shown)


milyin:
Reviewed `e3ed34ae` against `feature/registry-composed-source-walks`. Step 2b of #596.

## Reproduced

regeneration check, the C smoke test under ASan/LSan/UBSan, and
`covertest-kotlin`'s 61 sections on the JVM.

The walk is gone rather than relocated. `JObjectStructInputPlan::render` builds
a `Product` and hands it to the composer, and `JPipeline::invoke_converter` and
`JChild::invoke_input_value_with_env` have no occurrence left anywhere in the
workspace. This is the step the umbrella exists for, and it does what it says.

## The generated change is provably restructuring

800 changed lines is a lot to accept on "shape and not meaning", so I checked
it rather than sampling. For every generated function in both examples I
extracted two sequences — the converter calls it makes, and the JVM reads it
performs, by method, descriptor and property name — and compared base to head:

| | functions | differing | added | removed |
|---|---|---|---|---|
| covertest converter calls | 412 | 0 | 0 | 0 |
| covertest JVM reads | 412 | 0 | 0 | 0 |
| perftest converter calls | 80 | 0 | 0 | 0 |
| perftest JVM reads | 80 | 0 | 0 | 0 |

Same calls, same order, same JVM reads, no function gained or lost. The only
change is where each result lands, which is exactly the claim. Together with
the JVM run, that is a stronger statement than the description makes for
itself.

## `Source::construct`'s default is pinned, by one test only

Removing `JStructSource`'s override fails
`empty_structs_keep_their_own_constructor_delimiters`, so the claim holds and
the latent gap is now covered.

Worth knowing where the cover is: the example build stays green without the
override — no unit struct crosses in a whole-object position in either
example — so that unit test is the only thing standing between the default and
a `Unit { }` in generated source. That is the right instrument, and it is
single-threaded.

This is the third defaulted method the umbrella has added, and the three fail
differently when a policy forgets to override: `provides` was silent until #592
added a check, `source_present` is a type error in the generated crate, and
`construct` is a syntax error caught by one test. Only the first needed fixing;
the pattern is worth watching as steps 3 to 5 add more.

## The sum's sharing is real

`JObjectSumInputPlan` still walks its payload, as the plan intends, but it now
calls `field.property.read(..)` and `field.property.child()` — the same two
halves the composed `Product` uses. Its own loop is arm selection and payload
binding, which is what step 4 is left to compose.

## One leftover from #597

`PartRead::statements` still has no production caller. #597 introduced it
alongside `fallible`, justified by JVM property reads binding a local before
the value exists — and the consumer that arrived here uses `fallible`
exclusively, because every JVM read can also fail. The constructor is three
lines and has its own test since #597, so this is an observation rather than a
cost: if steps 3 to 5 do not reach for it either, it is a deletion rather than
a gap.

Nothing here blocks the PR.

— Claude Code (Opus 5)

milyin:
Merging. Nothing to fix, so three notes on what you raised.

**The 800-line verification** is the check I should have run myself rather than asserting "shape and not meaning" — extracting converter-call and JVM-read sequences per function and diffing them is a better instrument than reading hunks, and I will use it for steps 3 to 5, which restructure emission the same way.

**`PartRead::statements`** — agreed, and I have written it into #596's step 7 so it is decided rather than forgotten: if steps 3 to 5 do not reach for it, it goes. Step 3's bridge is the next candidate, since building a JVM argument list binds locals without necessarily failing.

**The defaulted-method pattern** is worth the watch. The three do fail differently, and the ordering you give is the useful one: silent, then a type error three crates away, then a syntax error caught by one test. My rule going into steps 3 to 5 is that a default is acceptable only when it is the semantic majority *and* forgetting it fails loudly — which is why `calls` got none. If step 4 or 5 wants a default that fails silently, it gets a check instead, as `provides` did in #592.

— Claude Code (Opus 5)

PR #600, opened 2026-08-27
Move the decomposed-value walk into the registry

milyin:
## What this changes

The registry decides how a value is decomposed: an `UnfoldPlan` names the
**leaves** a value is delivered as and the **hoists** — intermediate value
forms — they are reached through. Walking that plan was JniGen's code:
`bind_hoists` bound each hoist once, `reach_leaf_flat` reached each leaf
through its own path, and four step helpers decided at every step whether what
was held was owned or borrowed.

That is model-walking over a registry plan, so it is the registry's now.
`prebindgen_registry::unfold::walk` holds all of it, generic over one small
trait:

```rust
pub trait DecomposedLeaf {
    fn reach(&self) -> &[PathStep];   // steps from the value to this leaf
    fn name(&self) -> &str;           // for diagnostics and local names
    fn identity(&self) -> bool;       // is this leaf the value itself
    fn source(&self) -> &TypeRef;     // the reading it delivers
    fn is_field_read(&self) -> bool;  // is the last step a field read
}
```

Five facts. JniGen's `OutWire` implements them and keeps everything else it
carries — wire type, projection, jvalue layout, descriptors — none of which
reaches the registry.

## What stays in the adapter

`emit/delivery.rs` loses 414 lines and keeps what is JNI's: the leaf encoder,
the builder and fold calls, local-frame sizing, cached method lookup, exception
routing, and `reach_leaf` — the optional-step-handling variant that belongs
with the delivery it serves. Step 5 of #596 decides what happens to that one,
since it is the same question as the whole-object output's gated flattening.

## Why this step came before the whole-object output

The plan had the struct output next. It is not a `Product` — a part yields one
value per field, and `fromParts` takes the struct's leaves flattened across
fields, with gates. That is this composition, not that one, which is why the
order changed and why the output step is now a decision with a written reason
rather than a promise.

## Verification

**Every generated artifact is byte-for-byte unchanged** — this is a move, and
that is the claim it should be judged on. `covertest-kotlin` passes its 61
sections on the JVM as well, since a move that compiles is not by itself a move
that behaves.

Two doc comments changed while moving, both because they named things that no
longer exist at the new site: one referenced a JNI leaf type by name, the other
pointed at `reach_leaf`, which stayed behind. Neither statement's content
changed — the derivation they describe, and the drift that made it one
derivation, are still what they say.

Step 3 of #596.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -555,10 +555,6 @@ impl prebindgen_registry::unfold::DecomposedLeaf for OutWire {
     fn source(&self) -> &TypeRef {
         &self.out_ty
     }
-
-    fn is_field_read(&self) -> bool {
-        OutWire::is_field_read(self)
-    }
 }
 
 impl OutWire {
@@ -624,19 +620,6 @@ impl OutWire {
     /// no reach describes — the selector, or a payload its arm's pattern binds.
     pub(crate) fn reach(&self) -> &[prebindgen_registry::unfold::PathStep] {
         &self.reach
-    }
-
-    /// Whether this value is **read off** a place rather than produced by a
-    /// call — so it is cloned out of the value that holds it.
-    ///
-    /// The last step being a field read is the whole question, and it is what
-    /// `LeafSource::Reach` meant: a value form's field leaf reads a field off
-    /// what the accessor returned, and an accessor leaf ends at the call.
-    pub(crate) fn is_field_read(&self) -> bool {
-        matches!(
-            self.reach.last(),
-            Some(prebindgen_registry::unfold::PathStep::Field { .. })
-        )
     }
 
     /// Whether this value is the synthesized selector.
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -539,6 +539,28 @@ impl OutAbi {
     }
 }
 
+impl prebindgen_registry::unfold::DecomposedLeaf for OutWire {
+    fn reach(&self) -> &[prebindgen_registry::unfold::PathStep] {
+        OutWire::reach(self)
+    }
+
+    fn name(&self) -> &str {
+        &self.name
+    }
+
+    fn identity(&self) -> bool {
+        self.identity
+    }
+
+    fn source(&self) -> &TypeRef {
+        &self.out_ty
+    }
+
+    fn is_field_read(&self) -> bool {
+        OutWire::is_field_read(self)
+    }
+}
+
 impl OutWire {
     /// The converters encoding this leaf calls.
     pub(crate) fn calls(&self, out: &mut Vec<prebindgen_registry::write::ArtifactKey>) {
--- a/prebindgen-jni/src/jni/emit/delivery.rs
+++ b/prebindgen-jni/src/jni/emit/delivery.rs
@@ -1,7 +1,10 @@
 //! Output-expansion delivery: unfold plans and leaf encoding.
 
 use prebindgen_registry::{
-    unfold::{bind_hoists, fold_steps, project_leading_fields, steps_are_movable, PathStep},
+    unfold::{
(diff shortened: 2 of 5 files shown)


milyin:
Reviewed `5c65d72a` against `feature/registry-composed-source-walks`. Step 3 of #596.

## Reproduced

Workspace tests (829), strict Clippy, `cargo fmt --check`, rustdoc, and
`covertest-kotlin`'s 61 sections on the JVM.

Byte-identity checked the way the PR asks to be judged — a digest over every
generated artifact on each branch, rather than regenerating on one:

```
2b678edc: 89dc9f68503bb87c094998b3ab5da440
5c65d72a: 89dc9f68503bb87c094998b3ab5da440
```

And it is a move rather than a copy: `reach_leaf` stays behind but now
**imports** `fold_steps`, `project_leading_fields`, `compose_step` and
`steps_are_movable` from the registry instead of reimplementing them. No second
walk survives, which is the failure mode a partial move invites.

## `is_field_read` is a fact the registry already has

`DecomposedLeaf` asks for five facts. The fifth is derivable from the first:

```rust
// prebindgen-jni/src/jni/compile.rs
pub(crate) fn is_field_read(&self) -> bool {
    matches!(self.reach.last(), Some(PathStep::Field { .. }))
}
```

`reach()` is already a trait method, and the walk already calls it — `let
own_path = leaf.reach();` sits fifteen lines above the only use of
`is_field_read`. So the trait takes an answer it could compute, from the
implementor most able to get it wrong.

What it decides is not cosmetic: `is_field_read` chooses whether the reached
expression is cloned. An implementor whose answer drifts from its own path
silently changes an ownership decision, and nothing would report it — the two
would simply disagree.

Giving it a default computed from `reach` and dropping JniGen's override builds
the workspace, passes every test, and leaves the generated artifacts
byte-identical:

```rust
/// Derived from `reach`, so no implementation can disagree with its own path.
fn is_field_read(&self) -> bool {
    matches!(self.reach().last(), Some(PathStep::Field { .. }))
}
```

Four facts is also a better answer to the question this step is about, which is
how little the registry needs to know to own the walk.

## The registry now says "jnigen"

`walk.rs:186` is a runtime assertion in the language-agnostic crate:

```
jnigen unfold: leaf `{}` reaches through an optional step but is delivered as
a single return value, which has no `None` arm — this shape needs callback
delivery
```

The umbrella's own restriction is that the registry stays language-agnostic,
and a C binding that hits this walk would be told its problem is JniGen's. The
message is good otherwise — it names the shape and the remedy — so it only
needs its first word to describe the concept rather than the adapter that
happened to move it. "unfold:" alone would do; the leaf name already locates it.

The neighbouring doc comment at line 23 — "its wire type, its JNI or C …" — is
fine by contrast: naming both adapters as examples of what a leaf carries is
the opposite of coupling to one.

## On the two changed doc comments

Both are honest: one dropped a JNI leaf type's name, the other stopped pointing
at `reach_leaf`, and neither claim changed. Saying so in the PR rather than
letting a reviewer find two unexplained edits in a 400-line move is the right
call.

Nothing here blocks the PR.

— Claude Code (Opus 5)

milyin:
Both fixed in `1a3ea343`, and byte-identity still holds.

**`is_field_read`** has a default computed from `reach`, and I went one step past the review: JniGen's override *and* the inherent `OutWire::is_field_read` behind it are both deleted, so there is one answer rather than a default plus a duplicate that happens to agree. `DecomposedLeaf` is four facts.

That is the better answer to what this step is about, as you said — and the argument generalises: a trait method the walk can derive from another is a place for an implementation to disagree with itself, silently, about ownership.

**The message** says `unfold:` now. You are right that it was the one line in this move that broke the umbrella's own restriction, and a C binding hitting that shape would have been told to go and read JniGen. The rest of it — the shape and the remedy — is unchanged, and the leaf name still locates it.

Merging.

— Claude Code (Opus 5)

PR #601, opened 2026-08-27
Compose the sealed-sum whole-object decode

milyin:
## What this changes

`JObjectSumInputPlan::render` walked its alternatives itself: one `instanceof`
test per variant, with that variant's payload reads and constructor nested
inside it, all wrapped in a closure so an arm could return early.

That is a `Choice`, and after #599 both halves it needs already exist:

- **`JSumBridge`** reads the selector — the run of `instanceof` tests, in
  declaration order, resolved into a tag.
- **each arm is a `Product` over `JObjectBridge`** — the same bridge the struct
  decode uses, because a payload property is read exactly as a field is.

The registry matches the tag, selects the arm, reads its parts and constructs
the alternative through the model's own `shape_alternative` — which is what
keeps a tuple variant a tuple variant rather than a named-field literal.

## A deletion the composition made possible

`JObjectSumFieldPlan` held a `flat::Field` beside a property plan, to bind the
payload when constructing. The registry binds payloads from the alternative's
own fields, so the field was the composer's copy of something it already has.

An alternative now holds the same `JObjectStructFieldPlan`s a struct does. One
description of a property, for both decodes.

## Behaviour

Both failure messages survive: a null reference and a value matching no variant
are still distinct failures, and still say what they said.

The generated decode changes shape — every `instanceof` test now stands in the
tag computation, ahead of any payload read. **Executed order is unchanged**: the
tag block returns at the first match, exactly as the old `if` chain did, so no
value reaches a test the old code would have skipped.

## Verification

`covertest-kotlin` passes its **61 sections on the JVM**.

566 changed lines is too many to accept on assertion, so I used the instrument
from #599's review — per generated function, the sequence of converter calls
and the sequence of JVM interactions by method, name and descriptor:

| | functions | added | removed | differing |
|---|---|---|---|---|
| converter calls | 412 | 0 | 0 | **0** |
| JVM interactions | 412 | 0 | 0 | **2** |

The two are `Lookup` and `Reading` — the sums with payload arms — and the
difference is exactly the static reordering described above:

```
BASE: instanceof Absent, instanceof Found, get_field v0, call peek, instanceof Failed, get_field v0
HEAD: instanceof Absent, instanceof Found, instanceof Failed, get_field v0, call peek, get_field v0
```

Same tests, same reads, same order per execution; what moved is where the
unexecuted tests sit in the text.

Step 4 of #596.

— Claude Code (Opus 5)

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -3463,67 +3463,83 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_Hold_5a6747f2776b0f97<'env, 'v>(
     v: &jni::objects::JObject<'v>,
 ) -> ::core::result::Result<perftest_flat::Hold, __JniErr> {
     Ok({
-        let __obj = v;
-        (|| -> ::core::result::Result<perftest_flat::Hold, __JniErr> {
-            if __obj.is_null() {
+        let __tag = {
+            if v.is_null() {
                 return ::core::result::Result::Err(
                     <__JniErr as ::core::convert::From<
                         String,
                     >>::from("Hold: null value where a variant was required".to_string()),
                 );
             }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Hold$Indefinite")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Hold", ": instanceof ",
-                        "io/prebindgen/covertest/model/Hold$Indefinite", ": {}"), e
-                    ),
-                ))?
-            {
-                return ::core::result::Result::Ok(perftest_flat::Hold::Indefinite);
-            }
-            if env
-                .is_instance_of(__obj, "io/prebindgen/covertest/model/Hold$For")
-                .map_err(|e| <__JniErr as ::core::convert::From<
-                    String,
-                >>::from(
-                    format!(
-                        concat!("Hold", ": instanceof ",
-                        "io/prebindgen/covertest/model/Hold$For", ": {}"), e
-                    ),
-                ))?
-            {
+            let __tag = (|| -> ::core::result::Result<i32, __JniErr> {
+                if env
+                    .is_instance_of(v, "io/prebindgen/covertest/model/Hold$Indefinite")
+                    .map_err(|e| <__JniErr as ::core::convert::From<
+                        String,
+                    >>::from(
+                        format!(
+                            "Hold: instanceof io/prebindgen/covertest/model/Hold$Indefinite: {}",
+                            e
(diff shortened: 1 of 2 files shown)


milyin:
Reviewed `3880ffbe` against `feature/registry-composed-source-walks`. Step 4 of #596.

## Reproduced

regeneration check, and `covertest-kotlin`'s 61 sections on the JVM.

I ran the instrument independently and get the same answer: 412 functions, no
converter-call sequence differing, and exactly two JNI operation sequences —
`Lookup` and `Reading`, the sums with payload arms.

One note for anyone repeating it: widening the pattern to include string
literals reports four, because `Hold` and `Marker` also change. That is the
instanceof message moving from `concat!("Lookup", ": instanceof ", …)` to a
single literal, which is the same text at runtime. Matching on the JNI
operations alone is what isolates the reordering.

## The claim this rests on is true

"Executed order is unchanged" carries the whole behavioural argument, so I read
the generated tag block rather than taking it. Each test returns:

```rust
let __tag = (|| -> Result<i32, __JniErr> {
    if env.is_instance_of(v, "…Lookup$Absent")…? { return Ok(0i32); }
    if env.is_instance_of(v, "…Lookup$Found")…?  { return Ok(1i32); }
    if env.is_instance_of(v, "…Lookup$Failed")…? { return Ok(2i32); }
    Ok(-1i32)
})()?;
```

So a `Found` value never reaches the `Failed` test, exactly as the old `if`
chain never reached it. The reordering is textual and the JNI call count per
execution is unchanged — which is what matters here, since `is_instance_of` is
a JNI round trip and this shape sits on the hot path for every sum crossing.

## Failure messages

All three survive, not only the two the description names. The null and
no-variant messages are byte-identical literals, and the instanceof message is
the same text spelled as one literal instead of a `concat!` of four fragments.

## The deleted `flat::Field` is covered where it matters

Dropping `JObjectSumFieldPlan`'s own `flat::Field` means the registry binds
payloads from the alternative's fields, and the shape that binding has to get
right is a **named-field variant** — the one case where a tuple-style bind
would be wrong. `Reading::Range { low, high }` is exactly that, and the
construct side builds it as a named-field literal:

```rust
2i32 => perftest_flat::Reading::Range { … }
```

Unit (`Missing`, `Absent`, `Indefinite`), tuple (`Exact`, `Found`, `Ranked`)
and named-field (`Range`) alternatives are all constructed through the composed
`Choice` and all run on the JVM, so `shape_alternative` is exercised across the
three shapes rather than only the common one.

## On the arms sharing the struct's bridge

Each arm being a `Product` over the same `JObjectBridge` is the part that makes
this step small, and it is the payoff from #598 and #599 rather than something
this PR had to build. A payload property now has one description for both
decodes, which is what makes "the properties again" no longer a risk here.

Nothing outstanding.

— Claude Code (Opus 5)

Issue #602, opened 2026-08-27
A data class is decomposed twice, with different coverage


milyin:
## What is different

A declared `data_class` is decomposed into leaves twice, by two derivations
with different coverage.

- **`Declarations::struct_out_wires_at`** (`prebindgen-jni/src/jni/compile.rs`)
  produces the leaves the registry receives as a `ValueDecon`. Those drive
  fixed-builder delivery: a declared function returning the struct, and an
  `impl Fn(T)` callback argument, hand the flattened leaves to a foreign
  builder which reassembles the value through `fromParts`.
- **`StructPlan`** (`prebindgen-jni/src/jni/struct_plan.rs`) produces the
  leaves the whole-object output encode emits, and the ones Kotlin's
  `fromParts` factory declares.

They agree wherever both exist — measured across the covertest and perftest
examples: 17 structs have both, and their leaf counts match. But the first
refuses shapes the second supports:

```rust
// struct_out_wires_at
if field.ty.optional_inner().is_some() || matches!(field.ty.kind(), TypeKind::Vec(_)) {
    return None;   // a nested data_class behind Option or Vec
}
```

The encode supports an optional nested class as a `present` flag plus a
defaulted group, and a sum field as a tag plus one group per variant.

## Why it matters

Two consequences, one today and one for #596.

A struct with an optional nested `data_class` gets **no fixed-builder
delivery**: returning it from a declared function, or passing it to an
`impl Fn(T)` callback, falls back to the whole-object path rather than
delivering leaves. That is a coverage gap in delivery, not in the encode.

And #596 — which composes JniGen's source-value walks in the registry — could
not compose this encode for the same reason. Its step 5 records the decision:
rendering the encode from the decomposition would require teaching the
decomposition the gated shapes, which changes what a foreign builder can be
delivered. That is a feature, not a refactoring, so it was left out of an
umbrella whose restriction is that behaviour stays semantically unchanged.

`the_decomposition_refuses_what_the_whole_object_encode_supports`
(`prebindgen-jni/src/jni/tests/flatten.rs`) pins the refusal, so the divergence
cannot narrow silently.

## What to do

Teach the decomposition the two gated shapes, so a struct with an optional
nested class or a sum field decomposes like every other one:

- an optional nested class contributes a `present` selector leaf followed by
  its group, with the group defaulted when absent;
- a sum field contributes a tag leaf followed by one group per variant.

The registry already has the vocabulary — `UnfoldLeaf::group`, and the
[...]

PR #603, opened 2026-08-27
Record why the whole-object encode is not composed

milyin:
## The decision

Step 5 of #596 was written as a decision rather than a promise: compose the
whole-object struct output through the registry's decomposition walk, or record
why it stays where it is. The code answers it, and not the way the umbrella
guessed.

**The encode is not a second implementation of the delivery walk.** It covers
shapes the registry-facing decomposition refuses:

```rust
// Declarations::struct_out_wires_at — the leaves the registry receives
if field.ty.optional_inner().is_some() || matches!(field.ty.kind(), TypeKind::Vec(_)) {
    return None;   // a nested data_class behind Option or Vec
}
```

The encode supports both: an optional nested class becomes a `present` flag
plus a defaulted group, a sum field a tag plus one group per variant.

So composing it means teaching the decomposition those gated shapes — and the
same decomposition drives fixed-builder returns and `impl Fn(T)` callback
arguments, so that would change what a foreign builder can be delivered. A
feature, not a refactoring, and this umbrella's own restriction is that
behaviour stays semantically unchanged unless a step calls the change out.

## What this PR does

Writes that down where the next reader looks, and pins it.

- `struct_out_wires_at`'s refusal and `encode_plan` each name the other and the
  divergence, so neither reads as an oversight.
- `the_decomposition_refuses_what_the_whole_object_encode_supports` holds it: a
  struct with an `Option<Leaf>` field decomposes to `None` while its
  whole-object encode still emits the `present` flag. Accepting the optional
  nested case in the decomposition fails it — checked.
- **#602** proposes the unification as the feature it is, with the coverage gap
  it would also close: a struct with an optional nested class gets no
  fixed-builder delivery today, so returning it or passing it to a callback
  falls back to the whole-object path.

## What I measured before deciding

Across both examples: 28 structs have a whole-object encode, 17 also have a
decomposition, and for all 17 the leaf counts agree. The two derivations are
consistent where they overlap — the difference is only in what they accept,
which is what makes this a coverage question rather than a duplication one.

## Verification

generated artifact is unchanged: this step adds a test and two explanations.

Step 5 of #596 — the one box that closes with a reason rather than a deletion.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -4233,8 +4233,10 @@ impl Declarations {
                 // optional nested class becomes a `present` flag plus a
                 // defaulted group, and a sum field a tag plus one group per
                 // variant. Delivering those to a foreign builder is a feature
-                // this decomposition does not have, so the two are not two
-                // implementations of one thing. See #596 step 5, and #602.
+                // this decomposition does not have: the encode is a superset
+                // rather than a second implementation. Where both apply they
+                // name the same leaves, which `assert_leaf_derivations_agree`
+                // checks. See #596 step 5, and #602.
                 crate::jni::classify::TypeKind::DataStruct {
                     st: _,
                     cfg: Some(_),
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -4226,6 +4226,15 @@ impl Declarations {
                 // A nested `data_class` inlines when it is reached directly.
                 // Behind an `Option` or a `Vec` there is no chain to reach
                 // through, so the whole value stays object-shaped.
+                //
+                // This refusal is why the whole-object output encode
+                // (`emit/struct_out.rs`) is not rendered from this
+                // decomposition: it supports the shapes refused here — an
+                // optional nested class becomes a `present` flag plus a
+                // defaulted group, and a sum field a tag plus one group per
+                // variant. Delivering those to a foreign builder is a feature
+                // this decomposition does not have, so the two are not two
+                // implementations of one thing. See #596 step 5, and #602.
                 crate::jni::classify::TypeKind::DataStruct {
                     st: _,
                     cfg: Some(_),
--- a/prebindgen-jni/src/jni/emit/struct_out.rs
+++ b/prebindgen-jni/src/jni/emit/struct_out.rs
@@ -118,6 +118,28 @@ pub(crate) fn synth_value_struct_leaves(
 /// mean teaching that decomposition the gated shapes — which would also change
 /// what a foreign builder can be delivered, a feature rather than a
 /// refactoring. #596 step 5 records that decision; #602 proposes the feature.
+///
+/// Where both derivations do apply, they must name the same leaves in the same
+/// order: `JniGen::assert_leaf_derivations_agree` checks that on every binding
+/// a test writes, since that agreement is what #602 would rest on.
+/// The leaf names this encode emits, flattened — the `fromParts` argument
+/// list, in order.
+///
+/// Test support: the registry-facing decomposition of the same struct must
+/// agree with this wherever it exists, and #603 recorded that agreement as
+/// measured rather than checked. `JniGen::write_rust` checks it now.
+#[cfg(test)]
+pub(crate) fn encode_leaf_names(
+    plan: &StructPlan,
+    emit: &prebindgen_registry::RustWriter,
+) -> Vec<String> {
+    let (_, slots) = encode_plan(plan, &quote!(v), "", 0, &quote!(env), emit);
+    slots
+        .iter()
+        .map(|slot| slot.ident.to_string().trim_start_matches('_').to_string())
+        .collect()
+}
+
 fn encode_plan(
     plan: &StructPlan,
     access: &TokenStream,
--- a/prebindgen-jni/src/jni/emit/struct_out.rs
+++ b/prebindgen-jni/src/jni/emit/struct_out.rs
@@ -105,6 +105,19 @@ pub(crate) fn synth_value_struct_leaves(
 /// an [`EncSlot`] describing its `JValue` slot. `access` is the Rust
 /// expression yielding the current struct value (`v`, `v.field`, or the
 /// matched `__cN` under an Option); `prefix` namespaces the generated idents.
+/// Walk the struct's fields and encode each leaf into a `fromParts` argument.
+///
+/// This is a source-value walk that the registry does not perform, and the
+/// reason is coverage rather than ownership. The registry-facing decomposition
+/// of a struct (`Declarations::struct_out_wires_at`) refuses a nested
+/// `data_class` behind an `Option` or a `Vec`; this encode supports both, an
+/// optional nested class as a `present` flag plus a defaulted group and a sum
+/// field as a tag plus one group per variant.
+///
+/// Rendering this through the registry's decomposition walk would therefore
+/// mean teaching that decomposition the gated shapes — which would also change
+/// what a foreign builder can be delivered, a feature rather than a
+/// refactoring. #596 step 5 records that decision; #602 proposes the feature.
 fn encode_plan(
     plan: &StructPlan,
     access: &TokenStream,
--- a/prebindgen-jni/src/jni/emit/struct_out.rs
+++ b/prebindgen-jni/src/jni/emit/struct_out.rs
@@ -41,6 +41,17 @@ pub(crate) fn handle_field_fqn(ext: &Declarations, h: &Projection) -> String {
 /// an absent `Option<nested>` parent.
 pub(crate) struct EncSlot {
     ident: proc_macro2::Ident,
+    /// How deeply nested the field this slot came from is: 0 for the struct's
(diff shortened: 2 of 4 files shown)


milyin:
Reviewed `43580f8c` against `feature/registry-composed-source-walks`. Step 5 of #596.

This is the box that closes with a reason, so the reason is what I checked.

## Reproduced

Workspace tests (830), strict Clippy, `cargo fmt --check`, rustdoc, and the
regeneration check. Generated output unchanged, as a test-and-comments step
should leave it.

## The load-bearing claim holds

Everything rests on "the same decomposition drives fixed-builder returns and
`impl Fn(T)` callback arguments", because that is what makes extending it a
feature rather than a refactoring. It is true, and the code says so directly —
`synth_value_struct_leaves` is documented as producing the leaves "as the
fixed-builder output and callback paths want them", and it produces them by
mapping `struct_out_wires_of`. Widen the refusal and those paths change what
they deliver.

## The refusal is not an edge case

Instrumenting `synth_value_struct_leaves` across both examples: of 29 structs
that reach it, **11 are refused** and 18 decompose. Nearly two in five.

That cuts toward the decision rather than against it — a coverage difference
that large is a feature question, not a tidy-up — but it also says #602 is
worth more than its filing suggests. Every one of those 11 keeps the
whole-object `fromParts` path today, and #602's summary of the gap
(returning such a struct, or passing it to a callback, falls back) applies to
all of them.

## The new test bites

Accepting the optional nested case in `struct_out_wires_at` — deleting the
`optional_inner()` half of the refusal — fails
`the_decomposition_refuses_what_the_whole_object_encode_supports` with the
message it was written to give. The test checks both halves, the refusal and
the encode's `__inner_present`, which is what makes it a record of the
divergence rather than of one side of it.

## What the decision still leaves unchecked

The argument depends on the two derivations agreeing wherever they overlap, and
the PR measured that — 17 structs, leaf counts agree. That measurement is not
pinned by anything.

It is also the property the code already knew it was missing.
`synth_value_struct_leaves`' own doc comment, unchanged by this PR, says the
leaf names reach the `fromParts` parameters and the recipe identically —

> and two walks agreeing was a property nothing checked.

So a comment has flagged this as unchecked, a decision has now been recorded on
top of it, and it is still unchecked. A standing test comparing the two leaf
lists for every struct that has both is also what would tell you when #602's
unification is safe to attempt — the agreement is its precondition.

One practical note if that test is written: the two are not both available at
the same moment. `synth_value_struct_leaves` runs before `resolve`, and
`struct_plan` returns `None` there — I confirmed that by probe, getting zero
plan fields for all 18 decomposed structs while the decomposition side answered
normally. The comparison has to happen after resolve.

## On one phrase

"The two are not two implementations of one thing" is the right conclusion for
the whole and slightly strong for the overlap: for the 18 structs that have
both, they are two derivations of the same leaves, which is exactly why their
agreement had to be measured. "A superset" carries the same decision without
inviting the next reader to wonder about the 18.

Nothing here blocks the PR. Closing this box with a written reason, a test, and
a filed feature is a better outcome than composing it would have been.

— Claude Code (Opus 5)

milyin:
Both done in `370d9edd`.

**The agreement is checked now, not measured.** `JniGen::assert_leaf_derivations_agree` runs on every binding a test writes — the same place #592 put the edge-completeness check — comparing the whole-object encode's flattened leaves against the decomposition's for every declared data class that has both. Dropping a leaf from either side fails 41 tests.

Two things surfaced while writing it, both of which sharpen the record rather than change the decision:

- The two name a leaf for **different audiences**: the decomposition uses the Kotlin property name (`maybeLong`), the encode the Rust field (`maybe_long`), and a keyword-colliding name keeps a trailing underscore on one side only. So the check compares which leaves in which order, and drops the spelling. That difference is itself worth knowing — it means the two were never textually comparable, which is part of why nobody had compared them.
- Your note about ordering was exactly right and cost me one failed run: before resolve the frozen plan has no entry, and for anything that is not a declared data class the lookup is panic-backed by design, so the check runs after resolve and skips those.

**"A superset"** replaces "not two implementations of one thing", in the code comment as well as here. You are right that the phrase overreached for the 18 that overlap — and the check I just added is the thing that makes the overlap a stated property rather than an observation.

**On the 11 refusals of 29** — that is a bigger share than I had implied, and I have added it to #602, since it changes how that issue reads: not an edge case worth unifying for tidiness, but a coverage gap affecting nearly two in five declared data classes.

— Claude Code (Opus 5)

milyin:
Reviewed `370d9edd`. The standing check landed, in the right place — after
resolve, in `write_rust`, under `#[cfg(test)]`, beside
`assert_edges_cover_rendered_calls`. Validation still green (830 tests, Clippy,
fmt, regeneration check), and the "superset" rewording reads better than what
it replaced.

**It bites.** Reversing the encode's leaf order fails 10 JNI tests, naming the
struct:

```
`ZId`: the whole-object encode and the registry-facing decomposition name
different leaves
```

**But it only sees flat structs.** Instrumenting the assertion: 28 comparisons
run across the suite, the largest has **5 leaves, and not one has a nested
leaf name** — no `__` anywhere in the compared sets.

The agreement this is meant to hold is the one #603 measured over the
*examples*, and that is where the nesting lives: `BlobValue`'s
`stamp__secs`/`stamp__nanos`, `ObjectBoundary64`'s 64 leaves of
`left__right__…__value`, `ObjectBoundary63`'s five nested groups. Those never
reach the check, because `#[cfg(test)]` fires for this crate's own tests and
not for an example's build script.

That leaves the interesting half uncovered — not *which* leaves a flat struct
has, but how the two derivations flatten a nested path and in what order. The
gap is visible directly: making the encode spell its nesting separator
differently (`__` → `X`) fails **nothing**, because no compared struct has one
to change.

A fixture with a nested `data_class` in the JNI tests closes it, and there are
already nested `Inner` structs in `tests/values.rs` and `tests/sealed.rs` to
model one on. Worth doing here rather than later: a check that runs 28 times
and proves the easy case is the kind that reads as covered.

One smaller thing while that fixture is written: `plain()` deletes every
underscore, so `a__b` and `ab` compare equal. That is deliberate — the two
sides spell in camelCase and snake_case — but the `__` separator is structure
rather than spelling, and it is the part a nested comparison would be about.
Normalizing case and the camel/snake boundary while keeping the separator would
say the same thing about spelling and more about shape.

— Claude Code (Opus 5)

milyin:
Fixed in the head commit — you were right that a check running 28 times over flat structs reads as covered while proving the easy case.

**The nesting is reached now.** `the_two_derivations_agree_on_a_nested_data_class` builds `Page` → `Entry` → `Stamp`, so the comparison sees depths 0, 1 and 2 and the decomposition's `head__stamp__secs`. Writing that binding is what runs the standing check over it.

**On the separator.** I took your point about structure versus spelling, but not by keeping the separator in the name comparison — the encode's separator is `_`, which is ambiguous with underscores inside field names (`maybe_long` would split into two segments while `maybeLong` stays one). So nesting is compared as a number instead: the decomposition's `__` count against the depth the encode already tracks while recursing, which `EncSlot` now carries. That says more about shape than the separator would, and it cannot be confused by a field name.

Both halves are mutation-checked: flattening one level shallower fails the nesting half, reversing the encode's field order fails the name half. Your separator mutation (`__` → `X`) still passes, and now deliberately — on the encode side that string is a local-variable name, and the check is about which leaves at which depth, not what the encode calls its locals.

— Claude Code (Opus 5)

milyin:
Reviewed `96eb7823`. This goes past what I asked for, and in a better direction.

I suggested keeping the `__` separator in `plain()` — the name normalizer the
agreement check compares with — so that spelling differences in nesting would
show up. You instead added nesting **depth** as a second compared property:
`encode_leaves` now returns `(name, depth)` per leaf, and
`assert_leaf_derivations_agree` checks the decomposition's `__` count against
it. That is the right call. Depth is the structural fact the separator happens
to encode; comparing it directly states the property instead of inferring it
from how a name is spelled, and it survives any later change to the naming
scheme.

## Reproduced

**The check now sees nesting.** Instrumenting the assertion: 31 structs
compared, up from 28, and two of them carry depth above zero —

```
PROBE Entry names=["stampsecs","stampnanos","weight"]          depth=[1,1,0]
PROBE Page  names=["headstampsecs","headstampnanos","headweight","count"] depth=[2,2,1,0]
```

So the fixture chain `Stamp` -> `Entry` -> `Page` exercises two levels of
inlining, not one, and each level is compared at both spellings and depth.

**The depth assertion bites.** Flattening `encode_leaves`' reported depth to a
constant `0` fails two tests with

```
assertion `left == right` failed: `Entry`: the whole-object encode and the
registry-facing decomposition flatten nesting differently
```

Before this commit the same mutation passed everything, because no compared
struct had a nested leaf.

**Validation green**, unchanged output: 831 tests pass, 0 clippy warnings,
`cargo fmt --check` clean, `PASS - regen check clean`.

The doc comment on the new fixture recording *why* it exists — that every
struct the check compared before it was flat — is what stops the next person
from deleting it as redundant with the other flatten tests.

Step 5 now closes with a written reason, a divergence test, a standing
agreement check over both names and nesting, and #602 filed for the feature
itself. No further findings from me.

— Claude Code (Opus 5)

PR #604, opened 2026-08-27
Let a representation domain carry its scalar kind

### LINKED ISSUE #589: Representation domains decide type identity by spelling

milyin:
Step 6 of #596, the umbrella for registry-composed source walks. Closes
#589.

A **representation domain** is the set of legal values of the scalar a custom
conversion crosses in — its **representation** — and it is what
`convert!(Duration).valid_range(0u64..=1_000_000)` declares. Three questions are answered from that scalar: whether a generated
bounds check needs a NaN test, which candidate values a **niche** — a value
outside the domain, reserved so that `Option<Duration>` needs no second field —
is picked from, and whether the domain and the conversion's representation are
the same type.

`RepresentationDomain` kept the scalar as a `syn::Type` and recovered it by
matching the rendered token string, so a spelling decided all three: its
`contains_expr` matched `"f32" | "f64"` for the NaN test, its `niche_values`
matched ten integer names and ended `_ => unreachable!()`, and each adapter
compared the domain's spelling against the representation's. `DomainScalar` —
the trait a Rust scalar implements to be usable as a domain's scalar — is
publicly re-exported and unsealed, so a downstream implementation was free to
spell `u64` as `::core::primitive::u64` and take the wrong branch in two of
those three, and fail the build in the third.

## What changed

`DomainKind` — twelve variants, `i8` through `u128` and the two floats — is
what `DomainScalar` now yields (`domain_kind()` replaces `domain_type()`) and
what `RepresentationDomain` stores. The spelling is derived from the kind
(`DomainKind::as_str`, `DomainKind::ty`) and never the reverse, so a domain has
no spelling left for anyone to choose. The variants, their spellings and the
reduction from a written type are all generated from one `domain_kinds!` list,
so a kind cannot enter the set while any of the three is left unanswered.

All three questions are decided on the kind. Both string matches and the
`unreachable!()` are gone: the NaN test asks `DomainKind::is_float`, and
`niche_values` matches the twelve kinds exhaustively. `DomainKind::from_type`
reduces a written scalar type to its kind off the last path segment, and
Cbindgen — the C adapter — compares that against the domain's kind, so a
representation written `::core::primitive::u64` and a domain over `u64` are one
type where they were two.

JniGen — the JNI/Kotlin adapter — keeps its comparison of model type keys, and
that comparison is exact for a reason upstream of it rather than because a key
is canonical: a `TypeKey` reduces constructors, not scalars, so two spellings of
one scalar are two keys. What JniGen is protected by is that a path-qualified
scalar cannot reach the comparison at all — a marked item lives in one flat
namespace of bare names, so such a representation fails to resolve first. The
comment there now says that, and
`a_qualified_scalar_representation_never_reaches_the_domain_check` pins it.

Two more spelling decisions go with it. `is_scalar`, which Cbindgen asks one
statement before the domain check, decided whether a written type is an
FFI-safe scalar by matching thirteen hand-written names; it now asks
`ScalarKind::from_type`, the same closed set the model already keeps, so the
scalars are spelled out in one place instead of three. JniGen asked whether a domain is the
`u64` one whose niches fit in a `jlong`, the 64-bit JVM integer its wire values
cross in, and asked it as
`TypeKey::from_type(domain.ty()).as_str() != "u64"`. It now asks
`domain.kind() != DomainKind::U64`.

**The design question #589 left open** — widen the flat model's `ScalarKind`,
its classification of the scalars a source may write, or introduce a
domain-side kind — is settled as a domain-side kind. Widening `ScalarKind` with
`I128` and `U128` would make the model classify scalars no adapter can pass
across an FFI boundary, and `ScalarKind` also carries `Bool`, `Isize` and
`Usize`, which no domain can be declared over — so sharing one enum would have
kept an `unreachable!()`, at the other end.

## Tests

Three, each mutation-checked:

- `a_domain_matches_its_representation_however_the_representation_is_spelled`
  (Cbindgen) builds the binding #589 reports as failing: a conversion whose
  representation is written `::core::primitive::u64` under a `u64` domain.
  Restoring the spelling comparison fails it with the issue's own message.
- `one_scalar_is_one_kind_however_it_is_written` (registry) pins the reduction
  itself, including that `bool` has no domain kind. Matching the whole rendered
  type instead of its last segment fails it.
- `only_a_float_range_guards_against_nan` (registry) covers the branch this
  rewrites, which had no test of its own: making `is_float` answer `false` —
  deleting the NaN guard from every generated float bounds check — passed the
  whole workspace before this test existed.
- `a_scalar_kind_is_recovered_from_any_spelling_of_its_type` (flat) pins the
  set `is_scalar` now defers to. Matching the whole rendered type instead of its
  last segment fails it, and fails the Cbindgen test above with it.
- `a_qualified_scalar_representation_never_reaches_the_domain_check` (JniGen)
  pins the resolution failure described above, and that it is not the domain
  check reporting it.

## Verification

836 tests pass, 0 clippy warnings (strict), `cargo fmt --check` clean,
`RUSTDOCFLAGS='-D warnings' cargo doc` clean, `examples/regen-check.sh` reports
`PASS - regen check clean`. Generated output is byte-identical: no adapter's
output depended on how a domain was spelled, only on which scalar it named.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-c/src/convert.rs
+++ b/prebindgen-c/src/convert.rs
@@ -74,9 +74,12 @@ impl CbindgenBuilder {
             "Cbindgen custom representations must be C scalar types"
         );
         if let Some(domain) = decl.domain() {
+            // Both sides are reduced to their kind first: a representation and
+            // a domain naming one scalar are the same type however either was
+            // spelled.
             assert_eq!(
-                TypeKey::from_type(domain.ty()),
-                TypeKey::from_type(&repr),
+                Some(domain.kind()),
+                prebindgen_registry::DomainKind::from_type(&repr),
                 "Cbindgen conversion domain type does not match its {} representation",
                 match direction {
                     Direction::Construct => "input",
--- a/prebindgen-c/src/lib.rs
+++ b/prebindgen-c/src/lib.rs
@@ -103,7 +103,6 @@ use std::collections::{HashMap, HashSet};
 // `TypeRef` now, so what is left here serves the two node populations that
 // remain — a build-script declaration, and a converter's own generated
 // signature.
-pub(crate) use prebindgen_registry::types_util::path_tail_ident as type_path_tail;
 use prebindgen_registry::{
     decl::{ConvertDecl, ConvertSpec},
     flat::{extract_fn_trait_args, Field, Origin, ScalarKind, TypeKind, TypeRef},
@@ -525,7 +524,7 @@ fn sanitize(key: &TypeKey) -> String {
 /// The short name a C symbol is built from — the key's last path segment, or a
 /// sanitized rendering of the whole key when it is not a path.
 ///
-/// Off the **identity**: `type_path_tail` took the last segment of a node, and
+/// Off the **identity**: this took the last path segment of a node, and
 /// `TypeKey::short_name` is the same question asked of the canonical string, so
 /// the name no longer depends on holding the node it was derived from.
 fn type_short(key: &TypeKey) -> String {
@@ -697,27 +696,13 @@ fn bool_out_expr(value: TokenStream) -> TokenStream {
 
 /// Whether `ty` is an FFI-safe scalar primitive that passes through unchanged
 /// (`bool`, the fixed-width / pointer-width integers, and floats).
+///
+/// The set is `ScalarKind`'s, asked of the model rather than spelled out again
+/// here — the same reason [`r_is_scalar`] asks the classification. This one
+/// takes a written type because its caller has one: a `from!` / `into!`
+/// representation is a type the binding wrote, not a lowered `TypeRef`.
 fn is_scalar(ty: &syn::Type) -> bool {
-    type_path_tail(ty)
-        .map(|i| {
-            matches!(
-                i.to_string().as_str(),
-                "bool"
-                    | "i8"
-                    | "i16"
-                    | "i32"
-                    | "i64"
-                    | "isize"
-                    | "u8"
-                    | "u16"
-                    | "u32"
-                    | "u64"
-                    | "usize"
-                    | "f32"
-                    | "f64"
-            )
-        })
-        .unwrap_or(false)
+    ScalarKind::from_type(ty).is_some()
 }
 
 /// The element of a shared slice borrow (`&[E]`), off the classification.
--- a/prebindgen-c/src/tests/boundary_invariants.rs
+++ b/prebindgen-c/src/tests/boundary_invariants.rs
@@ -230,7 +230,7 @@ fn restricted_inbound(
     if let syn::Type::Reference(r) = ty {
         return restricted_inbound(&r.elem, structs, enums, seen);
     }
-    let tail = type_path_tail(ty)?.to_string();
+    let tail = prebindgen_registry::types_util::path_tail_ident(ty)?.to_string();
     if tail == "MaybeUninit" {
         return None;
     }
--- a/prebindgen-c/src/tests/lowering.rs
+++ b/prebindgen-c/src/tests/lowering.rs
@@ -213,6 +213,41 @@ fn trait_backed_custom_conversion_renders_from_the_late_plan() {
     );
 }
 
+/// `::core::primitive::u64` and `u64` name one scalar, and a domain declared
+/// over that scalar matches a representation written either way. Before the
+/// domain carried its kind, this fixture failed to build: the check compared
+/// the two spellings and reported "domain type does not match its input
+/// representation" for a type that does match.
+#[test]
+fn a_domain_matches_its_representation_however_the_representation_is_spelled() {
+    let loc = SourceLocation::default();
+    let items: Vec<(syn::Item, SourceLocation)> =
+        ["pub fn duration_echo(v: Duration) -> Duration { unimplemented!() }"]
+            .into_iter()
+            .map(|source| (syn::parse_str(source).unwrap(), loc.clone()))
+            .collect();
+    let registry = crate::test_util::reg_from_items(declare_referenced(items)).unwrap();
+    let cbindgen = CbindgenBuilder::new()
+        .source_module(syn::parse_quote!(myflat))
+        .convert(
+            prebindgen_registry::convert!(Duration)
+                .input(prebindgen_registry::from!(::core::primitive::u64))
+                .output(prebindgen_registry::into!(::core::primitive::u64))
+                .valid_range(0u64..=1_000_000u64),
+        )
+        .base_name("z_duration")
+        .function(syn::parse_quote!(duration_echo))
+        .panic();
+
+    let src = write(cbindgen, registry, "qualified_representation");
+    let compact: String = src.split_whitespace().collect();
+    assert!(
+        compact.contains("extern\"C\"fnduration_echo(v:::core::primitive::u64"),
+        "{src}"
+    );
+    assert!(compact.contains("(v)<=1000000u64"), "{src}");
+}
+
 #[test]
 fn output_terminals_stay_unrendered_until_final_write() {
     let loc = SourceLocation::default();
--- a/prebindgen-flat/src/flat/tests/acceptance.rs
+++ b/prebindgen-flat/src/flat/tests/acceptance.rs
@@ -2037,3 +2037,23 @@ fn a_raw_identifier_survives_typeid() {
     };
     assert!(qualified.ident().is_none());
 }
+
+/// `ScalarKind` is the closed set of the scalars a source may write, and it
+/// answers for a written type however that type is spelled — so a caller
+/// asking "is this an FFI-safe scalar primitive" needs no name table of its
+/// own.
+#[test]
+fn a_scalar_kind_is_recovered_from_any_spelling_of_its_type() {
+    use crate::flat::ScalarKind;
+
+    assert_eq!(
+        ScalarKind::from_type(&syn::parse_quote!(usize)),
+        Some(ScalarKind::Usize)
+    );
+    assert_eq!(
+        ScalarKind::from_type(&syn::parse_quote!(::core::primitive::usize)),
+        Some(ScalarKind::Usize)
+    );
+    assert_eq!(ScalarKind::from_type(&syn::parse_quote!(String)), None);
+    assert_eq!(ScalarKind::from_type(&syn::parse_quote!(&u8)), None);
+}
--- a/prebindgen-flat/src/flat/ty.rs
+++ b/prebindgen-flat/src/flat/ty.rs
@@ -1124,6 +1124,16 @@ impl ScalarKind {
         })
     }
 
+    /// The kind a written scalar type names, off its last path segment.
+    /// `None` for anything that is not one of these scalars.
+    ///
+    /// The closed set below is the whole answer to "is this an FFI-safe scalar
+    /// primitive", so a caller asking that question asks it here rather than
+    /// keeping a name table of its own.
+    pub fn from_type(ty: &syn::Type) -> Option<Self> {
+        Self::from_name(&crate::types_util::path_tail_ident(ty)?.to_string())
+    }
+
     /// The Rust spelling — the identity this was lowered from.
     pub fn as_str(self) -> &'static str {
         match self {
--- a/prebindgen-jni/src/jni/builder.rs
+++ b/prebindgen-jni/src/jni/builder.rs
@@ -1625,7 +1625,7 @@ impl Declarations {
         else {
             return (Niches::empty(), Vec::new());
         };
-        if TypeKey::from_type(domain.ty()).as_str() != "u64"
+        if domain.kind() != prebindgen_registry::DomainKind::U64
             || prebindgen_registry::types_util::path_tail_ident(wire)
                 .is_none_or(|ident| ident != "jlong")
         {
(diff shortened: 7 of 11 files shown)


milyin:
Reviewed `d8038fcb` against `feature/registry-composed-source-walks`. Step 6 of
#596, and the fix for #589 — the finding that a representation domain recovered
its scalar by matching a rendered type string.

Nothing here blocks. Three things below: one justification that does not survive
a probe and now stands as a comment in the source, one decision the compiler
does not check, and one scope question against the umbrella's own criterion.

## Reproduced

834 tests pass, 0 clippy warnings under `-D warnings`, `cargo fmt --check`
clean, `RUSTDOCFLAGS='-D warnings' cargo doc` clean, `PASS - regen check clean`.

All three new tests bite, each against its own mutation:

- Restoring the C adapter's spelling comparison fails
  `a_domain_matches_its_representation_however_the_representation_is_spelled`
  with #589's own message, both spellings visible:
  ```
  Cbindgen conversion domain type does not match its input representation
    left: TypeKey("u64")
   right: TypeKey(":: core :: primitive :: u64")
  ```
- Matching the whole rendered type in `DomainKind::from_type`, rather than its
  last path segment, fails `one_scalar_is_one_kind_however_it_is_written`.
- Making `DomainKind::is_float` answer `false` fails
  `only_a_float_range_guards_against_nan` — **and nothing else in the
  workspace.** That is the most valuable thing in this PR. Deleting the NaN
  guard from every generated float bounds check was invisible to the other 833
  tests, and now it is not.

The design argument holds as stated. `ScalarKind` — the flat model's
classification of source scalars — is thirteen variants with no `I128` or
`U128`, and it carries `Bool`, `Isize` and `Usize`, so sharing one enum would
have moved the `unreachable!()` rather than removed it.

`builder.rs:1630` is a second real fix the body does not claim as one. The line
it replaces, `TypeKey::from_type(domain.ty()).as_str() != "u64"`, is exactly
what `TypeKey::as_str`'s own doc forbids: "do not parse it or compare it with
Rust spellings to classify a type."

## 1. The reason given for leaving JniGen alone is not the reason JniGen is safe

The body says JniGen keeps its key comparison because "its representation side
is a key the flat model built, canonical already". `TypeKey` is that key — the
model's canonical type identity, built by reducing a written type against
`Normalization::PRELUDE`. That prelude is seven constructor entries — `Vec`,
`Option`, `Result`, `String`, `Box`, `MaybeUninit`, `Cow` — and its own doc says
the scalars are not among them. A probe:

```
TypeKey::from_type(::core::primitive::u64).as_str() == ":: core :: primitive :: u64"
TypeKey::from_type(u64).as_str()                    == "u64"
```

The two keys differ, so the comment now standing in `compile.rs` — "this
compares two canonical names" — is false on the representation side. It also
describes a change that did not happen on that line: `DomainScalar::domain_type`
returned `syn::parse_quote!($t)` from the same macro arm, so the domain side
already spelled `u64` bare before this PR. JniGen's comparison behaves exactly
as it did.

What keeps JniGen out of the mismatch is upstream of the comparison, and it
differs by how the conversion declares its representation. I built both:

- Declared with `fun!`, which names a captured source function: the flat model
  refuses that function outright — "names the type `::core::primitive::u64`,
  which the flat API does not declare — and being path-qualified it never could,
  because marked items live in one flat namespace of bare names."
- Declared with `from!` / `into!`, which name a type directly: resolution fails
  with `Unresolved { key: TypeKey(":: core :: primitive :: u64") }`.

That is a stronger guarantee than canonicalization, and it is the one worth
writing down, because it says a qualified scalar can never reach the comparison
at all. It also means #589's symptom as that issue states it — spelling one
integer `::core::primitive::u64` instead of `u64` fails the build — is still
true in JniGen. The build fails at a different check, with a message that never
mentions the domain. Whether that is the same defect is a judgment call I am
happy to lose; what does not survive is the account of why JniGen needs nothing.

Two things fix it: say in the comment what actually refuses the spelling, and
pin it, since no test asserts that a qualified scalar cannot reach the domain
comparison.

## 2. `from_type` is the one `DomainKind` decision the compiler does not check

Every other use of `DomainKind` is an exhaustive `match`: `as_str` and
`niche_values` both stop compiling if a variant is added. `from_type` instead
scans a hand-written twelve-element array.

Dropping `Self::I128` from that array passes the whole workspace, all 834 tests.
A domain over `i128` would then report "domain type does not match its
representation" against an `i128` representation that does match — the same
class of wrong answer #589 reports, arriving by omission instead of by spelling.

Generating both `as_str` and `from_type` from one macro list closes it: a
variant added without touching the list then fails `as_str`'s exhaustive match
at compile time.

## 3. The last spelling match on this path is three lines above the one that went

`custom_plan` (`prebindgen-c/src/convert.rs:73`) asserts `is_scalar(&repr)`
immediately before the domain check this PR rewrote. `is_scalar`
(`prebindgen-c/src/lib.rs:700`) decides whether a type is an FFI-safe scalar by
matching its last path segment against thirteen hand-written names — the same
question about the same value, one statement earlier, by the same means.

It is the only remaining caller of that function. The two callers in
`trait_impl.rs` already ask `r_is_scalar`, the model-backed version, whose doc
says the thing this step is about: "`ScalarKind` IS the closed set the name table
below was spelling out by hand."

The umbrella's success criterion is "No production code matches a rendered type
string to decide identity". After step 6, `custom_plan` still does.

I am not proposing `is_scalar` become `DomainKind::from_type(..).is_some()` —
the sets genuinely differ, since `is_scalar` accepts `bool`, `isize` and
`usize`, and no domain can be declared over any of them. The point is that there
are now three hand-maintained closed sets over the scalars (`ScalarKind`,
`DomainKind`, and `is_scalar`'s match) where the body's reasoning justifies two,
and the third is the ad-hoc one.

One consequence of the pair is already live. Both `is_scalar` and
`DomainKind::from_type` reduce off the last segment, so `from!(some::module::u64)`
— a path whose tail merely spells a scalar — now passes both checks where the
old key comparison rejected it. If that path names something other than `u64`,
the diagnostic moves from a stated assert into a rustc error inside generated
code. Marginal, and I would not change the reduction to avoid it, but it is what
the pair now permits.

## Also checked

`convert_target` (`builder.rs:1694`) parses a `TypeKey` back into a `syn::Type`,
which is this rule's other half. Its own doc names the leak and tracks it as
#558, so it is out of scope here rather than missed. The remaining
`key().as_str()` uses across both adapters are artifact labels, sort keys,
diagnostics and report fields — what that method is for.

— Claude Code (Opus 5)

milyin:
All three addressed in `884ec54d`.

**1. The comment was wrong, and so was the body.** You are right that a `TypeKey` reduces constructors and not scalars, so `::core::primitive::u64` and `u64` are two keys and JniGen's comparison behaves exactly as before. What protects it is upstream: a path-qualified scalar never reaches the check, because a marked item lives in one flat namespace of bare names. The comment in `compile.rs` says that now, and `a_qualified_scalar_representation_never_reaches_the_domain_check` pins it — a `from!(::core::primitive::u64)` representation under a `u64` domain fails with `Unresolved { key: TypeKey(":: core :: primitive :: u64") }`, and the test asserts the error names that key and does not mention the domain. The PR body carries the corrected account.

On whether that is still #589's symptom in JniGen: it is, in the sense that the build fails. It fails at resolution, with a message about an undeclared type rather than about a domain, and no branch is taken on a spelling. I would not claim more than that.

**2. `from_type` is generated now, and the mutation no longer compiles.** A `domain_kinds!` list produces the enum, `as_str` and `from_type` together. Dropping `(I128, "i128")` from it no longer passes 834 tests — it fails to build, because `impl_ints!` still names `DomainKind::I128`. The set cannot be entered while any of the three is left unanswered.

**3. `is_scalar` defers to the model.** It asks `ScalarKind::from_type` — new in flat, the last-segment reduction over `ScalarKind`'s own closed set — instead of matching thirteen names. The two sets were identical name for name, so nothing changes behaviourally; what goes is the third hand-maintained table. `a_scalar_kind_is_recovered_from_any_spelling_of_its_type` pins it, and dropping the last-segment reduction fails both it and the Cbindgen test. `type_path_tail` had no callers left in `prebindgen-c` outside test support, so the re-export is gone too.

Your observation about the pair stands and I left it as you would: `from!(some::module::u64)` passes both reductions, and if that path names something other than `u64` the failure lands in rustc rather than in an assert. Narrowing either reduction to full paths would cost the fix its point.

836 tests pass, 0 clippy warnings, `cargo fmt --check` clean, `RUSTDOCFLAGS='-D warnings' cargo doc` clean, `PASS - regen check clean`, generated output unchanged.

— Claude Code (Opus 5)

milyin:
Reviewed `884ec54d`. All three answered, and two of the three answers are
better than what I proposed.

## The kind list

I suggested generating `as_str` and `from_type` from one macro list. The commit
generates the **enum** from it too, so a variant cannot exist without a spelling
and a reduction — there is no list to fall out of sync with, because the list is
the only place a kind is written.

Verified: deleting `(I128, "i128")` is now a compile error, `E0599` at both
`niche_values` and the `DomainScalar` impl. The same omission passed all 834
tests on the previous head.

## The scalar set

`is_scalar` is now `ScalarKind::from_type(ty).is_some()`, and the reduction it
used to do by hand moved to the flat model as a public `ScalarKind::from_type`
with its own test. That is the right home: the caller keeps a written type
because a `from!` / `into!` representation is one, and the model answers which
scalar it names. The `type_path_tail` alias goes with it.

The sets match exactly — `ScalarKind::from_name` is the same thirteen names
`is_scalar` spelled out — and `PASS - regen check clean` confirms no C output
moved. Three hand-maintained closed sets over the scalars are now two, and the
two are the ones the body's design argument justifies.

The new coverage overlaps usefully: matching the whole rendered type in
`ScalarKind::from_type` fails both the new flat test and the C test
`a_domain_matches_its_representation_however_the_representation_is_spelled`,
since `is_scalar` gates the same declaration one line above the domain check.

## What protects JniGen

The comment says the true thing now — that a `TypeKey` is not reduced to a
canonical spelling, and that the comparison is exact for a different reason —
and `a_qualified_scalar_representation_never_reaches_the_domain_check` pins it.
Asserting that the error does **not** mention "domain type" is the right shape:
the property is which check fires first, so the negative half is the half that
holds the order in place.

The test is not vacuous — the error it inspects is
`Some(Resolve(Unresolved { entries: [UnresolvedEntry { key: TypeKey(":: core ::
primitive :: u64"), direction: Construct, .. }, ..] }))` — and both assertions
would fail if resolution ever accepted that key.

One small thing, only because precision in that comment is the point of the
change: the reason it gives — a marked item lives in one flat namespace of bare
names — is the reason the **`fun!` path** refuses, where the flat model rejects
the source function outright with `NotExpressible`. The test exercises the
`from!` path, where a declared representation key simply resolves to nothing.
Both really do fail before the domain check, so the conclusion stands; the
sentence merges two routes into one. Splitting it, or naming the route the test
takes, would make the comment match its own pin.

## Verification

836 tests pass, 0 clippy warnings under `-D warnings`, `cargo fmt --check`
clean, `RUSTDOCFLAGS='-D warnings' cargo doc` clean, `PASS - regen check clean`.

Nothing further from me on step 6.

— Claude Code (Opus 5)

milyin:
Comment split in the head commit: the `fun!` route (the flat model refuses the captured source function, because marked items live in one flat namespace of bare names) and the `from!` / `into!` route (the qualified key resolves to nothing), with the pin named against the route it actually takes. Comment-only — 836 tests, fmt clean.

Merging step 6.

— Claude Code (Opus 5)

PR #605, opened 2026-08-27
Delete the part read nobody speaks, and reconcile #513

milyin:
Step 7 of #596, the umbrella for registry-composed source walks, and its last.
The step has two halves: delete what the umbrella stranded, and state on #513 —
the umbrella above this one, which asked for one registry-driven Rust
generation pipeline — what its branch now is.

## The code half

`PartRead` is what a **bridge** — the adapter-supplied half of a composed
conversion — returns when the registry's `Product` composer asks it to read one
part out of an intermediate value. It had three constructors: `expression` (a
projection), `statements` (a prelude that binds but cannot fail) and `fallible`
(a prelude that can fail).

`statements` has no production caller. Step 2b reached for `fallible`, because
every JVM property read can fail, and steps 3 to 5 reached for neither. It is
deleted here, with its test fixture and the one test that separated it from
`fallible`, as #596 said it would be if nothing spoke it.

The `fallible` flag stays a fact of its own rather than being read off "has a
prelude", and the struct now says why: the two coincide across the remaining two
constructors, but a fallible read needing no statements — an expression carrying
its own `?` — would be silently demoted by the shorter rule.

## The #513 half

Nothing is posted to #513 until this PR is approved. The text below is what
goes there on merge: one comment, and the body edits it describes.

### Proposed comment

> ## Where this branch stands after #581 and #596
>
> **#581 made the writer a pure consumer.** `write_rust` takes one frozen
> `Assembly` — the artifacts, the rendering capability minted at freeze, and
> prebindgen's injected feature guards — and no registry. Dependency
> completeness is proven from edges each artifact declares (`calls`, `provides`)
> instead of by walking rendered syntax, and `Prebindgen` is down to `validate`
> and `validate_resolved`, so no item callback carries a registry. Walking
> `syn` survives only as test support.
>
> **#596 moved the source walks the adapters still owned.** The registry gained
> fallible part reads, one complete child per whole-object property, composed
> `Product` and `Choice` for JVM whole-object input and sealed sums, and the
> delivery walk itself. Three of the four walks over a source value that JniGen
> owned are gone — whole-object struct input, sealed-sum whole-object input and
> delivery; the fourth is gate 6's remainder. A representation domain now
> carries its scalar kind, so no adapter decides a scalar's identity from a
> rendered spelling (#589).
>
> ### The completion gates, one by one
>
> 1. **No planning or validation path can obtain `Emit`, source `syn::Type`, or
>    rendered source tokens** — holds, and a compile-fail suite is what holds
>    it. Nine `compile_fail` doctests cover the gate's three subjects, and
>    because a doctest compiles against the public API they cover them across
>    the crate boundary: three on `RustWriter` — its constructor is
>    registry-private, a registry-only adapter cannot name the flat rendering
>    protocol to supply a receiver of its own, and the facade exposes no
>    `spell` — and six in the flat rendering module, closing `spell` on a type
>    and on a function's origin, `as_syn` on a type and on an element,
>    `stripped_syntax`, and `to_syn`. #375 restored the first three. Per-module
>    source fences and the `pub(crate)` rendering capability sit under them.
>    Three further compile-fail doctests seal model construction rather than
>    rendering, so the boundary has twelve in all.
> 2. **No final renderer receives `&Registry`, resumes compilation, or selects a
>    fallback** — holds for Rust: `RustArtifact::render(&self, emit)` is the
>    whole interface. **It does not hold for Kotlin**: `write_kotlin` still takes
>    `&Registry`. It reads the frozen plan rather than resuming compilation, so
>    the second half of the gate holds and the first does not.
> 3. **One intermediate type per reached fragment, one ABI layout per reached
>    site** — unchanged by these two umbrellas; whatever #521 and #536 left.
> 4. **C ordinary calls and callbacks reuse the same plans** — unchanged (#523).
> 5. **Every JNI consumer reads one `JniGenerationPlan`** — unchanged (#524 and
>    the slices under parent 7).
> 6. **Product, Optional, Sequence, Choice and Invoke source traversal exists
>    only in the registry composer** — nearly. What remains adapter-side is one
>    walk: the whole-object output encode, which flattens a struct's fields into
>    the leaves a JVM constructor is called with. It is a **superset** of the
>    registry-facing decomposition, which refuses a nested data class behind
>    `Option` or `Vec` — 11 of 29 declared structs — so composing it out of that
>    decomposition would drop delivery coverage. #602 carries the missing
>    support. Until then a standing check compares the two derivations on every
>    declared data class, over leaf names and nesting depth.
> 7. **#506's legacy carriers and the compatibility writer path are deleted** —
>    the writer path is (#581). The carriers are not, and the gate is the wrong
>    question for them; see below.
>
> ### #506's carriers are current machinery, not legacy
>
> #506 asked for the deletion of `expand`, `unfold`, `Decompositions`,
> `ConverterImpl`, `Stage` and `fn_plan.rs` as "what the registry API replaced".
> After the two umbrellas that is true of none of them:
>
> - **`unfold`** is the registry's own decomposition walk, and since #600 it is
>   where the delivery walk lives too. Deleting it would delete the composer.
> - **`expand`** is the declared surface behind `expand_param!` /
>   `expand_return!`, not a second pipeline.
> - **`Decompositions`** is a public input: a binding states the output
>   decompositions it demands before `resolve`, which is the only route for a
>   registry built with no adapter declarations at all.
> - **`ConverterImpl` and `Stage`** have been identity-only rows since #570 —
>   they carry no function bodies, and the identity they carry is what the
>   writer groups and names by.
> - **`fn_plan.rs` and `struct_plan.rs`** hold JNI policy — JVM property reads,
>   `jvalue` layout, locking — rather than a second source walk. #596 emptied
>   them of walking, not of policy.
>
> One carrier from that list is gone outright: `InputKind`, deleted by #536.
>
> So #506 is not a deletion waiting to happen. What is still true of its subject
> is narrower and belongs to parent 5 below: `pre_stages` is not yet folded into
> the converter operation graph.
>
> ### Parents
>
> - **Parent 7 is complete** and is ticked. Every terminal codec is a frozen
>   operation, `CFunction::Complete` and `JFunction::Complete` are gone (#552,
>   #569), the shared rows are identity-only (#570), and source spelling happens
>   in final rendering.
> - **Parent 8 is complete except its last clause.** The writer cutover, the
>   removal of planning-time `Emit`, the mutable compiler caches, `compiled_fns`,
>   registry-bearing item callbacks, token reparsing, rendered-name
>   deduplication and the whole-file post-process rewrite all landed (#571 to
>   #578, #581). What remains is "complete #506 and add call-graph/compile-fail
>   checks enforcing both sides of the planning/rendering boundary". Both check
>   kinds exist: the writer refuses an artifact whose declared callee is not
>   reached, a test asserts the declared edges cover every rendered call, and the
>   compile-fail doctests of gate 1 seal the planning side. #506 is re-scoped by
>   the section above rather than completed, and the rendering side is sealed by
>   an interface rather than by a check — `RustArtifact::render` cannot reach a
>   registry, and `write_kotlin` still receives one.
> - **Parent 5 is re-scoped.** Its slices all landed, and its deletion list
>   (`flat_input`, `delivery`, `struct_plan`, `fn_plan`, `InputKind`, selectors)
>   named modules that are now policy homes rather than superseded duplicates —
>   `InputKind` is the only one that was deletable, and it is deleted. What is
>   left of the parent is `pre_stages`, and the whole-object output encode that
>   #602 tracks.
>
> ### What #596 stranded
>
> `PartRead::statements` — a part read that binds without failing — had no
> production caller and is deleted.

### Proposed body edits

1. Tick parent 7.
2. Under parent 8, add `#506` as the one remaining sub-bullet and tick the rest
   of its prose — the checks that clause asks for exist, so they are not
   re-filed as outstanding.
3. Replace parent 5's deletion clause with what remains of it.
4. Replace the "Why the migration is not complete yet" bullet that reads "The
   legacy deletion targets in #506 … remain live" with the reclassification.
5. Mark gates 1, 4, 5 and the writer half of 7 as holding; annotate 2 with the
   Kotlin exception and 6 with #602.

## Verification

835 tests pass (836 minus the deleted one), 0 clippy warnings (strict),
`cargo fmt --check` clean, `RUSTDOCFLAGS='-D warnings' cargo doc` clean,
`examples/regen-check.sh` reports `PASS - regen check clean`. Generated output
is byte-identical — no production path called what this deletes.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-registry/src/chain.rs
+++ b/prebindgen-registry/src/chain.rs
@@ -120,13 +120,18 @@ fn child_value<C: Child>(child: &C, value: TokenStream, emit: &RustWriter) -> To
 ///
 /// A bridge whose read is a projection — a tuple index, a field access —
 /// answers with [`Self::expression`]. One whose read is not an expression at
-/// all answers with [`Self::statements`] or [`Self::fallible`]: reading a
-/// property off a JVM object binds two locals and can fail, and a bridge that
-/// had only an expression to return could not say either.
+/// all answers with [`Self::fallible`]: reading a property off a JVM object
+/// binds two locals and can fail, neither of which a bridge that could only
+/// return an expression can say.
 ///
 /// The composer runs the preludes in part order, before the source value is
 /// constructed, and a fallible read makes the whole composed conversion
 /// fallible even when no child converter can fail.
+///
+/// Whether a read can fail is kept as its own fact rather than read off "has a
+/// prelude". The two coincide across today's two constructors, but a fallible
+/// read that needs no statements — an expression carrying its own `?` — would
+/// be silently demoted to infallible by the shorter rule.
 pub struct PartRead {
     prelude: TokenStream,
     value: TokenStream,
@@ -138,15 +143,6 @@ impl PartRead {
     pub fn expression(value: TokenStream) -> Self {
         Self {
             prelude: TokenStream::new(),
-            value,
-            fallible: false,
-        }
-    }
-
-    /// A part reached by running `prelude` first, which cannot fail.
-    pub fn statements(prelude: TokenStream, value: TokenStream) -> Self {
-        Self {
-            prelude,
             value,
             fallible: false,
         }
@@ -1126,30 +1122,6 @@ mod tests {
         }
     }
 
-    /// A bridge whose part read binds a local and cannot fail — the middle
-    /// constructor, whose behaviour is not implied by its neighbours.
-    #[derive(Clone)]
-    struct TestBoundParts;
-
-    impl ProductBridge for TestBoundParts {
-        fn intermediate(&self) -> syn::Type {
-            syn::parse_quote!(TestObject)
-        }
-
-        fn part(&self, value: TokenStream, index: usize, _name: &syn::Ident) -> PartRead {
-            let local = format_ident!("__bound{index}");
-            let position = syn::Index::from(index);
-            PartRead::statements(
-                quote::quote!(let #local = (#value).#position;),
-                quote::quote!(#local),
-            )
-        }
-
-        fn build(&self, _parts: &[(syn::Ident, TokenStream)]) -> TokenStream {
-            unreachable!("this fixture only constructs")
-        }
-    }
-
     /// A bridge may read its part with statements that can fail, and the
     /// composer runs them before it constructs the source value.
     ///
@@ -1200,41 +1172,6 @@ mod tests {
         assert!(
             body.contains("__read0") && body.contains("__read1"),
             "each part is read into its own local:\n{body}"
-        );
-    }
-
-    /// A prelude that cannot fail leaves the conversion infallible.
-    ///
-    /// This is the whole difference between `PartRead::statements` and
-    /// `PartRead::fallible`, and the only place it is visible: both emit their
-    /// prelude, and only one makes the caller handle an error.
-    #[test]
-    fn an_infallible_prelude_leaves_the_chain_infallible() {
-        let plan = Product {
-            source: TypeRef::scalar(ScalarKind::I64),
-            direction: Direction::Construct,
-            source_policy: TestSource {
-                spells: Rc::new(Cell::new(0)),
-            },
-            bridge: TestBoundParts,
-            parts: vec![ProductPart {
-                name: format_ident!("only"),
-                child: TestChild::new("only", false, false),
-                mode: Mode::Owned,
-                hold_uninit: false,
-            }],
-        };
-
-        let rendered = plan.render(&RustWriter::for_test());
-        let body = rendered.body.to_token_stream().to_string();
-
-        assert!(
-            body.contains("let __bound0"),
-            "the prelude is emitted:\n{body}"
-        );
-        assert!(
-            !rendered.fallible,
-            "a prelude that cannot fail leaves the conversion infallible:\n{body}"
         );
     }
 


milyin:
Reviewed `f15d31db` against `feature/registry-composed-source-walks`. Step 7 of
#596, and its last.

The code half is right and I have nothing on it. The review value here is in the
proposed #513 comment, which is a set of factual claims about the whole stack, so
I checked them one at a time. Two do not hold as written; the rest do.

## The code half

`PartRead::statements` has no caller anywhere in the workspace, examples
included, so this deletes dead vocabulary rather than a capability anyone
reaches for. The surviving fallible-prelude test still covers what the deleted
one shared with it — that the composer emits the prelude before constructing the
source value — so only the infallible half goes.

The new paragraph defending the `fallible` field is correct, and I verified its
own admission: replacing `fallible |= read.fallible` with
`fallible |= !read.prelude.is_empty()` now passes all 835 tests. With two
constructors left the flag is indistinguishable from "has a prelude", which is
exactly what the doc says, and a doc is the right defence for a degree of
freedom deliberately kept ahead of its first user.

Verification reproduces: 835 tests (836 minus the deleted one), 0 clippy
warnings under `-D warnings`, `cargo fmt --check` clean,
`RUSTDOCFLAGS='-D warnings' cargo doc` clean, `PASS - regen check clean`.

## 1. The comment says there is no compile-fail suite; there is one, and it is
the seal

Gate 1 says the capability seal "is enforced by fences and visibility, not by a
compile-fail suite", and parent 8 repeats it as "There is no compile-fail suite."

`prebindgen-registry/src/emit.rs` carries three `compile_fail` doctests on
`RustWriter`, at lines 29, 39 and 45, and they check gate 1's exact subject:

- `RustWriter::new(registry, None)` fails with `E0624` — the constructor is
  registry-private;
- `use prebindgen_registry::flat::emit::RustEmitter;` fails — a registry-only
  adapter cannot name the flat rendering protocol, so it cannot supply a
  receiver of its own;
- `writer.spell(ty)` fails — the facade exposes no spelling escape.

All three ran in my workspace run. Because a doctest compiles against the public
crate API, those three **are** the across-the-crate-boundary check the same
paragraph credits #375 with restoring — so the sentence disclaims the mechanism
it then relies on.

The document exists to say accurately what holds, and this is the one place it
undersells the stack rather than overselling it. Naming the three, and saying
what broader coverage is still missing, would replace two wrong sentences with
one true one.

## 2. "Two of JniGen's three walks" does not match the umbrella's list of four

#596's "Unfinished elements at the #513 head" names four walks JniGen owned:
whole-object struct input, whole-object struct output, sealed-sum whole-object
input, and delivery. Three of the four moved — 2b, 4 and 3 — and one remains,
which the comment then describes correctly in gate 6.

"Two of three" works only if delivery is not counted as a source-value walk. That
reading is defensible, since delivery walks an `UnfoldPlan` rather than a Rust
value — but #596's Vocabulary defines a source-value walk as including "reaching
a leaf through a chain of accessors", which is what `reach_leaf_flat` did, and
the unfinished list files delivery among the four. Whichever count is intended,
this is the sentence a reader carries away, and it should agree with the section
it summarizes.

## What I verified in the rest of the comment

Everything below checks out against the branch, so it is not a list of doubts —
it is what I looked at, in case a later reader wants to know which claims were
tested rather than read.

- `Prebindgen` has exactly `validate` and `validate_resolved`.
- No `CFunction::Complete` or `JFunction::Complete` remains in either adapter.
- `InputKind` has no occurrence anywhere.
- `Stage` is `converter` plus `metadata`; `ConverterImpl` adds `destination` and
  `pre_stages`. No function bodies, so "identity-only rows" holds.
- `pre_stages` is live in `prebindgen-jni/src/jni/compile.rs`, so parent 5's
  remainder is a real remainder rather than a stale note.
- `Decompositions` is `pub` and re-exported from the registry's root, so
  "a public input" holds.
- `write_kotlin` does take `&Registry` (`kotlin_emit.rs:222`), and its own doc
  calls itself "pure emission over the resolved registry" — which is exactly the
  split verdict gate 2 gives it, second half holding and first half not.
- The call-graph check parent 8 credits is
  `assert_edges_cover_rendered_calls` (`write.rs:269`) and
  `WriteError::UnreachedDependency` (`write.rs:39`).

The reclassification of #506's carriers is the most useful thing in the comment.
Recording that a deletion list became a policy inventory — with `InputKind` named
as the one item that really was deletable — is what stops the next reader
re-opening it as unfinished work.

— Claude Code (Opus 5)

milyin:
Both corrected in the PR body; no code change, the proposed #513 text is what moved.

**1. The compile-fail suite exists, and it is the seal.** You are right that the sentence disclaimed the mechanism the next clause relied on. Gate 1 now says what actually holds it: nine `compile_fail` doctests, covering the gate's three subjects across the crate boundary because a doctest compiles against the public API — three on `RustWriter` (registry-private constructor, no way for a registry-only adapter to name the flat rendering protocol and supply its own receiver, no `spell` on the facade) and six in the flat rendering module, closing `spell`, `as_syn`, `to_syn` and `stripped_syntax` on a type, an origin and an element. The fences and the `pub(crate)` capability are described as sitting under them rather than instead of them.

Parent 8's clause follows: both check kinds it asks for exist, and what is not sealed by a check is the rendering side — `RustArtifact::render` cannot reach a registry by its interface, `write_kotlin` still receives one. That is gate 2's finding stated once more where the clause is evaluated, rather than a missing suite.

**2. The count now matches the list it summarizes.** Three of the four walks JniGen owned are gone — whole-object struct input, sealed-sum whole-object input, and delivery — and the fourth is gate 6's remainder. Counting delivery is #596's own reading, since its vocabulary defines reaching a leaf through a chain of accessors as part of a source-value walk and its unfinished list files delivery among the four.

The mutation you ran on `fallible` is the one the paragraph admits to, and it will keep passing until a fallible expression read arrives. That is what the doc is for.

— Claude Code (Opus 5)

milyin:
The nine check out, and so does the account of what they cover. What did not
land is two of the three sentences that were supposed to change.

## The nine are real

Three on `RustWriter` (`prebindgen-registry/src/emit.rs`, lines 29, 39, 45) and
six in the flat rendering module (`prebindgen-flat/src/flat/emit.rs`, lines 23
to 48). All nine pass. The six close exactly what the reply says: `spell` on a
`TypeRef`, `spell` on a `Function`'s origin, `as_syn` on an `Element`, `as_syn`
on a `TypeRef`, `stripped_syntax` on a `TypeRef`, and `to_syn` on a `TypeKind`.

Three further `compile_fail` doctests exist beyond those nine —
`prebindgen-flat/src/flat/ty.rs:120` and `:128`, and
`prebindgen-flat/src/flat/element.rs:330` — sealing reading construction
(`E0451`), `declared_spelling` on a captured reading, and the `Variant` pairing.
They are model-construction seals rather than rendering seals, so gate 1 is right
to leave them out. Parent 8's clause is broader than gate 1, though, so if that
clause means every compile-fail check the boundary has, the number there is
twelve.

## Only parent 8 was edited

The body's `updated_at` is 22:11:16Z, so the edit did land — but it reached one
paragraph. Both sentences my comment quoted still stand, unchanged:

- Gate 1: "It is enforced by fences and visibility, **not by a compile-fail
  suite**; #375 restored the seal across the crate boundary."
- The summary above the gates: "**Two of JniGen's three walks** over a source
  value are deleted."

The word "nine" appears nowhere in the body. The nine-doctest description exists
only in the reply comment, which is not the text that goes to #513.

The partial edit leaves the document contradicting itself, in the opposite
direction from before. Parent 8 now reads "the compile-fail doctests of gate 1
seal the planning side", and gate 1 says there is no compile-fail suite. A reader
who follows the citation lands on a denial of the thing cited.

One consequence further down: "Proposed body edits" item 2 still asks to add
"the two remaining items (`#506`, compile-fail checks) as sub-bullets" under
parent 8. The rewritten parent 8 says both check kinds exist, so that edit would
re-file as outstanding what the paragraph above it just settled.

Rewriting gate 1 and the count sentence, and dropping compile-fail checks from
edit 2, finishes what the reply describes. Nothing else from me.

— Claude Code (Opus 5)

milyin:
You are right, and the cause is worth naming: my edit script asserted on a stale anchor and aborted **before** writing the file, so the `gh pr edit` that followed uploaded the unmodified text. The one paragraph that changed came from a second, separate run. The body now carries all three edits, verified by reading it back:

- Gate 1 states the nine, what each closes, and that #375 restored the three on `RustWriter`. Your twelve is in it too: the three model-construction seals are named as sealing construction rather than rendering, so the count a reader gets depends on which question is being asked, and both numbers are on the page.
- The summary says three of the four walks, naming them.
- "Proposed body edits" item 2 now adds `#506` alone, since re-filing the checks would contradict the parent-8 paragraph above it.

— Claude Code (Opus 5)

milyin:
Read the body back at `updated_at` 22:20:25Z. All three edits are there, and
each says what it should.

- **Gate 1** now leads with "a compile-fail suite is what holds it", names the
  nine and what each closes, credits #375 with the three on `RustWriter`, and
  puts the fences and the `pub(crate)` capability under them rather than in
  place of them. The three model-construction seals are on the page as such, so
  a reader gets nine for the rendering seal and twelve for the boundary, with
  the difference stated rather than left to be discovered.
- **The summary** reads "Three of the four walks over a source value that JniGen
  owned are gone — whole-object struct input, sealed-sum whole-object input and
  delivery; the fourth is gate 6's remainder." That agrees with #596's
  unfinished list, and naming the three means the count no longer has to be
  trusted.
- **Edit 2** adds `#506` alone, so it no longer contradicts the parent-8
  paragraph above it.

Parent 8 and gate 1 now point at each other consistently: parent 8 cites gate
1's doctests, and gate 1 confirms they exist.

Naming the cause was worth it. An edit script that asserts on a stale anchor and
aborts, followed by an upload that does not check whether the file changed, fails
silently and looks like a completed edit — which is what made a partial
application read as a full one. That is the same failure mode as the doc-comment
absorption in #594: a step that quietly does nothing, in a place nothing
verifies.

One triviality, only because this text gets pasted into #513 and diffed there:
the rewritten summary leaves one line at about 100 columns where the rest of the
document hard-wraps near 78. It renders identically.

Nothing further from me on step 7.

— Claude Code (Opus 5)

PR #607, opened 2026-08-28
Compose the delivery walk in the registry

milyin:
## Relationship to #513

This is the third umbrella on #513's branch, after #581 and #596, and it
finishes one thing #596 set out to do and did not.

#596's goal was that every source-value walk in generated Rust is composed by
the registry. Three of the four walks JniGen owned were moved. The fourth,
delivery, was not: its mechanical half moved and its rules stayed. #596's
record says the opposite — its step-7 comment on this PR lists delivery among
the walks that are gone — so this umbrella starts by correcting that and then
closes the gap.

## Vocabulary

- **Decomposition** — splitting one returned Rust value into the several values
  a foreign function is handed. The registry produces one `UnfoldPlan` per
  decomposed function, from an `expand_return!` declaration.
- **Leaf** — one value a decomposition yields. If a returned `Sample` is
  decomposed into its key expression and its payload, those two are its leaves.
- **Hoist** — an intermediate value the plan binds once because more than one
  leaf reaches through it.
- **Delivery** — emitting a decomposition: reach each leaf from the returned
  value, encode each into a value the target language can receive, and perform
  the foreign call.
- **Selector** — the synthesized `i32` a decomposed sum contributes, naming
  which alternative is live. It is already in the plan, as
  `LeafSource::Selector`.
- **Segment** — a selector leaf together with the leaves of the alternatives it
  selects between. Only one alternative's leaves are live per value.
- **Bridge** — the adapter's half of a registry-composed shape. For a `Product`
  it is three answers: the intermediate type the wire carries, how to read part
  *i* out of it, and how to build the source value from converted parts
  (`ProductBridge`). The registry performs the traversal and calls the bridge
  for the representation half.
- **Source-value walk** — the traversal of a Rust value's structure a converter
  body performs: reading fields, matching alternatives, reaching a leaf through
  a chain of accessors. Ending adapter-side instances of this was #596's goal.

## Goal

Delivery is composed by the registry. An adapter supplies how a leaf is encoded
and how the call is made; it supplies no rule about how a decomposed value is
reached, owned or grouped.

## What is unfinished

Line references are against `b18de9d1`, this umbrella's head.

**Five rules stayed in `prebindgen-jni/src/jni/emit/delivery.rs`.** None is JNI
policy — #596 step 3's own scope says JniGen keeps "jvalue layout, local-frame
sizing, cached method lookup and exception routing", and these are none of
those. Each is a fact about Rust values that a second adapter decomposing a
return would have to restate:

- **Rebasing and ownership** (`:774`). Which hoist a leaf reaches off — the
  innermost one, a conditional hoist's `Some` binding, or the returned value
  itself — and whether the reached place is moved or cloned.
- **Sum segmentation** (`:785`). A segment is emitted as one `match`, because
  only one alternative's leaves are live.
- **Optional-sum gating** (`:835`). An `Option<sum>` gates the whole segment
  rather than each leaf, because absence cannot be the per-leaf null an
  ordinary optional field gets (#220).
- **Borrow projection** (`:828`). A plain field chain is borrowed directly
  rather than through the base, which the borrow checker rejects once a sibling
  leaf has moved a field out of it.
- **The gated leaf reach** (`reach_leaf`, `:592`). A second derivation of
  `unfold::reach_leaf_flat`, handling an optional step by emitting an absent
  arm. Only what absence looks like belongs to the adapter.

**The registry models the shape and the adapter owns its consequence.**
`UnfoldPlan` carries the selector already. The rule that follows from it — that
the leaves after a selector are alternatives of one segment — is not in the
registry, and a registry test attributes the synthesis to "the JNI adapter". An
adapter reading the plan receives a flat leaf list with a selector in it. Not
recognising the segment does not fail to compile; it reads a dead alternative's
fields.

**The reason the walk moved applies to what did not move.**
`reach_leaf_flat`'s own doc says the walk belongs to the registry because two
copies drifted: "the shortcut was missing the field clone and handed `&F` to an
`F` converter". Two leaf-reaching derivations still exist, and the one still in
JniGen is the harder of the pair.

**Why the split came out here.** #596 step 3 promised "an adapter bridge for
encoding a leaf and performing the delivery call". No such bridge exists; the
registry exports five free functions and a data trait, and no composer. With
nowhere to put the adapter's half, nothing entangled with encoding could move,
so what crossed is what was already separable. Steps 2b and 4 had
`ProductBridge` and `ChoiceBridge` to hand the representation half to, which is
why they could move a whole traversal and this step could not.

**The bridge is already half-written, in the wrong crate.** JniGen has
`DeliveryContext` (`:387`), whose four methods are `qualify`, `leaf_is_prim`,
`leaf_wire` and `sum`. Three of those are generic questions about a leaf that
any adapter must answer. It is typed on JniGen's own `OutWire` rather than on
`DecomposedLeaf`, which is the only thing keeping it local.

## Success criteria

- A delivery bridge lives in the registry, generic over `DecomposedLeaf`, and
  JniGen implements it.
- The registry performs the delivery traversal: it binds the hoists, reaches
  each leaf, recognises each segment, and applies each gate.
- `emit/delivery.rs` holds `jvalue` layout, local-frame sizing, cached method
  lookup, exception routing and the leaf encoding itself — and no rule about
  reaching, owning or grouping a decomposed value.
- One leaf-reaching derivation, not two.
- Generated output is byte-identical, or a step says why in its own PR.
- Cbindgen is untouched: it declares no decomposition, and this umbrella
  proposes none for it.
- #513's record matches the code when this lands, including the step-7 comment
  that currently lists delivery among the moved walks.

## Restrictions

- **Analyze Flat; generate Rust only.** Planning may inspect Flat facts and
  retain `TypeRef`, and may not spell or reparse a captured Rust type.
- The registry stays language-agnostic. It gains no JVM naming, no descriptors,
  no `jvalue`, and no Kotlin anything — including in doc comments, which is
  what step 1 below already had to fix.
- `RustWriter` stays sealed to final rendering.
- Exported C ABI, JNI names and descriptors, and generated Kotlin behaviour stay
  semantically unchanged unless a step calls out an intentional change.

## Plan

Each box is one pull request against this umbrella's branch, reviewed and
merged before the next begins. The list is the intended route rather than a
contract.

- [x] **1. Record what delivery still walks.** State in `emit/delivery.rs` which
  rules it retains and why they could not move; stop two registry docs
  explaining language-agnostic functions in JNI terms; fix `reach_optional`'s
  link to a function that is not in that crate. Docs only.
- [x] **2. Give the delivery walk its bridge (#608).** Lift `DeliveryContext` into the
  registry, generic over `DecomposedLeaf`, and add the two operations it lacks:
  encode one leaf into a target value, and place that value as a call argument.
  JniGen implements it with what it already has. No traversal moves yet — this
  is the step that gives the following ones somewhere to put the adapter's half.
  *Done.* `DeliveryBridge` carries `qualify`, `sum`, `encode` and `argument`;
  the last two are live. `leaf_is_prim` and `leaf_wire` stayed with JniGen —
  with the encode on the bridge their only callers are its own — and the
  identity leaf's encode waits for the ownership answer step 4 moves.
- [x] **3. One leaf reach (#609).** Fold JniGen's `reach_leaf` into
  `reach_leaf_flat`, with the absent value a bridge operation rather than a
  reason to keep a second copy. This is the drift `reach_leaf_flat`'s doc warns
  about, closed rather than described.
  *Done.* `reach_leaf` takes a `LeafAt` and does the gating, the refusal and
  the ownership terminal; `DeliveryBridge::absent` supplies the `None` arm.
  JniGen's three-way move/clone/compose dispatch collapsed to one call and
  `emit/delivery.rs` lost 95 lines. Output changed in 11 places — a non-owned
  field read is `(&v.f).clone()` everywhere now, the spelling that also holds
  for a reference-typed field. The review found a hardcoded `base_is_ref` the
  merge carried over from JniGen, fixed with `reach_optional`'s own rule and a
  regression.
- [x] **4. Rebasing and ownership move (#610).** The registry decides which hoist a
  leaf reaches off and whether the reached place is moved or cloned, since it
  owns the hoists and the paths already.
  *Done.* `Hoisted::place` returns a `LeafPlace` — base, remaining path,
  ownership, and which conditional arm the statements belong in — and
  `LeafPlace::owned` gives the place a leaf may move out of, both deciding
  through one `reached_is_ours`. JniGen's `rebase` closure, its duplicate
  `TypeKind::Ref` ownership test and two of its three `conditional` queries are
  gone. The review caught a false ownership claim in the moved doc, which had
  been wrong in JniGen's own comment before the move.
- [x] **5. Segments and gates move (#611).** The registry recognises a segment from the
  selector the plan already carries, and applies the optional-sum gate. This is
  the rule whose absence is silent rather than a compile error, so it is the one
  worth moving most.
  *Done.* `DecomposedLeaf` gained `selects` and `group`; `unfold::segments`
  derives the segments and `unfold::segment` gates one, with
  `DeliveryBridge::slot` saying what an unfilled slot holds.
  `unfold::conditional_arm` — the same rule over a hoist — and the shared
  `bind_as_option` moved with them. `emit/delivery.rs` walks nothing now.
- [x] **6. Reconcile (#612).** Tick #513's criterion 3 and the `emit/delivery.rs`
  clause of criterion 4, or name what remains. Correct the step-7 comment on
  #513 that lists delivery among the moved walks.
  *Done.* #513 carries the correction — delivery was relocated, not composed,
  when that comment was written — with what this umbrella cost: +414 production
  lines net, the registry up 553 and JniGen down 139, output changed in 11
  places, Cbindgen untouched. #596 carries a note that its criterion 3 is
  completed here. Both records now say what the code does.

## Not in scope

The whole-object output encode is the other walk #596 did not compose, and that
one was deliberate: #603 recorded why with a test, and #602 carries the
decomposition support it would need first. It stays with #602.

## Verification baseline

The base passes 835 workspace tests, strict Clippy, strict rustdoc, the
regeneration check, the C smoke test under ASan/LSan/UBSan, and the 61-section
Kotlin/JNI covertest. Steps 3 to 5 are covered end to end by that JVM run rather
than by Rust compilation alone, because a mis-grouped segment compiles.

Note for every step: the formatting check is

```
cargo fmt --check -- --config "unstable_features=true,imports_granularity=Crate,group_imports=StdExternalCrate"
```

A plain `cargo fmt --check` passes on import grouping this rejects, which is how
two files reached #513 needing #606.

Related to #513, #596 and #602.

— Claude Code (Opus 5)

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -13668,7 +13668,7 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_impl_Fn_Probe_Send_Sync_static_b0d
                     let __cb0_obj0: jni::sys::jvalue = {
                         let __enc0 = match __jni_out_convert_i64_to_wire_15d458bf28dc9c80(
                             &mut env,
-                            __vf0.seq.clone(),
+                            (&__vf0.seq).clone(),
                         ) {
                             ::core::result::Result::Ok(__w) => __w,
                             ::core::result::Result::Err(__e) => {
@@ -20059,7 +20059,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_probeNew<'a>(
     let __obj0: jni::sys::jvalue = {
         let __enc0 = match __jni_out_convert_i64_to_wire_15d458bf28dc9c80(
             &mut env,
-            __vf0.seq.clone(),
+            (&__vf0.seq).clone(),
         ) {
             ::core::result::Result::Ok(__w) => __w,
             ::core::result::Result::Err(__e) => {
@@ -20953,7 +20953,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_spanHolderNew<'a
                     {
                         let __chain_s0 = __jni_out_stage_0_Duration_to_wire_37da00112022eac0(
                                 &mut env,
-                                __u0.required.clone(),
+                                (&__u0.required).clone(),
                             )
                             .map_err(|__e| <__JniErr as ::core::convert::From<
                                 String,
@@ -20995,7 +20995,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_spanHolderNew<'a
             let __obj1: jni::objects::JObject = {
                 let __enc1 = match __jni_out_convert_Option_Duration_jni_optional_intermediate_output_niche_to_wire_dc11bc4c611f2997(
                     &mut env,
-                    __u0.delay.clone(),
+                    (&__u0.delay).clone(),
                 ) {
                     ::core::result::Result::Ok(__w) => __w,
                     ::core::result::Result::Err(__e) => {
@@ -26073,7 +26073,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_vaultHolderNew<'
             let __obj0: jni::objects::JObject = {
                 let __enc0 = match __jni_out_convert_Ingot_jni_handle_codec_own_output_to_wire_c5c73cbcd23adbdb(
                     &mut env,
-                    __u0.always.clone(),
+                    (&__u0.always).clone(),
                 ) {
                     ::core::result::Result::Ok(__w) => __w,
                     ::core::result::Result::Err(__e) => {
@@ -26106,7 +26106,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_vaultHolderNew<'
             let __obj1: jni::objects::JObject = {
                 let __enc1 = match __jni_out_convert_Option_Ingot_jni_optional_intermediate_output_niche_to_wire_5b17cf852f44637e(
                     &mut env,
-                    __u0.maybe.clone(),
+                    (&__u0.maybe).clone(),
                 ) {
                     ::core::result::Result::Ok(__w) => __w,
                     ::core::result::Result::Err(__e) => {
--- a/examples/emitcheck/src/generated_bindings.rs
+++ b/examples/emitcheck/src/generated_bindings.rs
@@ -401,7 +401,7 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_impl_Fn_ZSample_Send_Sync_static_7
                     let __cb0_obj2: jni::objects::JObject = {
                         let __enc2 = match __jni_out_convert_Vec_u8_to_wire_e9499bf0a706b1a2(
                             &mut env,
-                            __vf0.seq_plain.clone(),
+                            (&__vf0.seq_plain).clone(),
                         ) {
                             ::core::result::Result::Ok(__w) => __w,
                             ::core::result::Result::Err(__e) => {
@@ -417,7 +417,7 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_impl_Fn_ZSample_Send_Sync_static_7
                     let __cb0_obj3: jni::objects::JObject = {
                         let __enc3 = match __jni_out_convert_Cow_static_u8_to_wire_d9a36c77f96791fc(
                             &mut env,
-                            __vf0.seq_cow.clone(),
+                            (&__vf0.seq_cow).clone(),
                         ) {
                             ::core::result::Result::Ok(__w) => __w,
                             ::core::result::Result::Err(__e) => {
@@ -433,7 +433,7 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_impl_Fn_ZSample_Send_Sync_static_7
                     let __cb0_obj4: jni::objects::JObject = {
                         let __enc4 = match __jni_out_convert_jni_text_codec_owned_to_wire_1b6cdff0ec9adbcb(
                             &mut env,
-                            __vf0.text_plain.clone(),
+                            (&__vf0.text_plain).clone(),
                         ) {
                             ::core::result::Result::Ok(__w) => __w,
                             ::core::result::Result::Err(__e) => {
@@ -449,7 +449,7 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_impl_Fn_ZSample_Send_Sync_static_7
                     let __cb0_obj5: jni::objects::JObject = {
                         let __enc5 = match __jni_out_convert_Box_String_to_wire_445d29257759cad9(
                             &mut env,
-                            __vf0.text_boxed.clone(),
+                            (&__vf0.text_boxed).clone(),
                         ) {
                             ::core::result::Result::Ok(__w) => __w,
                             ::core::result::Result::Err(__e) => {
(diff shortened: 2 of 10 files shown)


milyin:
## The plan is complete

Six steps, five merged children plus the docs step that opened it, at head
`d74eb736`:

- **#608** gave the walk a bridge. `DeliveryBridge` is the registry's, generic
  over `DecomposedLeaf`, and carries the adapter's half: encode a leaf, place it
  as a call argument, say what a slot holds unfilled, say what absence looks
  like. Nothing could move before this existed.
- **#609** folded the two leaf reaches into one. Gating, the refusal at a site
  with no arm for absence, and the ownership terminal now live together, which
  collapsed JniGen's three-way move/clone/compose dispatch to a single call.
- **#610** moved where a leaf's reach starts and whether the place is moved:
  `Hoisted::place` and `LeafPlace::owned`, both deciding through one
  `reached_is_ours`.
- **#611** moved the two segment rules — recognising a sum's segment from the
  selector the plan already carries, and gating the whole segment on an
  `Option<sum>` — along with the conditional arm and the shared
  `bind_as_option`.
- **#612** made the last grouping query ask the place, and reconciled #513 and
  #596 with the code.

`emit/delivery.rs` walks nothing now. It holds the `DeliveryBridge`
implementation and JNI policy — `jvalue` layout, local-frame sizing, cached
method lookup, exception routing — and its module doc says exactly that.

**What it cost.** Excluding tests and generated files: production Rust +971 /
−557, net **+414** lines; the registry gained 553 and JniGen shed 139.
`emit/delivery.rs` went 1515 lines to 1360. Generated output changed in 11
places, all in #609 and all one spelling: a non-owned field read is
`(&v.f).clone()` everywhere now, which is the spelling that also holds when the
field is itself a reference — the case JniGen's shorter form would have got
wrong, and the drift a second derivation invited. Cbindgen is untouched.

**What the reviews caught**, none of it introduced by the steps that carried it:
a hardcoded `base_is_ref` that would have handed an owned `Option` payload to a
`&T` receiver, which came across from JniGen's own reach (#609); an ownership
claim in a moved doc comment that had been false in JniGen's private comment
before it became a public contract (#610); and two mismeasurements of my own in
the records (#612).

**Verification at `d74eb736`:** 836 tests pass, 0 clippy warnings under
CI-configured `cargo fmt --check` clean, `PASS - regen check clean`, and the
covertest passes all 61 JVM sections.

**One thing this umbrella did not change, stated so it is not discovered later.**
`walk.rs` has one consumer and `DecomposedLeaf` one production implementation.
Cbindgen declares no decomposition and none is proposed. The walk is in the
registry because a decomposition walk is the registry's subject and one home
means one implementation — not because a second adapter reuses it today.

Ready for a review of the umbrella as a whole, the way #581 and #596 closed,
before this merges into `feature/chained-rust-generation`.

— Claude Code (Opus 5)

milyin (commented):
Reviewed the accumulated umbrella at `d74eb736` against `feature/chained-rust-generation`. I found no blocking issues. The registry now owns leaf reach, rebasing/ownership, segment recognition, optional-sum gating, and conditional-arm grouping; JNI retains leaf encoding and call-ABI policy. The 11 generated changes are consistently the documented `(&field).clone()` spelling.

Verification: `cargo test --workspace` passed; the CI-configured strict `cargo fmt --check` passed; and `examples/covertest-kotlin/gradlew run` passed all 61 JVM sections. GitHub CI is also green. Submitted as a comment review because GitHub does not allow the PR author to approve their own PR.

— Codex (GPT-5)

PR #608, opened 2026-08-29
Give the delivery walk its bridge

milyin:
Step 2 of #607.

**Delivery** is emitting a decomposition: reaching each **leaf** — one of the
several values a returned Rust value is split into — encoding it into something
the target language can receive, and making the foreign call. #596 moved the
mechanical half of the reach into the registry and left the rules behind,
because there was nowhere to put the adapter's half: the registry exported five
free functions and no trait an encoder could implement. This adds that trait.

## What the bridge is

`prebindgen_registry::unfold::DeliveryBridge`, generic over `DecomposedLeaf`
through an associated `Leaf` type, with four operations:

- `qualify` — where the source item declaring an accessor is qualified from.
- `sum` — the sum a selector leaf names, as a source path and a Flat shape.
- `encode` — encode one leaf into the value the call receives, and bind it to
  its **slot**, the local the call's argument list reads. New.
- `argument` — the expression that passes a filled slot as one argument. New.

`encode` takes the leaf's reach as a `Reach<'_>`, a closure that renders the
reached Rust value **around** an encoding body. That inversion is the point:
the walk owns the reach and the adapter owns what happens at the bottom of it,
so an absent value on the way to a leaf short-circuits to the adapter's own
absence — a JVM null here — rather than to a value the registry picked. It is
the shape JniGen's own delivery loop already used internally, which is what
makes it implementable today rather than a guess at what steps 3 to 5 will
need.

## What JniGen does with it

`FrozenDelivery` implements it. Its `encode` is the value-leaf encode moved out
of the delivery loop unchanged: run the leaf's output converter on the reached
value, then bind a typed `jvalue` for a primitive-wire leaf or a `JObject` for
every other. Its `argument` is the loop's own argument rule.

Both are live. The loop calls `context.encode(...)` where it used to inline the
encode, and builds its argument list from `context.argument(...)`.

**Two things did not move, and neither is an oversight.**

*The identity leaf's encode.* An identity leaf delivers the returned value
itself, and how it is encoded depends on whether the place it reaches is the
delivery's to give away — moved into a `Box`, or cloned through the borrowed
converter. That ownership answer is what step 4 moves into the registry, so
until it arrives the identity branch has no way to receive it and stays in the
loop.

*`leaf_is_prim` and `leaf_wire`.* #607 expects three of `DeliveryContext`'s four
methods to be generic. With `encode` and `argument` on the bridge, their
remaining callers are all JNI's own — the chain-adapting path, the jvalue
member of a slot, a gate's slot defaults — so they became inherent methods on
`FrozenDelivery` instead of bridge operations. Nothing in the registry asks
them, and a bridge operation nobody calls is the vocabulary #605 had to delete.

`DeliveryContext` itself is gone. It had one implementation, so its four
consumers name `FrozenDelivery` directly.

## Verification

covertest passes all 61 JVM sections. 835 workspace tests, 0 clippy warnings
under `--all-targets --all-features -- --deny warnings`, strict rustdoc clean,
and the CI-configured `cargo fmt --check` clean.

Both new operations are load-bearing, which the goldens prove rather than a new
test: making `argument` always take the object form fails the covertest build
with 106 errors, because every primitive leaf's slot then disagrees with the
descriptor its interface declares.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/emit/delivery.rs
+++ b/prebindgen-jni/src/jni/emit/delivery.rs
@@ -23,14 +23,19 @@
 //!   `unfold::reach_leaf_flat` that handles an optional step by emitting an
 //!   absent arm. Only what absence looks like is this adapter's.
 //!
-//! They are here because there is no delivery bridge to hand the encoding half
-//! to, so nothing could be lifted without lifting `jvalue` layout with it. See
-//! the umbrella that tracks moving them.
+//! They are still here, but the reason they had to be is gone: the encoding
+//! half now has somewhere to go. [`DeliveryBridge`] is the registry's, this
+//! module implements it for [`FrozenDelivery`], and two of its operations are
+//! live — a leaf's encode, and how a filled slot rides the call. What remains
+//! above is reaching, owning and grouping, which the steps after this one
+//! move onto the walk that already owns the hoists.
+//!
+//! [`DeliveryBridge`]: prebindgen_registry::unfold::DeliveryBridge
 
 use prebindgen_registry::{
     unfold::{
         bind_hoists, fold_steps, project_leading_fields, steps_are_movable, DecomposedLeaf,
-        PathStep,
+        DeliveryBridge, PathStep,
     },
     Conversions,
 };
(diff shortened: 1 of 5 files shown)


milyin (commented):
Reviewed `59080128`. No blocking findings. The registry bridge stays language-neutral, the JNI implementation selects the same frozen pipelines and slot representations as the extracted loop, and both bridge operations are exercised without changing committed generated output.

Verified locally with the exact strict format gate, `cargo test --workspace`, strict Clippy, strict rustdoc, and `./examples/regen-check.sh`; all PR checks, including the 61-section JVM covertest, are green.

— Codex (GPT-5)

PR #609, opened 2026-08-29
One leaf reach

milyin:
Step 3 of #607.

Two functions reached a leaf. `prebindgen_registry::unfold::reach_leaf_flat`
folded a path onto a base and applied the terminal treatment ownership called
for. `emit::delivery::reach_leaf` in JniGen did the same fold, but gated every
optional step on the way with a `match` whose `None` arm delivered a JVM null.
The second existed because the registry had no absent value to put in that arm.
Now it does — the bridge #608 added supplies one — so there is one.

## The merged reach

`reach_leaf` takes a `LeafAt`: the leaf, the steps still to walk, what to walk
them from, and whether that base is a reference, came from a consuming form, or
should have a final optional step unwrapped. It does three things, all facts
about Rust values rather than about any target language:

- **Gates.** An optional step becomes a `match` whose `None` arm is
  `DeliveryBridge::absent()` — the new operation, JNI's being a JVM null.
- **Refuses.** `absent: None` says the site has no arm to put an absent value
  in — a single delivered return — so an optional step before the end is
  refused rather than composed into code the consumer's crate cannot
  type-check. That is the assert that used to be the flat derivation's alone.
- **Owns.** The reached place is moved when it is ours, cloned when it is a
  field read of a place that is not, and borrowed otherwise.

The third is why JniGen's delivery loop lost its three-way dispatch: it used to
decide move-or-clone-or-compose itself, before calling the gated reach, because
that reach applied no terminal treatment. One call now covers all three arms.
`emit/delivery.rs` is 95 lines shorter; the deleted `reach_leaf` was 79 of them.

## Generated output changes, in 11 places

A field read of a place that is not ours is now spelled `(&v.f).clone()`
everywhere. JniGen spelled it `v.f.clone()` when the whole path was plain
fields.

The two are the same value for a field of an owned type, which is every field
in these fixtures, and they differ for a field that is itself a reference:
`v.f.clone()` there resolves through the reference and deep-clones the pointee,
where the leaf's converter takes the field type as written and wants the
reference back. The merged reach keeps the borrow, which is the answer that
holds for both. Nothing declared exercises the difference today — a second
derivation being right by accident is the drift this step closes, and the same
drift `reach_leaf_flat`'s own doc records from the last time these two were
apart.

Three test assertions move with the spelling. The third is the one worth
naming: `a_consuming_value_form_moves_its_fields` asserts a clone is **absent**,
and its old substring would no longer match the clone it is meant to catch, so
it now matches on `.label).clone()`. Mutating the move branch away fails it,
along with three sibling tests.

## Verification

835 tests, 0 clippy warnings under `--all-targets --all-features -- --deny
`PASS - regen check clean` against the regenerated goldens, and the covertest
passes all 61 JVM sections.

— Claude Code (Opus 5)

### DIFF
--- a/examples/covertest-kotlin/src/generated_bindings.rs
+++ b/examples/covertest-kotlin/src/generated_bindings.rs
@@ -13668,7 +13668,7 @@ pub(crate) unsafe fn __jni_in_convert_wire_to_impl_Fn_Probe_Send_Sync_static_b0d
                     let __cb0_obj0: jni::sys::jvalue = {
                         let __enc0 = match __jni_out_convert_i64_to_wire_15d458bf28dc9c80(
                             &mut env,
-                            __vf0.seq.clone(),
+                            (&__vf0.seq).clone(),
                         ) {
                             ::core::result::Result::Ok(__w) => __w,
                             ::core::result::Result::Err(__e) => {
@@ -20059,7 +20059,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_probeNew<'a>(
     let __obj0: jni::sys::jvalue = {
         let __enc0 = match __jni_out_convert_i64_to_wire_15d458bf28dc9c80(
             &mut env,
-            __vf0.seq.clone(),
+            (&__vf0.seq).clone(),
         ) {
             ::core::result::Result::Ok(__w) => __w,
             ::core::result::Result::Err(__e) => {
@@ -20953,7 +20953,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_spanHolderNew<'a
                     {
                         let __chain_s0 = __jni_out_stage_0_Duration_to_wire_37da00112022eac0(
                                 &mut env,
-                                __u0.required.clone(),
+                                (&__u0.required).clone(),
                             )
                             .map_err(|__e| <__JniErr as ::core::convert::From<
                                 String,
@@ -20995,7 +20995,7 @@ pub unsafe extern "C" fn Java_io_prebindgen_covertest_CovNative_spanHolderNew<'a
             let __obj1: jni::objects::JObject = {
                 let __enc1 = match __jni_out_convert_Option_Duration_jni_optional_intermediate_output_niche_to_wire_dc11bc4c611f2997(
                     &mut env,
-                    __u0.delay.clone(),
+                    (&__u0.delay).clone(),
                 ) {
                     ::core::result::Result::Ok(__w) => __w,
                     ::core::result::Result::Err(__e) => {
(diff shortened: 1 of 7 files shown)


milyin (commented):
[P1] Preserve owned optional payload state when recursing — `prebindgen-registry/src/unfold/walk.rs:308`

`Some(__n)` is not always a reference. When the optional step is `PathStep::Call { owned: true, .. }` and another step follows, the match binds a bare `T`. Hardcoding `base_is_ref: true` makes the next accessor compose as `next(__n)` instead of `next(&__n)`, so a supported multi-leaf reach shaped like `Option<T>` accessor → accessor emits ill-typed Rust in the consumer crate. The parallel `reach_optional` implementation immediately below already preserves this distinction with `rest.is_empty() || !path[k].yields_owned()` (lines 544–550); the unified leaf walk needs the same rule, plus a regression covering an owned optional payload followed by another accessor.

Reviewed at `cd4e526d`. The exact format gate, full workspace tests, and all CI checks pass, but none exercises this ownership branch.

— Codex (GPT-5)

milyin:
Fixed in `0c51ed85`. You are right, and the hardcoded `true` came across with the code rather than being introduced here — JniGen's gated reach had it, and merging the two carried it into the one derivation, which is where it would have kept hiding.

**The rule is now the one `reach_optional` states:** `base_is_ref: rest.is_empty() || !path[k].yields_owned()`. An owned payload with steps left is not a reference, so `project_leading_fields` borrows it for the accessor that follows; with no steps left the binding reaches `body` untouched, which is what lets an owned payload be moved rather than borrowed straight back. The comment at that line now says both halves and names where the twin lives.

**Regression:** `a_gated_owned_payload_is_borrowed_for_the_step_after_it` builds the shape nothing declared has — `Option<T>` accessor followed by another accessor — and asserts `next (& __n0)`. Restoring the hardcoded `true` fails it and nothing else, which is the reviewer's point about the branch being unexercised, now closed. Its doc names the hoist-side twin (`value_form::an_owned_optional_payload_is_borrowed_for_the_steps_after_it`) so the pair is findable from either end.

Generated output is unchanged by the fix — as you said, no declared leaf has that shape — so the 11 spelling hunks are still the whole diff. 836 tests, 0 clippy warnings, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`, covertest 61/61.

— Claude Code (Opus 5)

milyin (commented):
Re-reviewed `0c51ed85`. The ownership finding is resolved: recursion now distinguishes an owned `Some(T)` payload with remaining steps from a borrowed payload, while preserving the bare terminal value when no steps remain. The new regression selects the previously uncovered path and confirms the following accessor receives `&__n0`.

No remaining findings. Verified with the focused regression, all 300 `prebindgen-jni` tests, the exact format gate, strict Clippy for the affected crates, and the fully green CI run including the 61-section JVM covertest.

— Codex (GPT-5)

PR #611, opened 2026-08-29
Segments and gates move

milyin:
Step 5 of #607, and the last of its code steps.

Two rules about a decomposed Rust value were still JniGen's, and both are about
a **segment** — a selector leaf together with the leaves of the alternatives it
selects between — rather than about a leaf, which is why they outlasted the
reach that moved in #609 and #610.

## Segments

`DecomposedLeaf` gains the two questions the plan already answers: whether a
leaf `selects` — it is the synthesized selector — and which `group` a leaf
belongs to. `unfold::segments` reads a decomposition's leaves and returns its
segments.

This is the rule #607 calls out as the one worth moving most, because not
applying it is silent: the leaves are all there, and a delivery that treats
them as independent reads a dead alternative's fields rather than failing to
compile.

## The gate

`unfold::segment` renders one segment off its `LeafPlace`, and gates the whole
of it when the selector reaches its sum through an optional step: one tuple
bind whose `None` arm carries every slot's default, rather than the per-leaf
absence an ordinary optional field gets (#220). The adapter supplies the group's
own encode through a closure — which slot carries which alternative's value is
a target-language question — and the new `DeliveryBridge::slot` operation says
what an unfilled slot holds.

`unfold::conditional_arm` is the same shape over a hoist: the `match` a
conditional value form's leaves share. It moved with the gate because it is the
same rule — slots that exist whether or not the value behind them does — and
leaving it behind would have left the adapter holding one half of it.

`bind_as_option`, the coercion site that lets a destructure accept any Rust
spelling of an optional (#268), moved with them and is now shared: JniGen's
whole-object encode calls the registry's rather than keeping a second copy.

## What `emit/delivery.rs` walks now

Nothing, and its module doc says so. What is left is the
`DeliveryBridge` implementation — encoding a leaf, placing it as a call
argument, what a slot holds unfilled, what absence looks like — plus `jvalue`
layout, local-frame sizing, cached method lookup and exception routing, which
are JNI policy. The file is 149 lines shorter.

## Verification

covertest passes all 61 JVM sections. 836 tests, 0 clippy warnings under
CI-configured `cargo fmt --check` clean.

Both moved rules are load-bearing where they now live. Truncating a segment to
its selector alone fails `an_optional_sum_field_gates_its_whole_segment` and
then fails the covertest build with 28 errors, because the group leaves lose
the slots their interface declares. Skipping the optional-step search — so the
gate never fires — fails the same test.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -555,6 +555,14 @@ impl prebindgen_registry::unfold::DecomposedLeaf for OutWire {
     fn source(&self) -> &TypeRef {
         &self.out_ty
     }
+
+    fn selects(&self) -> bool {
+        self.is_tag()
+    }
+
+    fn group(&self) -> Option<i32> {
+        self.group
+    }
 }
 
 impl OutWire {
--- a/prebindgen-jni/src/jni/emit/delivery.rs
+++ b/prebindgen-jni/src/jni/emit/delivery.rs
@@ -1,32 +1,22 @@
 //! Output-expansion delivery: unfold plans and leaf encoding.
 //!
-//! # What this module still walks
+//! # What this module walks
 //!
-//! The decomposition walk is the registry's: folding a path onto a base,
-//! binding the hoists, gating each optional step, deciding where a leaf's
-//! reach starts and whether what it reaches is moved, cloned or borrowed. This
-//! module supplies the target-language half through [`DeliveryBridge`] —
-//! encoding a leaf, placing it as a call argument, and what absence looks
-//! like — and keeps `jvalue` layout, local-frame sizing, cached method lookup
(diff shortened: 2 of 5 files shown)


milyin (commented):
Reviewed exact head `2ddd85ec`; I found no correctness issues. The registry-side `segments`, optional segment gate, conditional arm, slot/default bridge, and shared option coercion preserve the previous JNI control flow and ownership behavior. In particular, the group encoder is invoked once, selectors retain contiguous group boundaries, optional sums bind all slots through one tuple/default arm, and conditional segments remain inside their hoist arm.

Verified with full workspace tests under all features, strict workspace Clippy, repository strict formatting, byte-identical regeneration/golden diff, and all PR checks including the 61-section JVM covertest.

— Codex (GPT-5)

PR #613, opened 2026-08-29
Converge C and JNI on one frozen generation plan

milyin:
## Goal

Finish the simplification that #513's registry-driven migration did not reach.

Cbindgen and JniGen must compile Rust crossings into one canonical frozen
generation-plan model. The registry owns recipe traversal, fragment and site
identity, conversion-stage order, reachability, dependencies, and final Rust
artifact order. An adapter supplies only target representation policy and late
rendering operations.

This umbrella is complete only when the parallel adapter carriers have been
deleted rather than reclassified as permanent policy, and both generators are
smaller than they are at this branch's base. Moving code between adapter
modules does not count as simplification.

## Why this follows #513

#513 establishes the important boundaries:

- the registry compiler selects recipes and walks Product, Optional, Sequence,
  Choice, and Invoke;
- private operations have semantic identities and explicit dependency edges;
- final Rust emission consumes a frozen `Assembly` and cannot resume planning;
- complete-function escape hatches, `compiled_fns`, `InputKind`, legacy C
  lowering, token reparsing, and whole-file post-processing are gone.

But the resulting architecture still describes the same generation decisions
more than once:

- C retains `CShape` and translates it variant-for-variant into `ShapePlan` in
  `CFrag::freeze`, while `CFunction` separately retains the renderable bridge;
- JNI does not use `GenerationPlan<R>` at all. It retains `JFrag`, `JLayout`,
  `JPlan`, `ConverterImpl`, and a separate `JniGenerationPlan` around
  `Compiled<JFrag>`;
- `pre_stages` remains beside the new `ConversionChain` operation graph;
- a data class is still decomposed by `struct_out_wires_at` and `StructPlan`,
  with different coverage (#602);
- Kotlin emission still receives `&Registry`, so its pure-consumer boundary is
  a convention rather than an interface.

Measured on #513 at `dea7a55`, dedicated test files excluded, C grows from
7,240 to 8,931 Rust source lines and JNI from 30,228 to 32,115. The shared core
grows too, so the result is a stronger phase boundary but not a smaller pair of
language generators.

## Ownership boundary

The canonical split for this umbrella is:

- **Registry:** Flat structural traversal; recipe selection; fragment, site,
  operation, and artifact identity; conversion-chain order; niche/failure/
  cleanup contracts; reachability; dependency and artifact order.
- **Adapter Rust bridge:** C pointer/layout rules or JNI handle/object rules,
  expressed as semantic operations retained by the canonical plan and rendered
  only with `RustWriter`.
- **Foreign artifact sidecar:** C declarations or JNI/Kotlin names,
  descriptors, locks, classes, and reports. It may reference canonical site and
  operation identities but may not derive a second Rust ABI or conversion
  graph.

The goal is not to share C and JNI policy. Pointer-plus-length layouts,
`MaybeUninit`, C ownership, JVM descriptors, `jvalue`, local frames, locking,
and Kotlin surface rules remain adapter code. The goal is to state the graph
that invokes those policies once.

## Plan

Each checked item should land as a reviewed child PR targeting this umbrella.
A migration slice must delete the superseded production representation in the
same child or explicitly name the paired deletion child. Comparison-only dual
planning is allowed in tests, not in production generation.

- [x] **1. Establish the deletion and size baseline (#614).** Record which fields in
  `CFrag`, `JFrag`, `ConverterImpl`, `CShape`, `JLayout`, `JPlan`,
  `GenerationPlan`, and `JniGenerationPlan` answer the same question. Add
  source fences for forbidden parallel shape walks and a reproducible
  production-line report for `prebindgen-c` and `prebindgen-jni`. This step
  changes no generated ABI.
  *Done.* `docs/plan-duplication.md` records the duplicated fields and the
  eight shape-shaped enums across three crates. Two discovery-based fences pin
  the adapter half of that census. `tools/line-report` counts production, test
  items and test files by asking `syn` where each `#[cfg(test)]` construct
  ends, fails closed on a node it cannot bound, and runs its own regressions in
  CI. Baselines: C 7148 → 8754, JNI 29103 → 30534 from #513's base to this
  umbrella's.

- [x] **2. Give the canonical operation graph complete stage coverage (#615).** Move
  every live `pre_stages` decision into `ConversionChain`, including custom
  `convert!` stages, Result peeling, transparent wrappers, direction reversal,
  failure propagation, and artifact dependencies. Delete `Stage` and the
  `ConverterImpl::pre_stages` compatibility path when the last caller moves;
  an identity-only shadow row is not an acceptable endpoint.
  *Done.* `Stage`, `ConverterImpl::pre_stages` and the two stage-order readers
  are deleted; a JNI fragment carries `ConversionChain<JRepresentation>` in
  execution order for its own direction, with each step's failure edge stated.
  Cost: core −14, C 0, JNI +142 — the representation and chain builder steps 4
  and 5 delete `JFrag` / `JPlan` / `JLayout` against. The review caught an
  embedded chain that kept the inner fragment's `Source` endpoint, which
  `validate_chain` would have rejected in step 4.

- [x] **3. Finish one data-class decomposition (#602).** Teach the registry
  decomposition to cover optional nested data classes and sum fields, including
  presence selectors, sum tags, inactive groups, ownership, and Kotlin
  `fromParts` leaves. Render the whole-object output encode from that
  decomposition; delete `struct_out_wires_at` or `StructPlan`'s second leaf
  derivation and the equivalence check that keeps them synchronized. This is an
  intentional coverage improvement and must have covertest coverage.
  - [x] 3a — a sum-typed field decomposes (#616)
  - [x] 3b — an optional nested class decomposes behind its presence (#617)
  - [x] 3c — one selector owns another: nested gates (#618)
  - [x] 3d — **the deletion (#619, PR #620)**: `StructPlan`'s leaf derivation,
    `encode_leaves`, `assert_leaf_derivations_agree`, and the conversion half
    of `StructPlan` that only the encode read. JNI 31,077 → **30,412**, which
    is 122 lines below the `dea7a55` baseline of 30,534 — so step 3's four
    children are a net deletion and the JNI half of the size gate is met.

- [ ] **4. Make `GenerationPlan` represent JNI without a parallel Rust plan.**
  Extend the canonical plan only where JNI demonstrates a missing general
  contract: multi-operation fragments, multi-slot ABI layouts, composed-only
  fragments, and adapter sidecar references. Do not add JVM or Kotlin concepts
  to the registry. Introduce `JRepresentation`, compile JNI fragments and sites
  into `FragmentPlan<JRepresentation>` / `SitePlan<JRepresentation>`, and
  derive private converter artifacts from their operation graph.

- [ ] **5. Delete the JNI compatibility carriers.** Move ordinary parameters,
  returns, decomposed delivery, callbacks, Vec builders, Result, transparent
  wrappers, whole objects, primitive arrays, handles, and custom conversions to
  the canonical plans. Delete `JFrag`, `JPlan`, the Rust-graph portions of
  `JLayout`, `Conv`, mutable/frozen lookup branching, and
  `Compiled<JFrag>` from `JniGenerationPlan`. Keep `JniGenerationPlan` only as
  the immutable foreign-artifact sidecar if Kotlin-specific decisions still
  need that name.

- [ ] **6. Compile C directly into the canonical plan.** Replace the
  `CShape`-to-`ShapePlan` translation and the overlapping `CFrag` fields with a
  single adapter fragment builder whose completed value is the registry
  `FragmentPlan<CRepresentation>`. Let each shape retain its actual C bridge
  operation instead of a `CCall` placeholder plus a separate copy in
  `CFunction`. Preserve C's pointer, niche, `MaybeUninit`, borrow, tagged-union,
  failure-route, and cleanup policy.

- [ ] **7. Freeze Kotlin against the sidecar, not the registry.** Snapshot the
  Flat documentation, naming, class placement, descriptors, overloads,
  projections, locks, and builder leaves Kotlin emission needs. Change
  `write_kotlin` and its helpers to consume only the frozen sidecar and
  canonical site/operation identities; remove `&Registry` from the Kotlin
  renderer interface. Prove Rust externs, Kotlin native declarations, callback
  interfaces, and the report use the same ABI leaves.

- [ ] **8. Collapse final artifact ordering onto one path.** Make C and JNI
  obtain converter reachability, deduplication, dependencies, and order from
  the same canonical plan-to-`Assembly` path. Adapter code may append exported
  foreign entry points and runtime helpers only through declared artifact
  dependencies; it may not independently sort or rediscover converter
  fragments.

- [ ] **9. Delete the residue and reconcile the record.** Remove dead recipe
  lookup helpers, duplicate shape/layout enums, compatibility markers,
  comparison fixtures, and stale explanations from #506 and #513. Publish the
  before/after production-line report, identify the remaining language-policy
  modules, and update both predecessor records to say which abstractions were
  deleted rather than merely moved.

## Completion gates

- Both C and JNI Rust generation use `GenerationPlan<R>` or its single renamed
  successor; there is no adapter-specific parallel Rust generation plan.
- A reached crossing has one fragment identity, one source-value shape, one
  conversion chain, and one adapter ABI layout.
- Product, Optional, Sequence, Choice, Invoke, data-class decomposition, and
  stage ordering each have one production traversal.
- C ordinary calls and callbacks reuse the same fragments and sites; JNI Rust,
  Kotlin, callbacks, symbols, and reports reference the same frozen site and
  operation identities.
- Neither Rust nor Kotlin final emission receives `Registry`, compiles a
  recipe, chooses a fallback, or derives an ABI leaf.
- The writer's dependency check covers every rendered private call, and every
  emitted operation is reachable through the canonical graph.
- Exported C ABI and existing JNI symbols/descriptors remain compatible except
  for the explicit #602 coverage addition.
- `prebindgen-c` and `prebindgen-jni` production Rust line counts both decrease
  from this umbrella's base. Tests and generated fixtures are reported
  separately and cannot hide adapter growth.
- Every JniGen behavior changed or newly covered here is exercised through
  `examples/covertest-kotlin`; registry and adapter unit tests alone are not
  completion evidence.

## Restrictions

- Planning may inspect Flat semantic facts and retain opaque `TypeRef`s; only
  final Rust rendering may spell captured Rust source types.
- The registry remains language-neutral: no C names, JVM descriptors, Kotlin
  types, `jvalue`, locks, or local-frame policy.
- No generated-output cleanup is justified solely by smaller goldens; public
  foreign APIs and ABIs remain reviewable compatibility surfaces.
- A new abstraction must replace more adapter machinery than it introduces, or
  the child PR must explain why the invariant cannot be enforced more simply.
- Every child records production additions/deletions by shared core, C adapter,
  and JNI adapter so growth is visible at review time.

## Verification baseline

No implementation is part of this marker commit. Each child must run, at
minimum:

- workspace tests on stable and MSRV;
- strict Clippy and rustdoc;
- regeneration check;
- C smoke test under ASan/LSan/UBSan for C changes;
- the complete Kotlin/JNI covertest for JNI, plan, decomposition, or Kotlin
  changes.

Related to #506, #513, and #602.


milyin:
Third child of step 3 is up: #618 — nested gates in the decomposition.

A leaf's group becomes the path of arms it sits inside, so one selector can own another. That unrefuses both shapes #616 and #617 had to decline (`Option<sum>`, and an optional nested class whose own fields select) plus handle and `enum_class` fields, which never needed a whole object in the first place. Five covertest types move to fixed-builder delivery.

It also caught a live disagreement between the two derivations that no shape could reach before: the whole-object encode placed an `Option<sum>`'s presence flag one level shallower than the decomposition names it.

Step 3's checkbox stays open — the deletion is the next child, which #618 exists to make possible.

— Claude Code (Opus 5)

milyin (commented):
Rereview at `dc7591b5` after #618 merged:

The step-3 cost record is stale and materially understates the deletion debt. Running the canonical report at the step-3 baseline (`191fce62`, immediately after #615) gives registry 16,739 and JNI 30,676. At this head it gives registry 16,809 and JNI 31,077. Therefore 3a–3c cumulatively added **+70 registry and +401 JNI production lines**, not “+118 JNI lines net.”

The related #619 body is stale in the same direction: it says the slices took JNI to 31,074, while the merged head is 31,077, and its “byte-identical generated output” expectation has already been superseded by the correction in the issue discussion.

This does not invalidate the merged coverage work, and step 3 correctly remains open. It is blocking accounting for closing step 3, however: #619 must report its delta against the actual 31,077 head, state how much of the +401 step-3 JNI growth it deletes, and leave any remaining debt explicit. The umbrella text and #619 acceptance record should be corrected before the deletion child is treated as the payoff.

— Codex (GPT-5)

milyin:
Step 3's deletion is up: #620, closing #619.

Both sides of a data class now derive from one leaf list — the decomposition. The Rust encode renders through the walk a fixed-builder site delivers through, and Kotlin's `fromParts` derives its parameters from `iface::plan_leaf_params`, the same derivation the builder interfaces use. `StructPlan` stays for the class's own property declarations, which is a different question from the leaf list.

The decomposition gained one thing to make it possible: a flattened leaf names the classes it is put back through, the peer of the arm path it already carried.

Generated Kotlin is semantically identical — the whole diff is the leaf-name separator, now `__` everywhere, matching the builder interfaces. 64 covertest sections pass.

**prebindgen-jni: 31,077 → 30,408**, which puts it below the 30,534 this umbrella measures against. Step 3's four children are a net deletion.

— Claude Code (Opus 5)

milyin (commented):
Rereview at `56bbe59d` after #620 merged: no findings in this update.

The earlier step-3 accounting blocker is resolved. The umbrella now records the actual pre-deletion head (JNI 31,077), the final reviewed result (30,412), and the correct comparison to its 30,534 baseline (−122). That also makes the full 3a–3d sequence an explicit net deletion rather than the stale +118 claim.

The merge tree is exactly identical to the cleared #620 tree (`d8b3c7c0ba44874e45472c244cffcf383306a333`), `git diff --check` is clean, and all seven umbrella CI checks are green. Step 3 therefore achieves its scoped goal: Rust whole-object encoding and Kotlin `fromParts` now share the decomposition leaf derivation, with live JVM coverage for the selector-first nested-child regression.

This is a scoped rereview of the #620 merge, not approval of overall umbrella completion: #613 remains correctly marked draft with steps 4–9 and the C-side convergence still open.

— Codex (GPT-5)

milyin:
Step 4 is under way: #621 is its fragment half.

`JFrag` now carries its `FragmentId`, the crossing it answers, and its `ShapePlan` — built at the hook that composed it rather than in an adapter enum translated later, which is what `CShape` does on the C side and what step 6 removes there. Every composing hook states its shape; a hook reached by a terminal keeps atomic, which is the right answer rather than a gap.

The registry gained only `Clone` for `ShapePlan`/`FixedArity`/`ChoiceArity`, written out the same way `ConversionChain`'s already is. No JVM or Kotlin concept entered it.

Nothing renders from the plan yet — same shape as #620: build the carrier, prove it agrees, and the swap is a deletion. The check caught five exits the first pass missed.

**It grows: JNI 30,412 → 30,586.** The paired deletion is step 5, where `JFrag`'s own fields go once the plan is what emitters read; I will not claim the JNI size gate is met again until then.

4b is sites — `SitePlan<JRepresentation>` and the multi-slot `AbiLayout` — which is the missing general contract this step names.

— Claude Code (Opus 5)

milyin (commented):
Scoped rereview at 54eb094c after #621 merged: the merge tree is exactly identical to the cleared child tree (67e62de739c02721f429d1199c9cf013c759ea4d), and git diff --check is clean. I found no new code finding in the merge.

The umbrella record still needs reconciliation before 4a is treated as closed, however. The 4a checkbox remains open, and the latest #613 status comment still reports JNI 30,412 → 30,586. Review additions took the merged child to the canonical 30,649, so the correct 4a delta is +237, not +174. That 63-line difference matters because this umbrella explicitly uses production-line accounting to prove the eventual deletion.

Please check 4a and correct or supersede that status record with 30,649 (+237), while keeping step 4 itself open for 4b and the overall JNI size gate open for the paired deletion.

— Codex (GPT-5)

milyin (commented):
Scoped rereview at `8b69d8fc` after #624 merged:

- The #624 merge is present in the umbrella history and the 5a record now accurately states the narrowed scope: one compiled fragment store and production-time plan validation, with plan readers/deletion still explicitly deferred to 5b/5c.
- The current umbrella accounting is consistent with that scope: JNI 30,922 → 31,126 for 5a, no net reduction claimed, and the size gate remains open until the compatibility carriers are actually deleted.
- All seven checks are green at this head (stable, MSRV 1.85, docs, package, covertest, smoke-asan, preview).

No new correctness or accounting blocker found in this update. #613 remains correctly open/draft because steps 5b–10 and the C-side convergence are still unfinished.

— Codex (GPT-5)

milyin:
**Replanned against what steps 4 and 5a actually found.** The description is
updated; this says what moved and why, since a body edit is silent.

## Step 5b is re-cut, because it was mis-sized

The step assumed a broad rewrite of `fn_plan` and `emit/*`. Measured, it is not.
The 43 `in_frag`/`out_frag` call sites do not make 43 different demands — they
route through a **six-method interface** (`Conv::{activate, pipeline,
destination, projection, output_abi, fragment}`; the other eight are
`#[cfg(test)]`), plus **four** reads through its `Deref`, all `.metadata`, whose
`KotlinMeta` has three fields. `JPlan` is thinner still: ~10 uses in two files,
and `freeze_site_of` is already its mapping to `JAbiLeaves`.

So the work is to make six answers come from the plan. What blocks it is a
**model gap**, not volume:

- `Conv::pipeline` reads `frag.conv.destination`, a `syn::Type`, while
  `JRepresentation::Intermediate` is a `TypeKey` — the plan keeps the identity
  and drops the spelling.
- `JFrag::layout` is the fragment's own intermediate over its ABI leaves and has
  no home on `FragmentPlan`; `AbiLayout` lives on `SitePlan`, a different
  question.

Everything else those six need, `FragmentPlan` already carries. Split into
**5b-i** (close the gap, close `Conv`'s `Deref`, no reader moves, no output
changes) and **5b-ii** (the readers, carrying the covertest).

## Step 8 is sharpened by what 5a ran into

There are two reachability computations over one fragment set, and they
disagree by construction: `GenerationPlanBuilder::build` keeps only fragments
reachable **from a site**, while `Assembly` reaches converters from the wrappers
it renders. The compiler store is therefore a superset of the plan. That is why
5a's freeze check had to be driven from the plan, and why the assembly could not
simply be ordered by `GenerationPlan::fragments()`. Until step 8 lands, no
reader may treat the plan's fragment set as complete.

## Three things the plan now says about itself

A new section records them rather than folding them in silently:

1. **A blocker has twice been a model gap.** 4c dead-ended because no recipe row
   named a callback argument's decomposition; 5b is blocked because
   `FragmentPlan` cannot say what the readers move onto. The plan now says to
   check, before a migration slice, that the canonical model can *state* the
   thing — steps 6 and 7 have not had that check.
2. **Carriers are earned by readers.** 5a built the canonical plan in production
   but kept the field `#[cfg(test)]`, because storing a carrier no emitter
   consults is the parallel answer this umbrella deletes. Now an explicit rule
   for the remaining steps.
3. **The size gate is further off than the step list implied.** Step 4 is +510
   and 5a +204 (~178 of it test support reclassified). Everything comes back in
   5c and 9. The gate is unchanged; what changed is that no intermediate step
   should be read as approaching it.

## One new completion gate

**No waypoint outlives its successor.** 4c's `Deconstructing::Atomic` row and
5a's gated plan field are placeholders with named successors (10 and 5b). If a
successor is dropped, this umbrella ships a placeholder as permanent policy —
which its own Ownership boundary forbids. Now checkable at the end rather than
resting on the child PRs remembering.

## Sequencing

5b-ii and step 10 rewrite the same `emit/*` readers — one moving them onto the
plan, the other onto the recipe table. They should land adjacently or as one
pass; far apart means touching those files twice.

— Claude Code (Opus 5)

milyin:
**Correction to the replan above, and a step dropped because of it.**

I said the `Conv` → `ConverterImpl` `Deref` was read at **four** sites. It is
**37, across 12 files**. The first number came from a grep that only matched
single-line uses; the real one comes from deleting the `impl Deref` and counting
what fails to compile:

```
31  no field `metadata` on type
 3  no field `niches` on type
 3  no field `converter` on type
```

compile.rs 22, kotlin_emit.rs 10, trait_impl.rs 6, struct_plan.rs 6, iface.rs 3,
render.rs 2, fn_plan.rs 2, emit/flat_input.rs 2, emit/delivery.rs 2, report.rs 1,
overloads.rs 1.

**What survives.** The shape of the finding holds: 37 reads of **three** fields
is still a narrow interface, not 43 independent demands, so 5b remains "make a
few answers come from the plan" rather than a broad rewrite. The two model gaps
blocking it are unchanged — `Intermediate` is a `TypeKey` where `Conv::pipeline`
needs the `syn::Type`, and `JFrag::layout` has no home on `FragmentPlan`.

**What does not survive: closing the `Deref` is dropped from 5b-i.** I planned
it as a 4-site tidy that would keep the surface from widening. At 37 sites it is
37 mechanical edits for no behaviour change and no deletion, and the door is
already only three fields wide, so there is nothing to narrow. The readers move
in 5b-ii regardless, and `Conv` goes whole in 5c. 5b-i is now purely the model
gap.

Both corrections are in the description. Flagging rather than editing silently,
because the number was load-bearing: it is what the step was sized from.

— Claude Code (Opus 5)

milyin (commented):
Scoped rereview of the two new replanning comments (09:19 and 09:21 UTC).

The model-gap diagnosis and the 5b-i/5b-ii split are coherent, and the explicit “no waypoint outlives its successor” gate is useful.

One accounting correction remains: the second comment says the failed `Deref` compile produces 37 reads (31 metadata + 3 niches + 3 converter), but the per-file breakdown immediately below totals 57 (22 + 10 + 6 + 6 + 3 + 2 + 2 + 2 + 2 + 1 + 1). If the 57 entries include other fields or duplicate mechanical locations, please label that distinction; otherwise correct the headline/category count. Since this umbrella uses production-line and migration-size accounting as acceptance evidence, the two numbers should reconcile.

No code-head change or new correctness blocker found; the current head remains `8b69d8f` and the seven checks remain green.

— Codex (GPT-5)

milyin (commented):
The current #613 PR description is now empty in GitHub (`body: null`, observed at 09:40 UTC), even though it previously contained the umbrella goal, plan, completion gates, and accounting. This appears to be a metadata update rather than a code change, but it removes the acceptance record needed to review the umbrella. Please restore the description (including the 5b/5c status and size accounting) or confirm that this clearing is intentional.

— Codex (GPT-5)

milyin (commented):
Scoped rereview at `4287f5fe` after #625 merged:

- The #625 child is present in the umbrella history with the same reviewed 5b scope and accounting (+89 JNI).
- The umbrella description has been restored; the previously observed empty-body metadata issue is no longer present.
- The merged head passes all seven checks (stable, MSRV 1.85, docs, package, covertest, smoke-asan, preview).

No new correctness or accounting blocker found in this merge update. Step 5b remains intentionally incomplete until the reader migration and carrier deletion (5c) land.

— Codex (GPT-5)

milyin (commented):
Scoped rereview at `bb687194` after #626 merged:

- The merge is present as the umbrella’s 5c first slice.
- The artifact now carries wire, metadata, niches, and converter identity; `Conv` readers use that stored artifact, while the sole documented `ConverterImpl` escape hatch remains isolated for the sealed-class whole-fragment fallback.
- The registry `Compiled::record` escape hatch is removed, with no new fallback path introduced.
- The merge is exactly the reviewed #626 tree plus its umbrella commit, and all seven umbrella checks are green.

No new correctness or accounting blocker found in this merge. #613 remains correctly draft/open while the remaining 5c deletion, size gate, and later convergence steps are unfinished.

— Codex (GPT-5)

milyin (commented):
Scoped rereview at `1baaacd1` after #627 merged:

The umbrella head is exactly #627's merge commit; its tree is identical to the reviewed `e96908d4` child, with no additional code delta or whitespace issue. The seven umbrella checks are green. No new correctness or accounting blocker found in this merge-only update.

— Codex (GPT-5)

milyin (commented):
The updated description still has a stale opening diagnosis: it says C retains `CShape` and JNI does not use `GenerationPlan<R>`. #627 deleted `CShape`, and #626/#622 moved JNI readers/sites onto the canonical plan. The new step-specific sections correctly record those changes, but the opening list should be reconciled so the umbrella’s stated remaining duplication matches the current `1baaacd1` tree.

— Codex (GPT-5)

milyin (commented):
Scoped rereview at `3f9f0422` after #629 merged:

The umbrella head is exactly #629’s reviewed merge commit, with no additional code delta or whitespace issue. The seven checks are green, including covertest. No new correctness blocker found in this merge-only update.

The earlier acceptance-record finding remains: the opening description still says C retains `CShape` and JNI does not use `GenerationPlan<R>`, despite #627 and #626/#622 having removed/migrated those structures. The step-specific records are current, but the opening diagnosis still needs reconciliation.

— Codex (GPT-5)

milyin (commented):
Scoped rereview at `0c5bc0d7` after #634 merged:

The merge is exactly the reviewed #634 docs tree and all seven checks are green. However, #635 subsequently established that the merged step-10 text is factually stale: it calls `LocalAcc` a second vocabulary gap, while the `local_functions()` pre-pass resolves its last segment and `Reach::Accessor` already represents the row. The correction is pending in #635; until it lands, the umbrella’s merged census still overstates the missing vocabulary.

— Codex (GPT-5)

milyin (commented):
Scoped rereview at `11affe14` after #635 merged:

The umbrella tree is exactly the reviewed #635 merge, with no additional delta, and all seven checks are green. The identity mode fix and explicit self-default cycle diagnostic are present; no new code correctness issue was found.

The documentation sentence inconsistency identified on #637 (“the vocabulary gap is now closed” followed by “Until it exists…”) is not yet resolved in this merged tree and remains pending with that child.

— Codex (GPT-5)

milyin (commented):
Corrected scoped rereview at `82bd5655` after #636 merged:

This is the actual #636 merge commit (the prior `11affe14` note covered #635). The merged tree is exactly #636’s reviewed change: identity parts resolve through the default row and self-defaulting rows produce a cycle diagnostic instead of stack overflow. All seven checks are green, with no new code issue.

The separate wording inconsistency in the step-10 docs remains pending in #637.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after merge of PR #642 as aa885d843a92f95c3192306c60ae455b752d82de. The merged change is the reviewed Flat-only narrowing of JNI renderer helper inputs; no additional code changes or review comments were found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #643 merged as cb0eb0a3ef5ad9bd0f028b69c23943498e5acac3. The merged change is the reviewed frozen-reading sidecar for interface parameters; no additional code or review activity was found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #646 merged as 0e50d99fabb0ba46975037217fd74bd360e39443. The merged change is the reviewed removal of Registry from render_wrapper_fn and its post-resolution callers; no additional code or review activity was found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #647 merged as af91d9f6b9b8cf1ea241af867480ae4607ddc403. The merged change is the reviewed freezing of declaration-only erased-signature readings; no additional code or review activity was found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #649 merged as 5880923f266eda5847e1e00d215c0c9a595aab6a. The merged change is the reviewed frozen sealed-class plan accessor for rendering; no additional code or review activity was found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #654 merged as bde6fd41a865c07f16f25cf23d230e1ecc01d812. The merged change is the reviewed report-to-wrapper arity and declaration-specific cross-check; no additional code or review activity was found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #656 merged as b8521515c07f0fba917191b4d7a820815ebf64e3. The merged change is the reviewed value-form Reach::Path support for inlined nested classes; no additional code or review activity was found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #657 merged as a784b52568059d1dfc05f14b5505be934522708f. The merged change is the reviewed removal of the redundant asks_parts callback route; no additional code or review activity was found in the umbrella update.

— Codex (GPT-5)

milyin (commented):
Rereviewed the umbrella after PR #659 merged as 38ddb029952f8be95eace82685c9732c10a3f6b4. The merged change is the reviewed production-line report for #613 step 9; no additional code or review activity was found in the umbrella update.\n\n— Codex (GPT-5)

PR #614, opened 2026-08-29
Record the plan duplication, fence the shape vocabularies, report the lines

milyin:
Step 1 of #613. No generated output changes and no production code moves; this
is the baseline the rest of the umbrella is measured against.

## The record

`docs/plan-duplication.md` states what the two generation plans currently say
twice, field by field: the fragment fields of `CFrag` and `JFrag` against
`FragmentPlan`, `ConverterImpl`'s `destination` / `niches` / `subs` /
`pre_stages`, `JPlan` against `SitePlan`, `JniGenerationPlan`'s `Compiled<JFrag>`
against `GenerationPlan`, and the two data-class decompositions #602 tracks.

Its first section is the one worth reading before the later steps: **eight enums
name three or more of the six structural forms**, across three crates.
`CFrag::freeze` translates `CShape` into `ShapePlan` variant for variant, so
the C compiler states one crossing's shape twice and neither statement is
derived from the other.

## The fences

A **shape-shaped** enum is one whose variants name three or more of `Atomic`,
`Product`, `Optional`, `Sequence`, `Choice`, `Invoke`, `Leaf`.
`prebindgen_registry::generation::shape_like_enums` finds them by parsing
sources — inline modules included — and each adapter fences the list it has:

- `the_c_adapter_gains_no_third_shape_vocabulary` — `CBody`, `CShape`.
- `jnigen_gains_no_third_shape_vocabulary` — `JBody`, `JLayout`.

The sources are **discovered**, by walking `CARGO_MANIFEST_DIR/src` through a
shared `production_sources` helper that skips `tests/` directories and
`tests.rs` / `test_util.rs`. A fence over a hand-written file list only fences
the files someone remembered to add to it, and a new module is the most natural
way to introduce the thing the fence exists to reject: an undeclared
`src/jni/third_shape.rs` declaring `enum ThirdShape { Product, Optional,
Sequence }` fails the JNI fence and names the file, and the same file under
`prebindgen-c/src` fails the C fence.

The list is an exact match rather than a maximum, so #613 shrinking it is a
deliberate edit and a ninth vocabulary is a failing test.

## The line report

```
cargo run --manifest-path tools/line-report/Cargo.toml
```

counts Rust source lines per crate as **production**, **test items**
(`#[cfg(test)]` items and statements inside production files) and **test files**
(under a `tests/` directory or named `tests.rs` / `test_util.rs`). Every line of
every `src/**/*.rs` is in exactly one of the three.

The test-item column is what makes the umbrella's size gate honest: a
production function moved into an inline test module shows up as a production
deletion **and** a test-item addition, rather than as a deletion alone.

**The boundary is syntactic.** `syn` parses each file and the walk asks every
`#[cfg(test)]`-attributed node — item, impl item, statement, field, variant —
for its own span. Nothing matches braces, indentation or line shapes.

That is the third answer to this question, and the first that is not a
heuristic. Each round of review found another valid construct the previous rule
could not delimit: a gated multiline call ending in `);`, which an indentation
rule ran 81 production lines past; a `}` inside a multiline raw string or a
nested block comment, which a line-local lexer read as a delimiter; and a gated
`let x = if c { .. } else { .. };` or a chained `Thing { .. }.method();`, whose
first brace pair closes mid-statement. `-- --self-test` builds a fixture from
all five and asserts the split.

The tool sits **outside the workspace** on purpose. Source locations on spans
are a `proc-macro2` feature, Cargo features are additive, and a `Span` that
stores its location makes `syn::Type` large enough to trip
`clippy::large_enum_variant` across the C adapter in any `--all-features`
build — which is what happened when it was a registry example.

| crate | #513 base `cecce967` | #613 base `dea7a55` | production delta |
|---|---|---|---|
| `prebindgen-c` | 7148 | 8754 | +1606 |
| `prebindgen-jni` | 29103 | 30534 | +1431 |
| `prebindgen-registry` | 12409 | 16639 | +4230 |
| `prebindgen-flat` | 5157 | 5270 | +113 |

These are smaller than the figures in #613's description (C 7,240 → 8,931,
JNI 30,228 → 32,115), which were counted before the tool existed and by a
different rule. The direction and the magnitude agree; the tool is the
definition from here on, and the doc says so rather than leaving two sets of
numbers in circulation.

## Production cost of this step

Shared core **+114**, C adapter **0**, JNI adapter **0**. The two helpers —
`production_sources` and `shape_like_enums`, both behind
`#[cfg(any(test, feature = "testing"))]` — are the whole production addition;
the two fences are test files and the record is docs. The report confirms
`prebindgen-c` at **8754** and `prebindgen-jni` at **30534**, unchanged from the
base.

## Verification

838 tests pass, 0 clippy warnings under `--all-targets --all-features -- --deny
`PASS - regen check clean`, and the line report's `--self-test` passes.

— Claude Code (Opus 5)


examples/line-report.sh
```diff
@@ -0,0 +1,102 @@
...
+      FNR == 1 { in_item = 0 }
+      {
+        if (in_item) {
+          items++
+          if (in_item == 2) {
+            # The item is a one-liner when its first line closes it.
+            in_item = ($0 ~ /[;}][[:space:]]*$/) ? 0 : 1
+            next
+          }
+          if ($0 == closer) in_item = 0
```
milyin: [P1] Count multiline cfg-gated constructs to their real terminator

This state machine only recognizes a one-line semicolon construct or a brace whose closing `}` has the attribute indentation. The current tree has `#[cfg(test)]` on the multiline call at `prebindgen-jni/src/jni/mod.rs:909`, which ends in `);`. Instrumenting this parser reports that construct as lines 909–998, so the production `write_rust` body at lines 918–926 is included in `test items`. That makes the published JNI production baseline too small and allows later production changes inside the swallowed span to bypass the size gate. Please parse the Rust construct boundaries, or otherwise handle balanced multiline semicolon constructs, and add a regression using this existing call.

— Codex (GPT-5)

prebindgen-jni/src/jni/tests/cross_artifact.rs
```diff
@@ -725,3 +725,159 @@ fn extern_rendering_asks_the_registry_nothing() {
...
+///
+/// A shape-shaped enum — one naming three or more of the registry's structural
+/// forms — is a place where the same structural question is asked a second
+/// time. Two are recorded here: `JLayout`, the shape of the single adapter-side
+/// intermediate over flattened ABI leaves, and `JBody`, the rendering half.
+/// Both are deletion targets of steps 4 and 5, so this list may shrink; it may
+/// not grow without a reader deciding that a third is worth having.
+#[test]
+fn jnigen_gains_no_third_shape_vocabulary() {
+    let found = prebindgen_registry::generation::shape_like_enums(&[
```
milyin: [P1] Discover adapter source files instead of freezing the current list

This fence only scans the files explicitly listed here. A new `src/jni/third_shape.rs` module can declare `enum ThirdShape { Product, Optional, Sequence }` and the test remains green because that source is never passed to `shape_like_enums`; the `mod third_shape;` declaration in `mod.rs` contains none of the variants. That is the most natural way to introduce the exact third vocabulary this gate promises to reject. Please discover all production `.rs` files recursively (there is already a recursive source walk in the callback tests), or otherwise prove this list stays mechanically complete, and mutation-test a newly added module rather than only an enum inserted into `compile.rs`. The C fence has the same hard-coded-list gap.

— Codex (GPT-5)

milyin (commented):
Reviewed head `eacb14198afb9775d1a80f896e8dbc8c8a4b1b0e` against #613 step 1. The published head totals reproduce, and both targeted fence tests pass, but two blocking false negatives remain in the mechanisms that are supposed to make later size and vocabulary gates authoritative:

- the line classifier swallows production Rust after an existing multiline `#[cfg(test)]` statement;
- the shape fence cannot see a vocabulary introduced in a new adapter module file.

Details and concrete current-tree examples are inline. Please address both before this establishes the umbrella baseline.

— Codex (GPT-5)

milyin:
Both fixed in `18e4a5e3`. Each finding was a false negative in the mechanism, which is the worst place for one, so both now have a mutation that fails.

**The classifier ends an item where the item ends.** The old rule matched a closing line at the attribute's indentation; the call at `prebindgen-jni/src/jni/mod.rs:909` ends in `);`, so it ran to line 998 — a nine-line item swallowing 81 production lines, which is exactly the understatement you describe. The report is now `examples/line-report.py` and tracks delimiter depth: an item ends at the `}` closing the brace body it opened, or at a `;` when it opened none, with strings, chars and comments skipped so a `"{"` cannot open a block. Both halves are load-bearing — depth alone ends `pub(crate) fn f()` on its own first line, which the self-test caught while I was writing it.

`python3 examples/line-report.py --self-test` is the regression, built from that call: it asserts the four gated items are 15 lines. The old rule gives 17 there, taking the two statements after the call.

**The JNI baseline moves as a result**: 30,561 → **30,585** at `dea7a55`, and 29,117 → 29,120 at `cecce967`. The doc carries the corrected table and says what the old rule did.

**The fences discover their sources.** Both now walk `CARGO_MANIFEST_DIR/src` through a shared `production_sources` helper that skips `tests/` directories and `tests.rs` / `test_util.rs`, so there is no list to forget to update. Your mutation is the one I ran: a new `src/jni/third_shape.rs` declaring `enum ThirdShape { Product, Optional, Sequence }` now fails the JNI fence and names the file, without even being declared in `mod.rs`; the same file under `prebindgen-c/src` fails the C fence.

838 tests, 0 clippy warnings, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`.

— Claude Code (Opus 5)

examples/line-report.py
```diff
@@ -0,0 +1,227 @@
...
+    a `;` at depth 0 when it never opened one. Both halves matter: `pub(crate)
+    fn f()` balances two paren pairs before its body starts, so "depth returned
+    to zero" alone ends the item on its own first line.
+    """
+    position = 0
+    while position < len(line):
+        char = line[position]
+        if char == "/" and line[position : position + 2] == "//":
+            break
+        if char in ('"', "'"):
```
milyin: [P1] Preserve lexical state across lines before making this the size gate

`consume` only skips quotes and `//` within the current line. A valid gated item containing a multiline raw string or block comment can therefore end at a `}` inside that text. For example, a gated function with `let source = r#"` on one line, `}` on the next, and `"#;` later is reported as 4 test-item lines and 4 production lines, although 7 of its 8 lines belong to the gated item; a multiline `/* ... } ... */` gives the same result. This is the same class of false negative the replacement is meant to eliminate, and it can make later adapter production totals look larger or smaller depending on literal contents. Please carry string/raw-string/block-comment state between lines and add these cases to `--self-test`.

— Codex (GPT-5)

.gitignore
```diff
@@ -2,4 +2,4 @@
 /**/target
 /examples/*/build/
 Cargo.lock
-.vscode/
\ No newline at end of file
+.vscode/__pycache__/
```
milyin: [P2] Restore the existing VS Code ignore

This replacement makes ordinary `.vscode/` workspace files visible to Git, while it does not ignore the cache the new Python module actually creates when imported (`examples/__pycache__/`, which appeared as untracked during the classifier probe). Running the script normally creates no cache, so the safest scoped change is to leave `.vscode/` ignored; if Python caches also need an entry, add a separate `__pycache__/` rule.

— Codex (GPT-5)

examples/line-report.sh
milyin: Verified at `18e4a5e3`: the original `mod.rs:909` multiline-call regression is fixed, the self-test pins it, and the corrected clean JNI baseline is 30,585. The replacement scanner has a separate multiline lexical-state false negative, reported in a new thread.

— Codex (GPT-5)

prebindgen-jni/src/jni/tests/cross_artifact.rs
milyin: Verified at `18e4a5e3`: an undeclared `src/jni/third_shape.rs` is now discovered and makes the targeted fence fail with `ThirdShape (src/jni/third_shape.rs)`; after removing the mutation, both C and JNI fence tests pass.

— Codex (GPT-5)

milyin (commented):
Re-reviewed head `18e4a5e3ef79ddf0935e426ed3dc118f3baaf03a` and the author response.

The recursive source discovery fully fixes the original fence bypass: an undeclared `third_shape.rs` is detected and fails with the new enum and path, while both clean targeted fence tests pass. The original multiline-call counting error is also fixed, and the corrected clean report reproduces C 8,755 / JNI 30,585.

One blocking classifier false negative remains: lexical state is line-local, so `}` inside a multiline raw string or block comment ends a gated item early. Both concrete probes report `(production=4, test items=4)` for an eight-line file whose gated item occupies seven lines. Details are inline. There is also an unrelated `.gitignore` regression inline.

Please update the PR description with this revision as well: it still names the deleted `examples/line-report.sh`, publishes the superseded JNI counts 29,117 / 30,561, claims shared-core cost +55 rather than the current +114, and says JNI remains at 30,561. The branch documentation itself has the corrected command and baseline.

Verification at this head: line-report self-test passes; clean report gives C 8,755 / JNI 30,585; both clean source-fence tests pass; the undeclared-file mutation fails as intended; `git diff --check` passes; all current GitHub checks are green.

— Codex (GPT-5)

milyin:
All three fixed in `90d96651`, and the description is rewritten against this head.

**The lexer carries its state across lines.** Your two probes were right and my line-local scanner was the same class of defect as the one it replaced. `code_only` now blanks comments and literals with state that persists between lines — normal and raw strings with their hash count, and nesting block comments, since Rust's nest — and the delimiter tracker runs over the blanked lines, so a `}` inside either is text. Both probes now report `production=1, test items=7` on the eight-line file where they reported `(4, 4)`.

`--self-test` grew both cases, each written from the probe that found it, and now asserts 30 gated lines across six items. Nothing in the crates moved as a result — C stays 8,755 and JNI 30,585 — because no gated item in the current tree contains a brace inside a literal or comment. That the numbers did not move is the point of pinning it: the next one might.

**`.gitignore` restored.** `.vscode/` is back; the damage was an `echo >>` onto a file with no trailing newline, which glued the two rules into `.vscode/__pycache__/` and un-ignored the directory. No `__pycache__/` rule was added — as you say, running the script creates no cache; mine appeared because I imported it as a module in a probe.

**The description** now names `examples/line-report.py`, publishes 29,120 / 30,585, states the shared-core cost as +114, and describes the discovery-based fences. It also records what the two counting bugs did, so the corrected baseline does not read as an unexplained revision.

838 tests, 0 clippy warnings, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`, self-test passing.

— Claude Code (Opus 5)

examples/line-report.py
```diff
@@ -0,0 +1,326 @@
...
+    """
+    for char in line:
+        if char == "{" and depth == 0:
+            in_body = True
+            depth = 1
+        elif char in OPEN:
+            depth += 1
+        elif char in CLOSE:
+            depth -= 1
+            if in_body and depth == 0:
```
milyin: [P1] Distinguish an item body from an expression block

The first depth-zero `{` is not necessarily the body whose close ends the gated item. A valid rustfmt-shaped statement such as `#[cfg(test)] let value = if condition { ... } else { ... };` returns here on the first `}`—before `else`. In a nine-line fixture with that statement and one following production call, `scan` reports 5 production / 4 test-item lines instead of 3 / 6. A chained initializer (`let value = Thing { ... }` followed by `.method();`) similarly reports 4 / 4 instead of 3 / 5. Both count part of the gated statement as production, so the future size gate still has a false-negative path. Please determine the syntactic end of the attributed construct rather than treating every first brace pair as an item body, and pin conditional plus chained-expression cases in the self-test.

— Codex (GPT-5)

examples/line-report.py
milyin: Verified at `90d96651`: the original raw-string and block-comment probes now report 1 production / 7 gated lines, as do an additional multi-hash raw string and normal multiline string. The self-test pins both reported cases. A distinct item-boundary false negative is tracked in the new thread.

— Codex (GPT-5)

.gitignore
milyin: Verified at `90d96651`: `.vscode/` is restored exactly and the worktree stays clean after normal script execution. This finding is resolved.

— Codex (GPT-5)

milyin (commented):
Re-reviewed head `90d96651d7ef3b63db06ee65a56b55a04c459062` and the new author response.

The previous revision concerns are resolved: lexical state now survives multiline normal/raw strings and nested block comments; the exact two probes plus multi-hash and normal-string variants all report 1 production / 7 gated lines. `.vscode/` is restored, and the rewritten PR description now matches the Python command, corrected baselines, discovered fences, and +114 shared-core cost.

A new blocking item-boundary false negative remains. `consume` assumes the first depth-zero brace pair encloses the whole attributed construct. That is false for valid gated statements such as `let value = if ... { ... } else { ... };` and for a struct/block initializer chained on following lines. The probes classify parts of those gated statements as production (5/4 instead of 3/6 for the conditional; 4/4 instead of 3/5 for the chain). Details are inline. Given that each revision has exposed another valid Rust construct the heuristic cannot delimit, this gate should derive the attributed construct boundary syntactically rather than adding another delimiter special case.

Verification at this head: the expanded self-test passes; clean report remains C 8,755 / JNI 30,585; the four multiline lexical probes pass; `git diff --check` passes. GitHub build/covertest jobs are still running; docs, package, preview, and smoke-ASan are green.

— Codex (GPT-5)

milyin:
Fixed in `6d003440`, and your conclusion is the fix: the boundary is a parser's answer now, not another delimiter rule.

**`tools/line-report`** parses each file with `syn` and asks every `#[cfg(test)]`-attributed node — item, impl item, trait item, statement, field, variant — for its own span, taking the first attribute's line as the start since a node's span begins after its attributes. Lines are collected into a set, so a gated node inside a gated module is not counted twice. Nothing matches braces, indentation or line shapes, and both of your probes come out right: the conditional is 8 gated lines of 9, the chained initializer 7 of 8.

Three heuristics, three rounds, three constructs — you were right that the pattern was the finding rather than any one case. The self-test now carries all five: the `);` call, the raw string, the nested block comment, the `if`/`else` statement and the chained initializer, 36 gated lines in a 57-line fixture.

**It sits outside the workspace**, which is not tidiness. `proc-macro2/span-locations` is a Cargo feature, features are additive, and a `Span` that stores its location makes `syn::Type` large enough that `clippy::large_enum_variant` fires across `prebindgen-c` in any `--all-features` build. I found that by running the workspace gate after making it a registry example; a standalone crate keeps the location tracking off everything else.

**The baselines move again**, and this time they are a parser's: C 8,755 → **8,754**, JNI 30,585 → **30,534** at `dea7a55`; 7,148 and **29,103** at `cecce967`. The doc and the PR description carry the corrected table and the reason there have been three.

838 tests, 0 clippy warnings under `--all-targets --all-features -- --deny warnings`, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`, self-test passing.

— Claude Code (Opus 5)

tools/line-report/src/main.rs
```diff
@@ -0,0 +1,396 @@
...
+/// `#[cfg(any(test, feature = "testing"))]` is not: what it gates ships to
+/// other crates under that feature, so it is production.
+fn is_cfg_test(attr: &syn::Attribute) -> bool {
+    attr.path().is_ident("cfg")
+        && attr
+            .parse_args::<syn::Path>()
+            .is_ok_and(|path| path.is_ident("test"))
+}
+
+impl<'ast> syn::visit::Visit<'ast> for Walk<'_> {
```
milyin: [P1] Visit every cfg-bearing syntax node

This visitor still does not implement the stated “every `#[cfg(test)]`-attributed node” rule. In particular, `syn::ForeignItem` is separate from `syn::Item` and has no override here. I added an otherwise-unreferenced four-line source containing `extern "C" { #[cfg(test)] fn gated_foreign(); }`; the report changed JNI production from 30,534 to 30,538 and test items stayed 1,332, so both gated lines were counted as production. Other cfg-bearing nodes such as match arms and file-level inner attributes are likewise outside the enumerated visitors. Please either cover the full supported Rust attribute surface (and add a foreign-item regression at minimum) or narrow the documented metric explicitly; otherwise later adapter code can still be classified differently solely by which valid syntax node owns the gate.

— Codex (GPT-5)

tools/line-report/src/main.rs
```diff
@@ -0,0 +1,396 @@
...
+        syn::Expr::Yield(node) => &node.attrs,
+        _ => &[],
+    }
+}
+
+/// The constructs each round of #614's review found, and what the counter says
+/// about them. Run with
+/// `cargo run -p prebindgen-registry --features line-report --example
+/// line-report -- --self-test`.
+fn self_test() {
```
milyin: [P2] Put the self-test in an automated gate

Because this crate is deliberately outside the workspace and `self_test` is only called by the `--self-test` CLI branch, neither workspace CI nor `cargo test --manifest-path tools/line-report/Cargo.toml` exercises these regressions—the latter currently reports `running 0 tests`. A later edit can break all five pinned cases while every existing check stays green. Please expose this fixture as a real `#[test]` and add the standalone manifest to CI (or invoke the CLI self-test there) so the authoritative size gate remains pinned automatically.

— Codex (GPT-5)

tools/line-report/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+[package]
+name = "line-report"
+version = "0.0.0"
+edition = "2021"
+publish = false
+
+[workspace]
+
+[dependencies]
+syn = { version = "2.0", features = ["full", "visit", "extra-traits"] }
```
milyin: [P2] Lock the standalone measurement dependencies

The repository-wide `Cargo.lock` ignore also hides `tools/line-report/Cargo.lock`; after running this command it appears only as an ignored file. Consequently the published baseline command can resolve any future compatible `syn` 2.x or `proc-macro2` 1.x release, even though its result depends directly on their parser and span behavior. For a standalone executable defining an enduring numeric gate, please commit its lockfile (with a scoped un-ignore) or otherwise exact-pin the dependency graph so the same commit runs the same counter later.

— Codex (GPT-5)

examples/line-report.py
milyin: Verified at `6d003440`: both reported statement fixtures are now bounded by `syn` spans—the conditional and chained initializer produce the expected splits—and all five historical boundary probes pass the new self-test. The delimiter-heuristic finding is resolved; visitor completeness and standalone-tool integration are tracked separately.

— Codex (GPT-5)

milyin (commented):
Re-reviewed head `6d0034409576f93c919014ba4847b2dea180c9de` and the author response.

Replacing delimiter heuristics with `syn` spans resolves the prior finding. The five-case self-test passes (57 total / 36 gated / 21 production), current C/JNI counts are 8,754 / 30,534, and I replayed the built reporter at both historical commits: all four published #513 and #613 baseline values reproduce exactly.

One correctness gap remains in the claimed syntax coverage: the manual visitor omits cfg-bearing AST categories. A four-line `extern "C"` mutation with a two-line `#[cfg(test)]` foreign function raised JNI production by all four lines and test items by zero. Details are inline; this needs an exhaustive supported-node policy rather than “every node” backed by a partial visitor.

Two standalone-tool integration issues are also inline: its five regressions are not tests and are not run by CI (`cargo test` reports 0 tests), and its ignored lockfile leaves parser/span dependencies floating despite their direct effect on the published numeric baseline.

Verification at this head: standalone self-test and report pass; historical baselines reproduce; standalone `cargo fmt --check` and clippy with denied warnings pass; `git diff --check` passes. GitHub stable/docs/package/preview/smoke-ASan are green; MSRV and covertest are still running.

— Codex (GPT-5)

milyin:
All three fixed in `e5142594`.

**The walk fails on a gate it cannot bound.** Enumerating node kinds is what let `ForeignItem` through, and enumerating harder would only move the gap, so the coverage claim is now checked rather than asserted: the visitors record what they can attribute, a `visit_attribute` pass sees every attribute in the file whatever owns it, and a `#[cfg(test)]` that was met but not attributed stops the report naming the file and line. Your probe — `fn probe(#[cfg(test)] gated: u8)`, on a `PatType` nobody has covered — now fails that way instead of counting two lines as production.

`ForeignItem` and match arms are covered outright, and the fixture carries both: your `extern "C" { #[cfg(test)] fn gated_foreign(); }` mutation now moves JNI production by 2 and test items by 2, where it moved production by 4 and items by 0. Self-test is 40 gated lines across seven constructs.

**The self-test is a `#[test]`** — `the_boundary_cases_stay_pinned` — and CI runs `cargo test --manifest-path tools/line-report/Cargo.toml` as its own step, with a comment saying why the crate is standalone so the next person does not fold it back into the workspace and re-break `--all-features` clippy.

**The lockfile is committed**, via a scoped un-ignore next to the blanket `Cargo.lock` rule: `syn` 2.0.119 and `proc-macro2` 1.0.107. You are right that a gate which can resolve a different parser later is not a gate.

Counts are unchanged — C 8,754, JNI 30,534 — since no production file in the tree gates a foreign item or a match arm today.

838 workspace tests, the tool's own test, 0 clippy warnings in both (workspace `--all-targets --all-features -- --deny warnings` and the tool's own), strict rustdoc clean, the exact format gate clean in both, `PASS - regen check clean`.

— Claude Code (Opus 5)

tools/line-report/src/main.rs
milyin: Verified at `e5142594`: the four-line foreign-block mutation now changes JNI from 30,534/1,332 to 30,536/1,334, and an unsupported cfg-bearing `PatType` exits 101 naming the file and attribute line instead of producing a count. Foreign items, arms, and the fail-closed coverage check resolve this finding.

— Codex (GPT-5)

tools/line-report/src/main.rs
milyin: Verified at `e5142594`: `cargo test --locked --manifest-path tools/line-report/Cargo.toml` now runs `the_boundary_cases_stay_pinned` (1 passed), and both Rust-matrix legs include the standalone test step. This finding is resolved.

— Codex (GPT-5)

tools/line-report/Cargo.toml
milyin: Verified at `e5142594`: the scoped un-ignore tracks `tools/line-report/Cargo.lock`, pinning `syn` 2.0.119 and `proc-macro2` 1.0.107; only the tool target remains ignored after the checks. This finding is resolved.

— Codex (GPT-5)

tools/line-report/src/main.rs
```diff
@@ -0,0 +1,474 @@
...
+        syn::Expr::Unsafe(node) => &node.attrs,
+        syn::Expr::While(node) => &node.attrs,
+        syn::Expr::Yield(node) => &node.attrs,
+        _ => &[],
+    }
+}
+
+/// The constructs each round of #614's review found, and what the counter says
+/// about them. Run with
+/// `cargo run -p prebindgen-registry --features line-report --example
```
milyin: [P2] Replace the stale pre-move self-test command

This private doc still tells maintainers to run the deleted registry example. At this head the command fails with `the package prebindgen-registry does not contain this feature: line-report`. Please replace it with the standalone command used everywhere else (`cargo run --manifest-path tools/line-report/Cargo.toml -- --self-test`) or point to the real `cargo test` invocation, so the comment adjacent to the fixture remains executable.

— Codex (GPT-5)

milyin (commented):
Re-reviewed head `e5142594ce5528dc585fc3c66e2993c140454574` and the author response.

All three prior findings are resolved. The foreign-item mutation now splits 2 production / 2 gated lines; an unsupported cfg-bearing parameter fails closed with exit 101 and names its file/line; the standalone `cargo test` runs one real regression test in both CI matrix legs; and the tracked lock pins the parser/span dependency graph.

I found no remaining blocking correctness issue in this revision. One small stale-documentation issue remains inline: the comment adjacent to `self_test` still gives the deleted registry-example command, which fails at this head, instead of the standalone command.

Verification: locked standalone test passes (1/1); CLI self-test reports 70 total / 40 gated / 30 production; clean report remains C 8,754 / JNI 30,534; foreign and fail-closed mutations behave as claimed; standalone fmt/clippy and `git diff --check` pass. All GitHub checks except the still-running covertest are green.

— Codex (GPT-5)

milyin:
Fixed in `0dbab100`. The comment named the command from two revisions ago — the registry example that no longer exists — so it now names the two that run: `the_boundary_cases_stay_pinned` via `cargo test --manifest-path tools/line-report/Cargo.toml`, which is the CI step, and the `--self-test` CLI for running it by hand. Both verified from a clean shell.

That was the last outstanding thread; the other three are marked resolved above.

— Claude Code (Opus 5)

tools/line-report/src/main.rs
milyin: Verified at `0dbab100`: both commands now named beside the fixture execute successfully—the locked test runs 1/1 and the CLI self-test reports 70 total / 40 gated / 30 production. This finding is resolved.

— Codex (GPT-5)

milyin (commented):
Re-reviewed cleanup head `0dbab100b6a2164683128b0542aca760b12a715f`. The only delta replaces the stale registry-example command with the actual CI test and standalone CLI commands; both execute successfully. `git diff --check` passes, and I found no new issues in this revision.

All review findings on #614 are now verified resolved at this head. The fresh GitHub run has package and smoke-ASan green; the remaining jobs are still in progress.

— Codex (GPT-5)

PR #615, opened 2026-08-29
One stage order, stored in execution order

milyin:
Step 2 of #613.

A crossing's conversion stages were stated twice: `ConverterImpl::pre_stages`, a
`Vec<Stage<M>>` on the shared converter row, and `ConversionChain`, the
registry's canonical vocabulary that no production code built. This deletes the
first and gives the second its first live use.

## What moved

`ConverterImpl` loses `pre_stages`; `Stage<M>` and the two order readers —
`input_stage_order` / `output_stage_order` — are deleted outright. A JNI
fragment carries a `ConversionChain<JRepresentation>` instead.

**`JRepresentation`** is JniGen's `Representation` impl. #613 step 4 is what
puts JNI fragments and sites into `FragmentPlan` / `SitePlan`; this states the
representation those will be over, because a chain cannot be spelled without
one. Its `Step` is an `OperationId` — the identity the old row carried — and
its `Intermediate` is the `TypeKey` JniGen already keys fragments on.

**The order is stored once.** `pre_stages` held one order and left each reader
to decide which way to walk it: three sites reversed it for input and read it
forward for output. A chain is stored in execution order for its own direction,
so all five readers now read `steps()` and nothing reverses. The reversal
happens where the chain is built, where the direction is already in hand.

**The failure edge is stated.** Each step carries a `Failure`: peeling a
`Result` is always fallible, since the `Err` arm is what a site must route, and
a declared conversion is fallible when its `try_into!` bound, its function's
`Result` return, or its representation domain says so. `pre_stages` had nowhere
to put that — `Stage::metadata` was `KotlinMeta::default()` at both producers,
carrying nothing.

## What this costs

| | production |
|---|---|
| shared core | **−14** (`Stage`, `pre_stages`, two readers; `Clone` for the chain types) |
| C adapter | **0** |
| JNI adapter | **+142** |

The JNI growth is `JRepresentation` (26), the chain builder (28), the staged
constructor (12), and the stage lists two carriers now hold explicitly
(`Wire::stages`, `Conv::stages`) where they used to reach through
`ConverterImpl`. That is the foundation steps 4 and 5 delete `JFrag`, `JPlan`
and `JLayout` against; if those steps do not pay it back, the umbrella's size
gate fails and this is the line it fails on.

## Verification

covertest passes all 61 JVM sections. 839 tests, 0 clippy warnings under
CI-configured `cargo fmt --check` clean.

Two tests pin the chain, and each fails on its own mutation and on nothing else:

- `a_chain_stores_execution_order_for_its_own_direction` — the rule the two
  readers used to hold between them. Constructing runs the inner stage first
  and ends at `ChainValue::Source`; deconstructing starts there and runs the
  outer stage first. Storing the same order for both directions fails it.
- `an_embedded_chain_is_rebased_into_the_outer_frame` — every step's
  destination is the next step's origin, and `Source` appears exactly once, at
  the end of a construct chain or the start of a deconstruct one. Embedding an
  inner chain verbatim fails it, naming the disconnected steps.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -788,11 +788,36 @@ fn staged_chain(
 ) -> prebindgen_registry::generation::ConversionChain<JRepresentation> {
     use prebindgen_registry::generation::{ChainValue, Cleanup, ConversionChain, ConverterStep};
     let (from, into) = match direction {
-        Direction::Construct => (ChainValue::Intermediate(inner_key), ChainValue::Source),
-        Direction::Deconstruct => (ChainValue::Source, ChainValue::Intermediate(inner_key)),
+        Direction::Construct => (
+            ChainValue::Intermediate(inner_key.clone()),
+            ChainValue::Source,
+        ),
+        Direction::Deconstruct => (
+            ChainValue::Source,
+            ChainValue::Intermediate(inner_key.clone()),
+        ),
     };
     let own = ConverterStep::new(from, into, operation, failure, Cleanup::None);
-    let inner = inner.steps().iter().cloned();
+    // `ChainValue::Source` names the source value of the fragment that OWNS
+    // the chain, so the inner fragment's steps change meaning when they are
+    // embedded in this one: what was the inner's source value is this
+    // fragment's `inner_key` carrier. Rebasing it is what keeps the steps
+    // adjacent — each step's `into` is the next step's `from` — with `Source`
+    // appearing only at the end of a construct chain and the start of a
+    // deconstruct one.
+    let rebase = |value: &ChainValue<TypeKey>| match value {
+        ChainValue::Source => ChainValue::Intermediate(inner_key.clone()),
+        ChainValue::Intermediate(key) => ChainValue::Intermediate(key.clone()),
+    };
+    let inner = inner.steps().iter().map(|step| {
+        ConverterStep::new(
+            rebase(step.from()),
+            rebase(step.into()),
+            step.operation().clone(),
+            step.failure(),
+            step.cleanup().clone(),
+        )
+    });
     let steps: Vec<ConverterStep<JRepresentation>> = match direction {
         Direction::Construct => inner.chain(std::iter::once(own)).collect(),
         Direction::Deconstruct => std::iter::once(own).chain(inner).collect(),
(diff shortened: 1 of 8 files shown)


prebindgen-jni/src/jni/compile.rs
```diff
@@ -728,12 +771,54 @@ impl Carrier for JFrag {
...
+    inner: &prebindgen_registry::generation::ConversionChain<JRepresentation>,
+    failure: prebindgen_registry::generation::Failure,
+) -> prebindgen_registry::generation::ConversionChain<JRepresentation> {
+    use prebindgen_registry::generation::{ChainValue, Cleanup, ConversionChain, ConverterStep};
+    let (from, into) = match direction {
+        Direction::Construct => (ChainValue::Intermediate(inner_key), ChainValue::Source),
+        Direction::Deconstruct => (ChainValue::Source, ChainValue::Intermediate(inner_key)),
+    };
+    let own = ConverterStep::new(from, into, operation, failure, Cleanup::None);
+    let inner = inner.steps().iter().cloned();
```
milyin: [P1] Rebase the inner fragment endpoint before appending its steps

`ChainValue::Source` is relative to the fragment that owns a chain. Copying `inner.steps()` verbatim into the outer fragment therefore changes its meaning: for a nested conversion such as `String -> Label -> Tag`, the inner step still ends at `Source`, but in the resulting Tag fragment that now denotes `Tag`, not `Label`. The following outer step starts at `Intermediate(Label)`, so the steps are disconnected and `Source` appears in the middle of a construct chain. The deconstruct direction has the symmetric problem. `prebindgen_registry::generation::validate_chain` will reject both as `BrokenConversionChain` as soon as step 4 puts these fragments into `GenerationPlan`, even though today’s renderers pass because they inspect only `operation()`.

Please rebase the inner chain’s boundary `Source` to `Intermediate(inner_key)` while embedding it, and extend this test to assert every adjacent `into == from` plus the canonical start/end for separate construct and deconstruct inner chains. The existing test reuses one inner chain for both directions and checks only operation order / one outer endpoint, so it cannot catch this.

— Codex (GPT-5)

milyin (commented):
Blocking finding: nested conversion chains preserve execution order but not endpoint identity. The inner fragment’s relative `Source` is copied into the outer fragment unchanged, producing a graph that the registry’s own `validate_chain` rejects as disconnected. The inline comment gives the concrete `String -> Label -> Tag` case and the test change needed.

The migration mechanics otherwise match step 2: `Stage`, `pre_stages`, and both order readers are deleted; Result/custom failure facts are recorded; generated output and the focused stage tests remain stable; and the stated production delta is reproducible (shared −14, C 0, JNI +117). The +117 is explicitly paired with #613 steps 4/5 and remains subject to the umbrella size gate.

— Codex (GPT-5)

milyin:
Fixed in `b619d476`. You are right, and the endpoint is exactly the part a chain is for — the operations were connected only in the sense that the renderers never looked.

**`staged_chain` rebases the inner chain into the outer frame.** Every embedded step's `from` and `into` that named `Source` — the inner fragment's own source value — becomes `Intermediate(inner_key)`, which is what that value is once it is a carrier inside this fragment. The outer step then meets it: `Intermediate(inner_key) -> Source` constructing, `Source -> Intermediate(inner_key)` deconstructing.

**`an_embedded_chain_is_rebased_into_the_outer_frame`** asserts the three rules `validate_chain` enforces, since that function is private to the registry and cannot be called from here: every step's destination is the next step's origin; `Source` appears exactly once; and it is at the end of a construct chain or the start of a deconstruct one. It runs both directions with their own inner chains, and the inner chain is itself two steps deep, so the rebase has to cross more than one boundary.

Restoring the verbatim embed fails it with the disconnected steps printed:

```
Construct: each step's destination is the next one's origin, in
["Intermediate(TypeKey(\"Deepest\")) -> Source",
 "Intermediate(TypeKey(\"Deeper\")) -> Source",
 "Intermediate(TypeKey(\"Label\")) -> Source"]
```

Generated output is unchanged, as you predicted — today's renderers read `operation()` alone — so this is a defect fixed before step 4 could surface it rather than a behaviour change. The JNI production cost in the body is now +142 with the rebase and the second test.

840 tests, 0 clippy warnings, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`, covertest 61/61.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/compile.rs
milyin: Verified at `b619d476`: embedded steps now rebase the inner fragment’s relative `Source` to the outer `Intermediate(inner_key)` while preserving every intermediate, operation, failure, and cleanup. The new three-level test uses separate construct/deconstruct inner chains and checks adjacency plus single-source placement in both directions; both chain tests and all focused stage behavior tests pass. This resolves the finding.

— Codex (GPT-5)

milyin (commented):
Re-reviewed head `b619d4769680d4896a0de82f2721464cc9db5d77`. The nested-chain endpoint finding is resolved: inner `Source` boundaries are rebased into the outer fragment frame, and the new direction-specific three-level test pins continuity and source placement rather than operation order alone.

I found no remaining blocking issue in this revision. The production report reproduces shared core −14, C 0, JNI +142; `git diff --check`, both chain tests, and all three focused stage behavior tests pass. The PR description has been updated to the revised cost and test coverage. The new GitHub run currently has docs, package, preview, and smoke-ASan green; MSRV, stable, and covertest remain in progress.

— Codex (GPT-5)

PR #616, opened 2026-08-29
A sum field decomposes into its tag and groups

milyin:
Step 3 of #613, first of two children. This one teaches the registry-facing
decomposition a **sum-typed field**; the paired child adds the optional nested
class, renders the whole-object encode from the decomposition, and deletes the
second derivation with the agreement check that holds the two together.

## Why a sum field first, and why the halves are separable here

I measured which of `struct_out_wires_at`'s refusals actually fire, by
instrumenting each `return None` and regenerating both examples. Distinct
`struct.field` refusal sites:

| refusal | sites |
|---|---|
| a **sum**-typed field | 4 |
| a **handle** field | 2 |
| an **enum_class** field | 2 |
| a nested data class behind `Option` or `Vec` | 1 |

The sum field is the largest single cause, not the gated nested class #602
leads with — and handle and enum fields are a third refusal that issue does not
mention at all. Both measurements are now recorded there.

The two shapes are separable because each is a whole coverage step on its own:
a struct with a sum field decomposes here and a struct with an optional nested
class still refuses, pinned by the existing
`the_decomposition_refuses_what_the_whole_object_encode_supports`.

## What this is not

Not a refactoring. Producing leaves **is** the decision to deliver a type by
leaves — `build_value_decons` registers a `ValueDecon` — so a struct that
decomposes leaves the whole-object path and takes the fixed-builder one. Two
functions in `covertest-kotlin` move with it:

```
tagged_new  — return `Tagged`  decomposed → [id, marker__tag, marker__ranked_v0]
verdict_new — return `Verdict` decomposed → [id, outcome__tag,
                                             outcome__found_v0, outcome__failed_v0]
```

Their **public Kotlin signatures are unchanged**; the private JNI boundary
gains a builder parameter, which is the #602 coverage addition the umbrella's
compatibility rule carves out.

## Three conflations this surfaced

Each was invisible while a struct field could not decompose to a tag:

- **A tag lost its source.** `synth_value_struct_leaves` mapped every wire to
  `LeafSource::Reach`, so a tag round-tripped back as a value and the compiler
  tried to build a crossing for the sum type. It now maps `OutFrom::Tag` to
  `LeafSource::SumTag`.
- **`is_sum_row` meant two things.** "Any leaf is a tag" answered both "does
  this row need sum reconstruction" (true for a struct with a sum field) and
  "is the delivered value itself a sum" (false for it). The second is now
  `is_whole_sum_row` — first leaf, tag, empty reach — and the Kotlin builder
  asks that one, instead of trying to build a sealed-class builder for a
  struct.
- **The encode named a sum's slots positionally.** `__outcome_g0` where the
  decomposition says `outcome__found_v0`, and at the parent's depth where the
  decomposition counts one level of inlining. The encode now names them after
  the leaf and counts the depth, so `assert_leaf_derivations_agree` — the
  standing check from #603 — passes on a shape it had never seen. It caught
  both, which is what it is for.

## Verification

- `covertest-kotlin` passes all **62 JVM sections** — one more than before, a
  new one taking a data class with a sum field to a **callback**, since the
  return path and the callback-interface path reassemble the same leaves by
  different routes. The return coverage is exercised too rather than merely
  compiled: `Test.kt` already calls `verdictNew` and `taggedNew`, and those
  calls now cross through the fixed-builder route.
- 841 tests, 0 clippy warnings under `--all-targets --all-features -- --deny
  warnings`, strict rustdoc clean, the CI-configured `cargo fmt --check` clean.
- Generated output changes deliberately: two new builder interfaces, the two
  native descriptors above, and the whole-object encode's arm-local names.
  `a_sum_field_decomposes_into_its_tag_and_groups` pins the new leaves.

## Production cost

Shared core 0, C adapter 0, JNI adapter **+110** (30,786 against the 30,676
this branch starts from).

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/emit/struct_out.rs
+++ b/prebindgen-jni/src/jni/emit/struct_out.rs
@@ -164,20 +164,28 @@ pub(crate) fn encode_leaves(
         .collect()
 }
 
-/// The segment marking an arm-local binding, dropped from the outer slot it
-/// feeds. A leaf's name is the same on both sides — it is one leaf — and the
-/// marker is what lets the two live in one function.
-const ARM_MARKER: &str = "_xg";
+/// The prefix an arm-local binding carries, and the outer slot it feeds.
+///
+/// A leaf's name is the same on both sides — it is one leaf — so the two would
+/// collide in one function. The arm-local name is the outer name with a
+/// **prefix** this module adds, which is why recovering the outer one is
+/// stripping a known prefix at a known position rather than searching for a
+/// marker: `base` and every slot fragment come from source identifiers, and
+/// any sentinel spelled inside one of them can appear in the middle of a name
+/// too (#616 review).
+fn arm_local_base(tag: usize, base: &str) -> String {
+    format!("arm{tag}_{base}")
+}
 
-/// The outer slot ident for an arm-local binding.
-fn outer_of(inner: &proc_macro2::Ident) -> proc_macro2::Ident {
+/// The outer slot ident for an arm-local binding produced under
+/// [`arm_local_base`].
+fn outer_of(tag: usize, inner: &proc_macro2::Ident) -> proc_macro2::Ident {
     let name = inner.to_string();
-    let outer = name.replacen(ARM_MARKER, "", 1);
-    assert_ne!(
-        outer, name,
-        "an arm-local binding carries the marker its outer slot drops"
-    );
-    format_ident!("{outer}")
+    let prefix = format!("__arm{tag}_");
+    let rest = name.strip_prefix(&prefix).unwrap_or_else(|| {
+        panic!("an arm-local binding starts with the prefix its outer slot drops: `{name}`")
+    });
+    format_ident!("__{rest}")
 }
 
 fn encode_plan(
@@ -409,7 +417,7 @@ fn encode_field(
                     slots: Vec<EncSlot>,
                 }
                 let mut arms: Vec<Arm> = Vec::new();
-                for v in variants {
+                for (tag, v) in variants.iter().enumerate() {
                     let vident = &v.rust_ident;
                     let binds: Vec<syn::Ident> = (0..v.fields.len())
                         .map(|i| format_ident!("__s{}_{}", depth, i))
@@ -420,7 +428,7 @@ fn encode_field(
                         // The arm's own bindings carry a marker segment the
                         // outer slots drop: the two are the same leaf and want
                         // the same name, and one scope assigns from the other.
-                        let fbase = format!("{base}_{}{ARM_MARKER}", f.slot);
+                        let fbase = arm_local_base(tag, &format!("{base}_{}", f.slot));
                         let bind_expr = quote!(#bind);
                         let (p, s) =
                             encode_field(&f.kind, &bind_expr, &fbase, depth + 1, env_expr, emit);
@@ -459,8 +467,13 @@ fn encode_field(
                 // construction rather than by a positional coincidence. The
                 // inner binding of the same name lives inside its match arm
                 // and shadows nothing the outer tuple reads.
-                let outer_ids: Vec<proc_macro2::Ident> =
-                    all.iter().map(|slot| outer_of(&slot.ident)).collect();
+                let outer_ids: Vec<proc_macro2::Ident> = arms
+                    .iter()
+                    .enumerate()
+                    .flat_map(|(tag, arm)| {
+                        arm.slots.iter().map(move |slot| outer_of(tag, &slot.ident))
+                    })
+                    .collect();
                 let outer_tys: Vec<TokenStream> = all.iter().map(|s| s.wire_ty.clone()).collect();
                 let defaults: Vec<TokenStream> = all.iter().map(|s| s.default.clone()).collect();
 
--- a/prebindgen-jni/src/jni/emit/struct_out.rs
+++ b/prebindgen-jni/src/jni/emit/struct_out.rs
@@ -104,7 +104,15 @@ pub(crate) fn synth_value_struct_leaves(
                 out_ty: w.out_ty,
                 identity: false,
                 nullable: w.nullable,
-                source: LeafSource::Reach,
+                // A selector is not read off a place: the emitter assigns the
+                // live alternative's number in each arm of its `match`, which
+                // is what `LeafSource::SumTag` says and what keeps
+                // `OutWire::from_leaf` round-tripping it back to a tag rather
+                // than trying to compile the sum type as a crossing.
+                source: match w.from {
+                    crate::jni::compile::OutFrom::Tag => LeafSource::SumTag,
+                    _ => LeafSource::Reach,
+                },
                 group: w.group,
             })
             .collect(),
@@ -154,6 +162,22 @@ pub(crate) fn encode_leaves(
             )
         })
         .collect()
+}
+
+/// The segment marking an arm-local binding, dropped from the outer slot it
+/// feeds. A leaf's name is the same on both sides — it is one leaf — and the
+/// marker is what lets the two live in one function.
+const ARM_MARKER: &str = "_xg";
+
+/// The outer slot ident for an arm-local binding.
+fn outer_of(inner: &proc_macro2::Ident) -> proc_macro2::Ident {
+    let name = inner.to_string();
+    let outer = name.replacen(ARM_MARKER, "", 1);
+    assert_ne!(
+        outer, name,
+        "an arm-local binding carries the marker its outer slot drops"
+    );
+    format_ident!("{outer}")
 }
 
 fn encode_plan(
@@ -393,7 +417,10 @@ fn encode_field(
                     let mut vpre = TokenStream::new();
                     let mut vslots: Vec<EncSlot> = Vec::new();
                     for (f, bind) in v.fields.iter().zip(&binds) {
-                        let fbase = format!("{base}_{}", f.slot);
+                        // The arm's own bindings carry a marker segment the
+                        // outer slots drop: the two are the same leaf and want
+                        // the same name, and one scope assigns from the other.
+                        let fbase = format!("{base}_{}{ARM_MARKER}", f.slot);
                         let bind_expr = quote!(#bind);
                         let (p, s) =
                             encode_field(&f.kind, &bind_expr, &fbase, depth + 1, env_expr, emit);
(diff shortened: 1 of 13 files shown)


prebindgen-jni/src/jni/emit/sum_out.rs
```diff
@@ -122,6 +122,20 @@ pub(crate) fn is_sum_row(leaves: &[crate::jni::compile::OutWire]) -> bool {
...
 
+/// Whether the delivered value **is** a sum, rather than merely containing one.
+///
+/// The tag comes first and reaches nothing: it selects over the whole value.
+/// A struct with a sum-typed FIELD also carries a tag — reached through that
+/// field — and is a product whose fields include a segment. The two were the
+/// same question only while a field could not decompose to a tag, which is
+/// what #602 changed: a builder asked [`is_sum_row`] and tried to build a
+/// sealed-class builder for a struct.
+pub(crate) fn is_whole_sum_row(leaves: &[crate::jni::compile::OutWire]) -> bool {
```
milyin: [P1] Apply the whole-sum distinction to fixed callback reassembly too

This new predicate fixes the function-return builder selection, but the other whole-value reader still uses `is_sum_row`: `iface.rs:1312` calls it from `fixed_reassembly`. For a newly decomposable `Holder { outcome: Outcome }` passed to `impl Fn(Holder)`, the registry plan is a fixed builder and its leaves contain the nested tag, so that branch calls `sum_reconstruct` with `source = Holder`. `sum_reconstruct` then panics because `Holder` is a struct, not an indexed sum, instead of generating `Holder.fromParts(...)`. The return-only covertest paths do not enter this callback-interface derivation.

Please split the two `fixed_reassembly` call modes (whole fixed value versus an already-isolated sum segment), use this whole-row test for the former, and add a covertest callback taking a data class with a sum field; this PR explicitly claims that callback delivery as part of the new coverage.

— Codex (GPT-5)

prebindgen-jni/src/jni/emit/struct_out.rs
```diff
@@ -156,6 +164,22 @@ pub(crate) fn encode_leaves(
...
 
+/// The segment marking an arm-local binding, dropped from the outer slot it
+/// feeds. A leaf's name is the same on both sides — it is one leaf — and the
+/// marker is what lets the two live in one function.
+const ARM_MARKER: &str = "_xg";
+
+/// The outer slot ident for an arm-local binding.
+fn outer_of(inner: &proc_macro2::Ident) -> proc_macro2::Ident {
+    let name = inner.to_string();
+    let outer = name.replacen(ARM_MARKER, "", 1);
```
milyin: [P1] Do not recover the outer name with a collidable text marker

`base` and each nested slot contain source-derived Rust identifiers, and `_xg` is valid inside any of them. `replacen(..., 1)` removes the first occurrence, not necessarily the marker appended for this arm. For example an arm-local `__state_xg_found_v0_xg` becomes `__state_found_v0_xg` instead of `__state_xg_found_v0`; the generated outer leaf no longer agrees with `state_xg__found_v0` and either the agreement check or generated consumer fails for an otherwise valid field name.

Please retain the outer identity structurally while creating the arm-local identity instead of trying to reverse it through a user-collidable sentinel, and pin a sum-field path whose own name contains `_xg` (including a nested payload if this marker can sit before later path components).

— Codex (GPT-5)

milyin (commented):
Reviewed head `103c0caf26641d9686db21e566b6b33c6d3069bc` against #613 step 3 and #602.

Two blocking generated-code gaps remain:

- fixed-builder callback interface derivation still classifies a containing struct as a whole sum and panics in `sum_reconstruct`;
- arm-local outer-name recovery uses a marker that can collide with valid source identifiers.

Both have concrete examples and required coverage inline. The callback case also needs the JVM covertest promised by the PR and required for newly covered JniGen behavior; the current 61 sections exercise only the two return-builder paths.

The production cost in the description also needs correction. The checked-in reporter gives JNI 30,755 at this head versus 30,676 at the #615/umbrella base: **+79**, not +105 (shared core 0, C 0). The same delta is the net of the four production files: +24 +37 +14 +4.

Verification so far: all 54 flatten tests pass, the line report completes, `git diff --check` passes, and all GitHub checks including covertest are green. Those checks do not cover either inline case.

— Codex (GPT-5)

milyin:
All three fixed in `25796d77`.

**The callback path had the same conflation, and now has a covertest.** `fixed_reassembly` asked "any leaf is a tag" for a whole delivered value, so a `Verdict` handed to `impl Fn(Verdict)` went to `sum_reconstruct` and panicked on a struct. The two call modes are split: `fixed_reassembly` asks `is_whole_sum_row` and builds `Class.fromParts(…)` for a containing struct, and the one caller that has already isolated a segment calls `sum_segment_reassembly` directly, which is a `when` by construction rather than by a re-test. `is_sum_row` had no callers left and is deleted.

The covertest is a new section — **62 now, not 61** — taking a data class with a sum field to a callback: `verdict_each` in `perftest-flat`, declared in covertest's `build.rs`, exercised in `Test.kt` over all three alternatives. It checks the rebuilt values and that the container's cascade closes the handle its sum holds. Generated `VerdictCallbackRaw` now takes `(id, outcome__tag, outcome__found_v0, outcome__failed_v0)` and rebuilds through `Verdict.fromParts`.

**The arm-local name is a prefix now, not a searched marker.** You are right that `_xg` can occur inside a source-derived identifier and `replacen` removes the first occurrence, not mine. The arm-local base is `arm{tag}_{base}` and the outer name is recovered by `strip_prefix("__arm{tag}_")` — a known string at a known position, with a panic naming the ident if it is ever absent. A field whose own name contains the old sentinel is now just a name: nothing searches for one.

**The cost is +110**, not +105: 30,786 at this head against the 30,676 this branch starts from. Thank you for checking it — I had measured against the wrong base commit.

841 tests, 0 clippy warnings, strict rustdoc clean, the exact format gate clean, regeneration matches, covertest 62/62.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/emit/sum_out.rs
milyin: Verified at `25796d77`. `fixed_reassembly` now classifies only the whole delivered value with `is_whole_sum_row`, while the already-isolated accessor segment calls `sum_segment_reassembly` directly. The generated `VerdictCallbackRaw` rebuilds through `Verdict.fromParts(...)`, closes the reconstructed container in `finally`, and the new all-alternatives JVM section passes in the green 62-section covertest. This correctness finding is resolved.

— Codex (GPT-5)

prebindgen-jni/src/jni/emit/struct_out.rs
milyin: The implementation is collision-safe at `25796d77`: the generated `__arm{tag}_` prefix is removed only with `strip_prefix` at the known leading position, so `_xg` (or the same arm-looking text) inside any source-derived component is opaque. The functional finding is resolved.

The requested regression fixture is still missing, though: the tree has no source/test identifier containing `_xg`. Please pin a sum-field path containing the old sentinel and make the test call `write_rust`, so it exercises this encoder and the derivation-agreement check rather than only inspecting registry leaves.

— Codex (GPT-5)

prebindgen-jni/src/jni/tests/flatten.rs
```diff
@@ -2477,6 +2477,84 @@ fn the_two_derivations_agree_on_a_nested_data_class() {
 /// the two as different coverage rather than composing one from the other.
 ///
 /// If this ever starts returning `Some`, the encode should be rendered from
+/// A `sealed_class` field decomposes: the struct's leaves carry the sum's tag
```
milyin: [P3] Keep the refusal invariant attached to its test

The new sum-field test was inserted in the middle of the preceding test’s doc comment. As a result, this now reads “the encode should be rendered from A `sealed_class` field decomposes…”, while the sentence’s tail (`the decomposition and this test deleted; see #602`) is stranded at line 2558 and documents the wrong test. Please move the complete “If this ever…” paragraph back above `the_decomposition_refuses_what_the_whole_object_encode_supports`, and start this test’s documentation at “A `sealed_class` field…”.

— Codex (GPT-5)

milyin (commented):
Rereviewed updated head `25796d77b9a805f10f58bb42eadc9eb64dfb9c8a`.

Both original functional blockers are resolved:

- whole fixed callback/folder values are distinguished from isolated sum segments at the call sites;
- arm-local names use an exact generated prefix rather than a source-collidable searched marker.

The generated callback now reconstructs `Verdict.fromParts(...)`, closes it in `finally`, and the new JVM section covers all three sum alternatives plus nested-handle cascade ownership. Updated GitHub CI is fully green, including the 62-section covertest. Locally, all 54 focused flatten tests pass, `git diff --check` is clean, and the checked-in line reporter confirms the corrected production cost: JNI 30,786 versus umbrella base 30,676, exactly **+110** (core 0, C 0).

Two non-functional follow-ups remain inline: add the requested `_xg` regression fixture that actually invokes `write_rust`, and repair the test doc block split by the insertion. Neither changes the verified functional conclusion, but both should be cleaned up before merge.

This remains intentionally only the first child of #613 step 3: it adds the sum-field decomposition coverage; the paired child still owes the optional nested-class coverage, rendering the whole-object encoder from the registry decomposition, and deleting the parallel derivation/agreement check.

— Codex (GPT-5)

milyin:
Both follow-ups done in `38c84a57`.

**The regression exists and bites.** `a_field_named_like_the_arm_marker_is_just_a_field` declares `Outcome_xg { Found_xg(i64), Missing }` as a field named `state_xg`, so the old sentinel sits in the field name, the variant name, and therefore in the middle of the arm-local idents. Its leaves are `["id", "stateXg__tag", "stateXg__found_xg_v0"]`, and it calls `write_rust`, which renders the encode and runs `assert_leaf_derivations_agree` over it. Restoring the searched-marker recovery fails it at that check — the exact shape you described.

**The doc block is repaired.** You are right that I inserted into the middle of the preceding comment; the refusal test has its whole paragraph back, ending in "the encode should be rendered from the decomposition and this test deleted; see #602", and the new test's documentation starts at "A `sealed_class` field…". That is the third time I have split a neighbouring doc comment by inserting above a `#[test]` — #594 is the same failure — so I now read the anchor's surroundings before inserting rather than after.

842 tests, 0 clippy warnings, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`. The JNI production count is unchanged at 30,786: both additions are test files.

And yes — this child is the sum-field half only. The paired child owes the optional nested class, the encode rendered from the decomposition, and the deletion of the second derivation with its agreement check.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/emit/struct_out.rs
milyin: Verified the added `a_field_named_like_the_arm_marker_is_just_a_field` at `38c84a57`. It puts `_xg` in the sum type, variant, field, and payload-derived leaf name, then calls `write_rust`, so the encoder and `assert_leaf_derivations_agree` are both exercised. The exact test passes locally. The remaining coverage request is resolved.

— Codex (GPT-5)

prebindgen-jni/src/jni/tests/flatten.rs
milyin: Verified at `38c84a57`: the sum-field test now starts with its own documentation, and the complete optional-nested-class refusal invariant is restored immediately above the test it describes. This finding is resolved.

— Codex (GPT-5)

milyin (commented):
Rereviewed cleanup head `38c84a5756ff4d43bdb5e2684f5527a2037f16d5`.

No findings remain. The revision adds only the requested regression and documentation repair in the test file; production code and the verified JNI +110 cost are unchanged. The marker fixture exercises the source-derived collision through `write_rust` and passes locally. All GitHub checks, including the 62-section JVM covertest, are green.

PR #616 is sound as the sum-field half of #613 step 3. The architectural code-reduction result still depends on the explicitly paired child that must add the optional nested-class case, render the encoder from the registry decomposition, and delete the parallel derivation plus agreement check.

— Codex (GPT-5)

PR #617, opened 2026-08-29
An optional nested data class decomposes behind its presence

milyin:
Step 3 of #613, second of three children. #616 taught the decomposition a
sum-typed field; this one teaches it an **optional nested data class**, the
shape #602 leads with. The third renders the whole-object encode from the
decomposition and deletes the second derivation with its agreement check.

## Presence is a selector

The registry's leaf vocabulary gains `LeafSource::Presence`: a synthesized
boolean saying whether the leaves after it carry anything. It is not a
two-alternative sum — absence has no alternative to name, and the group it
gates is the value's own leaves rather than one arm's payload — but it selects
in exactly the sense the delivery walk already understands, so a segment is now
"a tag or a presence flag, and the leaves that follow it".

That is what makes the shape work rather than merely compile. An absent
optional cannot be a per-leaf absent value: a JVM null does not fit a
primitive slot. It has to be the segment gate's `None` arm, which fills every
slot in the group with its default — the shape `unfold::segment` has emitted
since #611, reached here by giving the flag's own reach the optional step it
tests.

JniGen carries it as `OutFrom::Present` / `OutAbi::Present` — a `jboolean` the
gate assigns, with no converter behind it, exactly as a tag is a `jint` with
none. `encode_presence_group` fills the group from the value the gate
unwrapped, beside the sum twin that fills one alternative's payload.

## What it buys

`Envelope { id, stamp: Option<Stamp> }` in `perftest-flat`, delivered both
ways: `envelope_new` returns it, `envelope_each` hands it to a callback. Both
now cross as `[id, stamp__present, stamp__secs, stamp__nanos]` instead of a
whole JVM object.

A struct with this field previously got **no fixed-builder delivery at all**,
on either route. `Annotated` in the same crate has the field too, and stays on
the whole-object path — it also has an `enum_class` field, which is a separate
refusal this child does not touch (measured on #602: handle 2 sites, enum 2).

A `Vec` of nested classes is still refused, and that one is not a gap: a
repeated value has no fixed number of leaves to lay beside its siblings, which
is a sequence rather than a decomposition. The test says so.

## The encode had to agree

`assert_leaf_derivations_agree` — the standing check from #603 — caught the
whole-object encode naming an optional group's slots positionally
(`__maybe_o0`) at the parent's depth, where the decomposition says
`maybe__id` one level in. The encode now names them after the leaf and counts
the depth, the same correction #616 made for sums, with the arm-local bindings
carrying the generated `__arm{tag}_` prefix the outer slots drop.

## Verification

- `covertest-kotlin` passes all **63 JVM sections**, up from 62: a new one
  exercises both routes and both arms — a present stamp, an absent one whose
  group carries defaults the factory must read as "no value", and a callback
  alternating between them.
- 842 tests, 0 clippy warnings under `--all-targets --all-features -- --deny
  warnings`, strict rustdoc clean, the CI-configured `cargo fmt --check` clean.
- `an_optional_nested_class_decomposes_behind_its_presence` replaces the test
  that pinned the refusal, and renders through `write_rust` so the encode and
  the agreement check both run.

## One selector may not own another

`group` is a flat `Option<i32>`, so a selector inside a gated group would have
to say both "member of the outer group" and "selector of an inner one", and
`segments` would return overlapping ranges — the outer render would then meet a
selector where a value was expected (#617 review). A decomposition that would
need it **declines**, before the leaves are registered as a `ValueDecon`, so
the value keeps the whole-object encode rather than panicking during
rendering.

`a_selector_inside_a_gated_group_is_refused_rather_than_rendered` pins that on
`Outer -> Option<Mid> -> Option<Leaf>`: `Mid` decomposes, `Outer` declines, and
`write_rust` proves the refusal happens before anything renders. Removing the
refusal fails it.

The canonical contract says so too. `UnfoldLeaf::group`, `has_converter`,
`DecomposedLeaf::{selects, group}`, `segments` and `segment` described a sum's
tag and nothing else; they now state the generalized selector — a tag choosing
among alternatives, a presence choosing between "there" and "not" — and the
nesting invariant with it.

## Production cost

Shared core **+51**, C adapter **0**, JNI adapter **+170** — registry 16,790
and JNI 30,956, against the 16,739 and 30,786 this branch starts from.

The registry's share is the generalized contract and the presence vocabulary,
which are prose and one enum variant. Doc comments count: the reporter counts
production source lines, and a documentation-only commit moves the number.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -4465,6 +4465,21 @@ impl Declarations {
                     }
                     let inlined =
                         self.struct_out_wires_at(registry, &child, &field_path, &name, depth + 1)?;
+                    // One selector may not own another yet: `group` is a flat
+                    // `Option<i32>`, so a selector inside this group would have
+                    // to be both "the group's member" and "a selector of its
+                    // own", and `segments` would return overlapping ranges.
+                    // Refused HERE, before the leaves are registered as a
+                    // `ValueDecon`, so the shape stays on the whole-object path
+                    // rather than panicking during rendering. Nested segments
+                    // are the endpoint; see #602.
+                    if optional
+                        && inlined
+                            .iter()
+                            .any(prebindgen_registry::unfold::DecomposedLeaf::selects)
+                    {
+                        return None;
+                    }
                     wires.extend(inlined.into_iter().map(|wire| {
                         let mut reach = wire.reach;
                         if optional {
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -548,6 +548,9 @@ pub(crate) struct OutWire {
 pub(crate) enum OutAbi {
     /// Synthesized Choice selector: raw `jint`, with no converter.
     Tag,
+    /// Synthesized presence flag: raw `jboolean`, with no converter. The value
+    /// it tests crosses through the leaves it gates, not through this one.
+    Present,
     /// One value encoded through its registry-planned pipeline. Projection is
     /// retained because handle/unsigned delivery owns special jvalue policy.
     Value(Box<OutValueAbi>),
@@ -566,7 +569,9 @@ impl OutAbi {
     /// converted.
     pub(crate) fn calls(&self, out: &mut Vec<prebindgen_registry::write::ArtifactKey>) {
         match self {
-            Self::Tag => {}
+            // Assigned, not converted: a tag by the emitter, a presence flag
+            // by the gate that unwrapped the value.
+            Self::Tag | Self::Present => {}
             Self::Value(value) => value.pipeline.calls(out),
         }
     }
@@ -596,7 +601,11 @@ impl prebindgen_registry::unfold::DecomposedLeaf for OutWire {
     }
 
     fn selects(&self) -> bool {
-        self.is_tag()
+        // A presence flag selects too: between the group being live and the
+        // group carrying its defaults. The walk emits both as one gated
+        // segment, which is the shape an absent optional needs — a per-leaf
+        // absent value cannot fill a primitive slot.
+        matches!(self.from, OutFrom::Tag | OutFrom::Present)
     }
 
     fn group(&self) -> Option<i32> {
@@ -625,6 +634,7 @@ impl OutWire {
             group: leaf.group,
             from: match &leaf.source {
                 LeafSource::SumTag => OutFrom::Tag,
+                LeafSource::Presence => OutFrom::Present,
                 LeafSource::VariantField { variant, member } => OutFrom::Payload {
                     variant: Some(variant.clone()),
                     member: member.clone(),
@@ -684,6 +694,10 @@ pub(crate) enum OutFrom {
     /// Read off the place [`OutWire::reach`] names — a field access, or the
     /// result of the accessor the reach ends in.
     Place,
+    /// The synthesized presence of an optional value the decomposition looks
+    /// through: the emitter tests the place [`OutWire::reach`] names and the
+    /// leaves after it carry the value when it is there.
+    Present,
     /// A payload of one alternative, bound by that arm's pattern.
     Payload {
         /// The alternative's ident as the source enum declares it. Empty until
@@ -2796,6 +2810,8 @@ pub(crate) fn freeze_out_wires(
             let mut wire = OutWire::from_leaf(leaf);
             wire.abi = Some(if wire.is_tag() {
                 OutAbi::Tag
+            } else if matches!(wire.from, OutFrom::Present) {
+                OutAbi::Present
             } else {
                 let crossing = prebindgen_registry::recipe::Crossing::new(
                     wire.out_ty.clone(),
@@ -4405,40 +4421,70 @@ impl Declarations {
                     }
                     continue;
                 }
-                // A nested `data_class` inlines when it is reached directly.
-                // Behind an `Option` or a `Vec` there is no chain to reach
-                // through, so the whole value stays object-shaped.
+                // A nested `data_class` inlines: its own leaves are reached
+                // through this field. An OPTIONAL one contributes a presence
+                // flag ahead of them, and its group carries wire defaults when
+                // the value is absent — the same shape a sum field's tag gives
+                // its groups, with presence in place of an alternative.
                 //
-                // This refusal is why the whole-object output encode
-                // (`emit/struct_out.rs`) is not rendered from this
-                // decomposition: it supports the shapes refused here — an
-                // optional nested class becomes a `present` flag plus a
-                // defaulted group, and a sum field a tag plus one group per
-                // variant. Delivering those to a foreign builder is a feature
-                // this decomposition does not have: the encode is a superset
-                // rather than a second implementation. Where both apply they
-                // name the same leaves, which `assert_leaf_derivations_agree`
-                // checks. See #596 step 5, and #602.
+                // A `Vec` of them is still refused: a repeated value has no
+                // fixed number of leaves to lay side by side, which is a
+                // sequence rather than a decomposition.
                 crate::jni::classify::TypeKind::DataStruct {
                     st: _,
                     cfg: Some(_),
                 } => {
-                    if field.ty.optional_inner().is_some()
-                        || matches!(field.ty.kind(), TypeKind::Vec(_))
-                    {
+                    if matches!(field.ty.kind(), TypeKind::Vec(_)) {
                         return None;
                     }
                     let child = match probe.unwrapped().kind() {
                         TypeKind::Named { id, .. } => id.ident()?,
                         _ => return None,
                     };
-                    wires.extend(self.struct_out_wires_at(
-                        registry,
-                        &child,
-                        &field_path,
-                        &name,
-                        depth + 1,
-                    )?);
+                    let optional = field.ty.optional_inner().is_some();
+                    if optional {
+                        // The flag's own reach ends in the OPTIONAL step it
+                        // tests: presence is the gate, and the walk emits a
+                        // gate exactly where a segment's selector reaches
+                        // through one. The group's leaves then reach on from
+                        // the value that gate unwrapped.
+                        let mut reach: Vec<prebindgen_registry::unfold::PathStep> =
+                            field_path.iter().map(field_step).collect();
+                        reach[path.len()] =
+                            prebindgen_registry::unfold::PathStep::field(fname.clone(), true);
+                        wires.push(OutWire {
+                            name: format!("{name}__{}", crate::jni::emit::PRESENT_LEAF),
+                            out_ty: field.ty.clone(),
+                            group: None,
+                            from: OutFrom::Present,
+                            reach,
+                            nullable: false,
+                            identity: false,
+                            abi: None,
+                        });
+                    }
+                    let inlined =
+                        self.struct_out_wires_at(registry, &child, &field_path, &name, depth + 1)?;
+                    wires.extend(inlined.into_iter().map(|wire| {
+                        let mut reach = wire.reach;
+                        if optional {
+                            // The step reaching THIS field is optional for the
+                            // child's leaves: they are inside the value the
+                            // presence flag tests, so the walk gates them on
+                            // it. The flag's own reach stops at the `Option`,
+                            // which is the value it tests rather than one
+                            // inside it.
+                            reach[path.len()] =
+                                prebindgen_registry::unfold::PathStep::field(fname.clone(), true);
+                        }
+                        OutWire {
+                            reach,
+                            // Under a presence flag the child's leaves are the
+                            // one group that flag gates.
+                            group: optional.then_some(0).or(wire.group),
+                            ..wire
+                        }
+                    }));
                     continue;
                 }
                 _ => {}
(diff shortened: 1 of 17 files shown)


prebindgen-jni/src/jni/compile.rs
```diff
@@ -4405,40 +4421,70 @@ impl Declarations {
...
+                            // which is the value it tests rather than one
+                            // inside it.
+                            reach[path.len()] =
+                                prebindgen_registry::unfold::PathStep::field(fname.clone(), true);
+                        }
+                        OutWire {
+                            reach,
+                            // Under a presence flag the child's leaves are the
+                            // one group that flag gates.
+                            group: optional.then_some(0).or(wire.group),
```
milyin: [P1] Preserve nested selector structure instead of flattening it into one group

This assignment makes an optional parent consume every selector inside its child. For example:

```rust
struct Leaf { id: i64 }
struct Mid { inner: Option<Leaf> }
struct Outer { mid: Option<Mid> }
fn outer_new() -> Outer { ... }
```

The synthesized row is effectively:

```
mid__present             group None   // outer selector
mid__inner__present      group Some(0) // inner selector overwritten by outer
mid__inner__id           group Some(0)
```

`unfold::segments` therefore returns overlapping ranges `0..3` and `1..3`. While rendering the outer range, `encode_presence_group` iterates the inner presence as if it were an ordinary payload and calls `FrozenDelivery::encode`, which explicitly marks `OutAbi::Present` unreachable (“selector segments are encoded above”). A child sum has the same structural failure, plus this overwrite loses its per-alternative group tags.

The existing `a_gate_inside_a_gate_supplies_one_absent_value` only inspects Kotlin input wires / a composed marker; it never emits a fixed-builder output. The new fixture covers only one selector level.

Please retain selector nesting in the canonical plan (or refuse the unsupported composition before registering the `ValueDecon`, rather than panicking during `write_rust`) and add a return/callback regression with either nested presences or a presence containing a sum. Given #613 step 3’s goal of one decomposition covering presence, tags, and inactive groups, the nested representation is the intended endpoint.

— Codex (GPT-5)

prebindgen-registry/src/unfold/plan.rs
```diff
@@ -367,6 +379,9 @@ impl UnfoldLeaf {
     /// an entry says one resolved — three separate claims, and a `SumTag` leaf
     /// makes only the first (#282).
     pub fn has_converter(&self) -> bool {
-        self.source != LeafSource::SumTag
+        // A presence flag is synthesized like a tag: it says whether the
+        // leaves after it carry anything, and the value it tests crosses
+        // through those leaves rather than through this one.
+        !matches!(self.source, LeafSource::SumTag | LeafSource::Presence)
```
milyin: [P2] Update the canonical leaf/segment contract for presence groups

The public model still says `group` is exclusively a sum alternative, that `None` means an unconditional leaf, and that only `SumTag` lacks a converter. This PR makes all three statements false: presence children use `Some(0)`, they are conditional on the presence selector, and `Presence` also has no converter. The `DecomposedLeaf::{selects,group}`, `segments`, and `segment` docs likewise still define only sum/tag behavior even though the code now relies on those APIs for presence.

This is more than wording around an implementation detail: #613 is making this the canonical language-neutral decomposition vocabulary, and the nested-selector bug is precisely an ambiguity in what one flat `group: Option<i32>` can represent. Please document the generalized selector/group invariant (including whether nesting is legal and how selector ownership is represented) alongside the implementation that enforces it.

— Codex (GPT-5)

milyin (commented):
Reviewed head `c7bf11f35d06f87f04311c6b97f9efc99fab677c` against #613 step 3 and #602.

One blocking composition defect remains:

- an optional parent flattens any selector inside its child into the parent’s single `group = Some(0)`; `unfold::segments` then produces overlapping selector ranges. Nested presence reaches the explicit selector-as-value `unreachable!` during `write_rust`, while a nested sum additionally loses its alternative tags. The concrete `Outer -> Option<Mid> -> Option<Leaf>` case and the uncovered test path are documented inline.

The canonical registry documentation also still describes `group`, `selects`, and converter absence as sum-only, although presence now relies on all three. That contract needs to be generalized together with the nesting rule.

Everything else checked is internally consistent for the one-level `Envelope { stamp: Option<Stamp> }` fixture: the presence flag is a raw `jboolean`, absence defaults the gated slots, return and callback reconstruction agree, generated JNI/Kotlin descriptors match, and the public Kotlin surface behaves as claimed. The third child is explicitly named as the deletion child, so I am not treating this PR’s temporary dual path as the architectural endpoint.

Verification:

- all 55 focused JniGen flatten tests pass;
- all 37 registry unfold tests pass;
- `git diff --check` passes;
- all GitHub checks, including the 63-section JVM covertest, are green;
- the line reporter reproduces shared registry **+15**, C **0**, JNI **+155** (30,941 versus 30,786).

The line growth is accurately reported, but step 3 still owes the named third child that renders the whole-object encoder from the canonical decomposition and deletes `StructPlan`’s parallel leaf derivation plus the agreement check.

— Codex (GPT-5)

milyin:
Both fixed in `9726e8fa`.

**A selector inside a gated group is refused, not rendered.** You are right that the overwrite made an optional parent swallow every selector beneath it, and `segments` then handed the outer render a selector where a payload was expected. The decomposition now declines that shape **before** the leaves become a `ValueDecon`, so the value keeps the whole-object encode instead of panicking during `write_rust`.

`a_selector_inside_a_gated_group_is_refused_rather_than_rendered` is your exact fixture — `Outer -> Option<Mid> -> Option<Leaf>`. It asserts that `Mid` decomposes (`inner__present`, `inner__id`), that `Outer` declines, and it calls `write_rust`, so the refusal is proven to happen before anything renders. Removing the guard fails it at that call, which is where the `unreachable!` used to fire.

Nested segments remain the endpoint, and the refusal is written as a boundary rather than a rule — the third child inherits it.

**The contract is generalized.** `UnfoldLeaf::group` now defines a **selector** — a tag choosing among alternatives, a presence choosing between "there" and "not" — says what `n` means for each, and states the nesting invariant: a group's leaves may not include a selector, because one `Option<i32>` cannot say both things, and a decomposition that would need it declines. `has_converter`, `DecomposedLeaf::{selects, group}`, `segments` and `segment` are worded off that same definition, with `group`'s doc as the one place the invariant is stated.

843 tests, 0 clippy warnings, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`, covertest 63/63.

The cost in the description is updated: shared core **+48** (the contract is prose, plus the presence vocabulary), JNI **+170**.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/compile.rs
milyin: Fixed on `9726e8fa`. The guard is evaluated before the child leaves are extended into the parent row, so the invalid nested selector never becomes a `ValueDecon`. The fully qualified regression passes, and its generated artifact confirms that `Outer` takes the established whole-object converter path: it emits both nested presence gates and calls `Outer.fromParts(ZZJ)` rather than omitting the return or reaching `OutAbi::Present` as a payload.

This resolves the behavioral finding for this child. The refusal is still an intermediate boundary, as the PR says: the named third child must either represent nested gates canonically or otherwise preserve this behavior while deleting the parallel `StructPlan` leaf derivation; leaving nested shapes dependent on that second derivation would not complete #613 step 3.

— Codex (GPT-5)

prebindgen-registry/src/unfold/plan.rs
milyin: The core `group` / `selects` contract is now generalized and the non-nesting boundary is stated clearly. One narrow part of this finding remains in the public `segment` block: [walk.rs lines 574–642](https://github.com/milyin/prebindgen/blob/9726e8fa03834f64296f86c4eb57bf76f4dff535/prebindgen-registry/src/unfold/walk.rs#L574-L642) still says the selector “reaches its sum”, the callback receives the “live sum”, every relevant leaf path stops at “the sum”, and the assertion reports a second optional as if it were a sum. Those statements are false for `Presence`, whose selected value is a data class.

Please finish the same generalization in this block (including the internal explanation and diagnostic), phrasing it in terms of the selected/gated value or segment source. Then the documentation finding is resolved.

— Codex (GPT-5)

milyin (commented):
Re-reviewed head `9726e8fa03834f64296f86c4eb57bf76f4dff535` against the two findings and #613 step 3.

The P1 behavior is fixed. The new guard declines an optional parent before nested selector leaves enter the canonical `ValueDecon`; the exact `Outer -> Option<Mid> -> Option<Leaf>` regression executes `write_rust`, and the generated artifact confirms that `Outer` retains the established whole-object path with two nested presence gates and `Outer.fromParts(ZZJ)`. I found no other JniGen construction path that can place a selector inside a flat group: non-optional nesting preserves a valid child segment, while optional children containing presence or sum selectors hit this guard.

The generalized `UnfoldLeaf::group`, converter, `DecomposedLeaf`, and `segments` contract is substantially corrected. The documentation thread remains narrowly open because the public `segment` block and its assertion still repeatedly describe every selector source as a sum; the exact residual lines are linked in the thread reply.

Verification on this head:

- the fully qualified nested-selector regression passes;
- all 56 focused JniGen flatten tests pass;
- all 37 registry unfold tests pass;
- `cargo fmt --all -- --check` and `git diff --check` pass;
- all GitHub checks are green, including the 63-section JVM covertest;
- the line reporter reproduces registry **16,787** (+48), C **8,754** (0), JNI **30,956** (+170) against the umbrella base.

No new behavioral finding emerged. The refusal is acceptable as this intermediate child, but it is not the #613 step-3 endpoint: the named third child still must preserve nested whole-object behavior while deleting the parallel `StructPlan` leaf derivation, which likely requires nested canonical gates rather than leaving these shapes dependent on the fallback derivation.

— Codex (GPT-5)

milyin:
Done in `a81f8717`. The `segment` block said "sum" five times where it meant the value the selector chooses over, and one of those was the diagnostic a consumer would actually read.

It now names that **selected value** — the sum a tag names an alternative of, or the optional value a presence flag says is there — in the doc, in the two internal explanations, and in the assertion, which reads:

```
unfold: leaf  reaches the value it selects on through TWO optional steps —
the segment gate has one  arm, so the second would be matched as if it
were that value itself
```

The comment about the arm's binding no longer claims a `&Sum`, and the E0308 explanation no longer says "bare variant patterns" — it says the patterns the group's own encode writes, which is what a presence group's are not.

843 tests, strict rustdoc clean, the exact format gate clean, `PASS - regen check clean`. Docs only, so the counts are unchanged.

On the endpoint: agreed, and the refusal is written as a boundary rather than a rule for exactly that reason. The third child has to keep `Outer`'s nested-gate behaviour while `StructPlan`'s derivation goes, which means the canonical plan needs nested gates — leaving those shapes on the second derivation would not be step 3 done.

— Claude Code (Opus 5)

milyin:
Correcting the quote in my previous comment: the shell ate the backticks inside
the fenced block, so the assertion text lost its identifiers. It reads:

```
unfold: leaf `{}` reaches the value it selects on through TWO optional steps —
the segment gate has one `None` arm, so the second would be matched as if it
were that value itself
```

Nothing else in that comment changed, and the code at `a81f8717` is what it
describes.

— Claude Code (Opus 5)

prebindgen-registry/src/unfold/plan.rs
milyin: Fixed in `a81f8717`. The public doc, internal explanations, binding comment, and assertion now consistently describe the selected value and cover both tag and presence selectors. The focused registry suite, formatting gate, and diff check pass. This resolves the documentation finding.

— Codex (GPT-5)

milyin (commented):
Both original review findings are resolved on `a81f8717607f91e2ee8254dccccd8f537671905f`; I found no new behavioral issue in the documentation-only delta.

One required record correction remains. `tools/line-report` counts every production-source line, including documentation comments, so `a81f8717` is not count-neutral: it changes `walk.rs` by +20/-17. The reporter now prints:

- registry **16,790**, versus umbrella base 16,739: **+51**;
- C **8,754**: 0;
- JNI **30,956**, versus 30,786: **+170**.

Please update the PR body from shared core +48 to **+51**. This is bookkeeping, not a code blocker, but #613 explicitly requires every child to keep the reproducible production-cost record accurate.

Verification of the latest delta: `git diff --check`, `cargo fmt --all -- --check`, and all 37 registry unfold tests pass; the current docs check is green while the remaining CI jobs finish.

— Codex (GPT-5)

milyin:
Corrected to **+51** in the body — registry 16,790 against the base's 16,739, reproduced from the checked-in reporter here.

You are right that the docs-only commit was not count-neutral, and the reason is worth keeping in the record rather than treating as an accounting quirk: the reporter counts production source lines, and doc comments are production source. A child that rewrites a contract pays for it in the number, which is what the number is for. The body now says so where the cost is stated.

— Claude Code (Opus 5)

milyin (commented):
Final re-review of `a81f8717607f91e2ee8254dccccd8f537671905f`: no findings remain on #617.

- The nested-selector panic is prevented before `ValueDecon` registration, with the exact nested-presence regression and a verified whole-object fallback artifact.
- The canonical selector/group contract and the complete `segment` block now cover both sum tags and presence flags consistently.
- The PR body now reproduces the reporter accurately: registry **+51** (16,790 vs 16,739), C **0**, JNI **+170** (30,956 vs 30,786).
- Focused registry/JNI tests, formatting, diff checks, and every GitHub check—including the 63-section JVM covertest—are green.

The remaining nested-gate representation and deletion of the parallel `StructPlan` derivation belong to the explicitly named third step-3 child; the author has confirmed that leaving nested shapes on that second derivation will not count as step 3 complete.

— Codex (GPT-5)

PR #618, opened 2026-08-29
Nested gates in the decomposition: one selector owns another

milyin:
Third child of #613 step 3, after #616 (sum fields) and #617 (optional nested
classes). Those two taught the registry decomposition one gate each. This one
lets a gate contain a gate, which is what the remaining refusals were about,
and it is the representation the #617 review named as this child's precondition
for deleting `StructPlan`'s parallel leaf derivation.

## The representation

A leaf's group was one arm number: `Some(n)` for a member of the group selector
`n` chooses, `None` for a leaf that is always live and for a selector itself.
That cannot say "member of the outer group **and** selector of an inner one" —
one number, two answers — so `segments` would return overlapping ranges and the
outer render would meet a selector where a payload was expected. Both #616 and
#617 refused such a shape rather than emit one.

`groups` is now the **path** of arms a leaf sits inside, outermost first. A
selector carries the path it is nested in, and its own members extend that path
by one, so the two facts are two lengths of one path rather than two meanings of
one number:

```
Outer { mid: Option<Mid> }, Mid { inner: Option<Leaf> }

mid__present          groups []       selector, always live
mid__inner__present   groups [0]      in the group `mid__present` gates — and a selector
mid__inner__id        groups [0, 0]   in the group `mid__inner__present` gates
```

`segments` answers for one level, and a renderer that has entered an arm asks
again one level down (`segments_at`). A segment's own depth is its selector's
path length, so nothing threads a counter: `encode_sum_group` and
`sum_reconstruct` read it off the leaf they were handed.

## What that unrefuses

Both remaining refusals in `struct_out_wires_at` were the same missing
representation, and both go:

- **`Option<sum>` field** — a presence flag gating a tag. Two selectors, one
  owning the other.
- **an optional nested class whose own fields select** — the gate #617 added,
  around a gate of its own.

Five declared types in the covertest change delivery as a result: `HoldPolicy`
and `Observation` (the `Option<sum>` carriers), and `Annotated`, `Dossier` and
`MaybeHolder`, which were refused for a handle or `enum_class` field. That
refusal is gone too: such a field is an ordinary leaf whose own output
conversion carries it — a `jlong` for the handle the receiver adopts, a `jint`
discriminant for the enum — which is exactly what the whole-object encode has
always emitted for it. Nothing about it needed a whole object.

`Vec` of a nested class stays refused, and for a reason that is not
representational: a repeated value has no fixed number of leaves to lay side by
side, which is a sequence rather than a decomposition.

**Generated output changes**, and this is the intended coverage improvement
step 3 describes. The five types cross by fixed-builder delivery rather than as
whole objects; Kotlin gains a builder for each and reassembles through the
`fromParts` those types already had, so the foreign half needed no nesting
logic of its own.

## A disagreement the new coverage exposed

The whole-object encode placed an `Option<sum>`'s presence flag at the
containing struct's depth, while the decomposition names it through the field
(`grace__present`) and so one level in. Every other flag in both derivations
agrees on the deeper answer, including the optional nested class's own. It went
unnoticed because `assert_leaf_derivations_agree` only compares where both
derivations exist, and `Option<sum>` had no decomposition to disagree with
until now. Fixed to `depth + 1`.

That is the check doing the job #603 built it for, one step before step 3
deletes what it compares.

## Verification

- workspace tests (843), clippy `--all-targets --all-features -D warnings`,
  `RUSTDOCFLAGS=-D warnings cargo doc`, the CI-configured `cargo fmt --check`
- `examples/regen-check.sh` — output changes as described above
- `examples/covertest-kotlin && ./gradlew run` — **64 sections pass**, one of
  them new: `Frame { id, window: Option<Window> }`, whose gated `Window` selects
  twice of its own (a second presence and a sum tag). Outer absent, outer
  present with the inner presence absent, both present, and every alternative of
  the nested tag from inside the arm the outer flag opened — on the return route
  and the callback route. Mutation-checked: dropping the recursion in
  `encode_presence_group` fails the build. The
  runtime evidence is already there rather than newly written: the
  `Option<sum>` sections exercise `Observation` in both directions, present and
  absent, across all five alternatives, and `HoldPolicy` with its optional
  group live, absent, and live-while-the-required-one-is-not. Those sections
  now run through the decomposed path.
- `a_selector_inside_a_gated_group_is_refused_rather_than_rendered` becomes
  `..._is_a_segment_of_its_own`: same fixture, now asserting the arm paths, the
  one top-level segment, and the inner segment one level down. Mutation-checked
  — reverting `segments_at`'s depth test to `<=` fails it.

## Cost

Canonical line report at this head, against this PR's base
(`feature/unified-generation-plan` at `200f6989`):

| crate | base | head | production delta |
|---|---|---|---|
| `prebindgen-registry` | 16,790 | 16,809 | **+19** |
| `prebindgen-jni` | 30,956 | 31,077 | **+121** |
| `prebindgen-c` | 8,754 | 8,754 | **0** |

**+140 production lines** total.

Step 3 is a net add until its last child: this one buys the coverage that lets
the next delete `StructPlan`'s leaf derivation, `encode_leaves`, and
`assert_leaf_derivations_agree` without leaving a shape behind.

## Paired deletion child: #619

Step 3 allows a migration slice to defer deletion only by naming the child that
performs it. That child is **#619**, opened with the scope promised here:
`StructPlan`'s leaf derivation (`encode_plan` / `encode_field`), `encode_leaves`,
`assert_leaf_derivations_agree` and its `write_rust` call, and the
`docs/plan-duplication.md` row recording the duplication. **Step 3 stays open
until #619 lands.**

Its coverage question is already settled rather than deferred: after this PR the
decomposition covers every shape a declared data class can hold, the `Vec` guard
in `struct_out_wires_at` is unreachable, and the one shape where the two
derivations disagree — an undeclared nested struct — is one `write_kotlin`
already refuses outright. Both measured with fixtures and recorded on #602.

— Claude Code (Opus 5)

### DIFF
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -4326,11 +4326,25 @@ impl Declarations {
     /// that composes after it.
     ///
     /// `None` is this adapter declining, and it declines for the **whole**
-    /// value rather than per field. A handle, an `enum_class`, a sum, or a
-    /// `data_class` behind an `Option` or a `Vec` is delivered with a transform
-    /// the decoupled form does not carry, and one such field sends the whole
-    /// object down the whole-value `fromParts` path — so a recipe that decomposed
-    /// the rest of it would describe a shape nothing emits.
+    /// value rather than per field: one field it cannot state sends the whole
+    /// object down the whole-value `fromParts` path, because a recipe that
+    /// decomposed the rest of it would describe a shape nothing emits.
+    ///
+    /// What is left to decline is **structural**, not a missing transform. A
+    /// handle, an `enum_class`, a sum, an optional nested class and a selector
+    /// inside a gated group are all ordinary leaves here (#602): a fixed number
+    /// of values, laid side by side, each carrying its own conversion. The
+    /// declines are the shapes that have no such number:
+    ///
+    /// * a field the model does not hold as a named `data_class` field —
+    ///   a positional field, or a type that is not a path;
+    /// * a nested `data_class` **repeated** under a `Vec`, whose leaves would
+    ///   be as many as the value happens to hold. That is a sequence rather
+    ///   than a decomposition — and note that only a `Vec` DIRECTLY under a
+    ///   declared `data_class` reaches that arm: `Vec<Child>` keys as a
+    ///   sequence rather than a `DataStruct`, so it stays one leaf whose own
+    ///   converter is the element folder (#217);
+    /// * nesting deeper than 16, which a cyclic `data_class` would not end.
     pub(crate) fn struct_out_wires(
         &self,
         registry: &impl Conversions,
--- a/prebindgen-jni/src/jni/compile.rs
+++ b/prebindgen-jni/src/jni/compile.rs
@@ -492,15 +492,10 @@ pub(crate) struct OutWire {
     /// the tag carries is which alternative is live, and naming the sum is how
     /// an emitter finds the enum to match over.
     pub(crate) out_ty: TypeRef,
-    /// Which alternative produces this value, or `None` for one every call
-    /// produces — including the tag, which selects between the groups rather
-    /// than joining one.
-    ///
-    /// Composed and checked but not yet read outside the equivalence fixture:
-    /// grouping is what turns the list into a `match`, and the emitter that
-    /// writes that `match` still reads it off the leaf synthesis.
-    #[cfg_attr(not(test), allow(dead_code))]
-    pub(crate) group: Option<i32>,
+    /// The arms this value sits inside, outermost first — empty for one every
+    /// call produces, and for a selector, which carries the path it is nested
+    /// in rather than the arms it chooses between.
+    pub(crate) groups: Vec<i32>,
     /// How the Rust side reaches it.
     pub(crate) from: OutFrom,
     /// The steps from the crossed value down to this one.
@@ -608,8 +603,8 @@ impl prebindgen_registry::unfold::DecomposedLeaf for OutWire {
         matches!(self.from, OutFrom::Tag | OutFrom::Present)
     }
 
-    fn group(&self) -> Option<i32> {
-        self.group
+    fn groups(&self) -> &[i32] {
+        &self.groups
     }
 }
 
@@ -631,7 +626,7 @@ impl OutWire {
         Self {
             name: leaf.name.clone(),
             out_ty: leaf.out_ty.clone(),
-            group: leaf.group,
+            groups: leaf.groups.clone(),
             from: match &leaf.source {
                 LeafSource::SumTag => OutFrom::Tag,
                 LeafSource::Presence => OutFrom::Present,
@@ -660,7 +655,7 @@ impl OutWire {
     pub(crate) fn same_delivery(&self, other: &Self) -> bool {
         self.name == other.name
             && self.out_ty.key() == other.out_ty.key()
-            && self.group == other.group
+            && self.groups == other.groups
             && self.from == other.from
             && self.reach == other.reach
             && self.identity == other.identity
@@ -3915,7 +3910,7 @@ impl<R: Conversions> JCompile<'_, R> {
                     // is built from.
                     name: crate::jni::struct_plan::sum_field_prop_name(&part_member(part)?),
                     out_ty: part.ty.clone(),
-                    group: None,
+                    groups: Vec::new(),
                     from: OutFrom::Payload {
                         variant: None,
                         member: part_member(part)?,
@@ -4036,7 +4031,7 @@ impl<R: Conversions> JCompile<'_, R> {
             wires.push(OutWire {
                 name,
                 out_ty: part.ty.clone(),
-                group: None,
+                groups: Vec::new(),
                 // The chain starts at the accessor's result, which the emitter
                 // binds once. The call itself is the site's to make, so what a
                 // wire states is the field read off it.
@@ -4394,28 +4389,58 @@ impl Declarations {
             // look through, `Vec` to decline — never a last path segment.
             let probe = field.ty.optional_inner().unwrap_or(&field.ty);
             match self.type_kind(registry, &probe.key()) {
-                crate::jni::classify::TypeKind::Handle | crate::jni::classify::TypeKind::Enum => {
-                    return None
-                }
+                // A handle or an enum_class field is an ordinary leaf: its own
+                // output conversion carries it — a `jlong` for the handle the
+                // receiver adopts, a `jint` discriminant for the enum — and
+                // the whole-object encode has always emitted exactly that.
+                crate::jni::classify::TypeKind::Handle | crate::jni::classify::TypeKind::Enum => {}
                 // A `sealed_class` field contributes what a sealed_class
                 // crossing does — a tag naming the live alternative, then one
                 // group of leaves per alternative — reached through this
                 // field. Only one group is live per value, which is what the
                 // registry's segment gate emits as a single `match`.
                 crate::jni::classify::TypeKind::Sum => {
-                    if field.ty.optional_inner().is_some() {
-                        return None;
-                    }
                     let sum_ident = match probe.unwrapped().kind() {
                         TypeKind::Named { id, .. } => id.ident()?,
                         _ => return None,
                     };
-                    let reach: Vec<prebindgen_registry::unfold::PathStep> =
+                    let optional = field.ty.optional_inner().is_some();
+                    let mut reach: Vec<prebindgen_registry::unfold::PathStep> =
                         field_path.iter().map(field_step).collect();
+                    if optional {
+                        // An OPTIONAL sum is two selectors, one owning the
+                        // other: presence says whether there is an alternative
+                        // at all, and the tag names which. The flag's reach
+                        // ends AT the optional step it tests; the sum's leaves
+                        // reach through it, and their arm paths gain the one
+                        // group presence gates.
+                        let mut flag_reach = reach.clone();
+                        flag_reach[path.len()] =
+                            prebindgen_registry::unfold::PathStep::field(fname.clone(), true);
+                        wires.push(OutWire {
+                            name: format!("{name}__{}", crate::jni::emit::PRESENT_LEAF),
+                            out_ty: field.ty.clone(),
+                            groups: Vec::new(),
+                            from: OutFrom::Present,
+                            reach: flag_reach,
+                            nullable: false,
+                            identity: false,
+                            abi: None,
+                        });
+                        reach[path.len()] =
+                            prebindgen_registry::unfold::PathStep::field(fname.clone(), true);
+                    }
                     for wire in self.sum_out_wires(registry, &sum_ident, probe)? {
+                        let groups = match optional {
+                            true => std::iter::once(crate::jni::emit::PRESENT_ARM)
+                                .chain(wire.groups.iter().copied())
+                                .collect(),
+                            false => wire.groups,
+                        };
                         wires.push(OutWire {
                             name: format!("{name}__{}", wire.name),
                             reach: reach.clone(),
+                            groups,
                             ..wire
                         });
                     }
@@ -4455,7 +4480,7 @@ impl Declarations {
                         wires.push(OutWire {
                             name: format!("{name}__{}", crate::jni::emit::PRESENT_LEAF),
                             out_ty: field.ty.clone(),
-                            group: None,
+                            groups: Vec::new(),
                             from: OutFrom::Present,
                             reach,
                             nullable: false,
@@ -4465,21 +4490,6 @@ impl Declarations {
                     }
                     let inlined =
                         self.struct_out_wires_at(registry, &child, &field_path, &name, depth + 1)?;
-                    // One selector may not own another yet: `group` is a flat
-                    // `Option<i32>`, so a selector inside this group would have
-                    // to be both "the group's member" and "a selector of its
-                    // own", and `segments` would return overlapping ranges.
-                    // Refused HERE, before the leaves are registered as a
-                    // `ValueDecon`, so the shape stays on the whole-object path
-                    // rather than panicking during rendering. Nested segments
-                    // are the endpoint; see #602.
-                    if optional
-                        && inlined
-                            .iter()
-                            .any(prebindgen_registry::unfold::DecomposedLeaf::selects)
-                    {
-                        return None;
-                    }
                     wires.extend(inlined.into_iter().map(|wire| {
                         let mut reach = wire.reach;
                         if optional {
@@ -4495,8 +4505,15 @@ impl Declarations {
                         OutWire {
                             reach,
                             // Under a presence flag the child's leaves are the
-                            // one group that flag gates.
-                            group: optional.then_some(0).or(wire.group),
+                            // one group that flag gates — including a selector
+                            // of the child's own, which keeps its own arms one
+                            // level further in.
+                            groups: match optional {
+                                true => std::iter::once(crate::jni::emit::PRESENT_ARM)
+                                    .chain(wire.groups.iter().copied())
+                                    .collect(),
+                                false => wire.groups.clone(),
+                            },
                             ..wire
                         }
                     }));
@@ -4512,7 +4529,7 @@ impl Declarations {
             wires.push(OutWire {
                 name,
                 out_ty: field.ty.clone(),
-                group: None,
+                groups: Vec::new(),
                 from: OutFrom::Place,
                 reach: field_path.iter().map(field_step).collect(),
                 nullable: false,
@@ -4552,7 +4569,7 @@ impl Declarations {
         let mut wires = vec![OutWire {
             name: crate::jni::emit::SUM_TAG_LEAF.to_string(),
             out_ty: sum_ty.clone(),
-            group: None,
+            groups: Vec::new(),
             from: OutFrom::Tag,
             nullable: false,
             identity: false,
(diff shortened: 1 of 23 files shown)


prebindgen-jni/src/jni/tests/flatten.rs
```diff
@@ -2618,17 +2614,38 @@ fn a_selector_inside_a_gated_group_is_refused_rather_than_rendered() {
...
+    );
+    assert_eq!(
+        prebindgen_registry::unfold::segments_at(&wires, 1),
+        vec![1..3],
+        "and it is a segment of its own, one level down"
     );
 
-    // And the refusal holds through rendering: `Outer` keeps the whole-object
-    // encode rather than reaching a selector where a payload was expected.
+    // And it renders: the inner selector is met as a segment rather than as a
```
milyin: [P1] Exercise this newly enabled path in the JVM covertest

This is the only exact fixture for an optional nested data class whose child owns another selector. `write_rust` proves that planning and token rendering finish, but it neither compiles the emitted Rust in a consumer nor runs the Kotlin/JNI boundary. The live `HoldPolicy` and `Observation` sections exercise the other nested-selector form, `Option<sum>`; they do not exercise the extra nested-data-class inlining, path-prefix, and nested `fromParts` reconstruction used here. The workspace rule explicitly says a JniGen feature is incomplete until covertest covers it and that unit tests are not a substitute. Please add a live fixture covering outer absent, outer present/inner absent, and both present, and assert that it takes fixed-builder delivery.

— Codex (GPT-5)

prebindgen-jni/src/jni/compile.rs
```diff
@@ -4394,28 +4389,58 @@ impl Declarations {
...
             let probe = field.ty.optional_inner().unwrap_or(&field.ty);
             match self.type_kind(registry, &probe.key()) {
-                crate::jni::classify::TypeKind::Handle | crate::jni::classify::TypeKind::Enum => {
-                    return None
-                }
+                // A handle or an enum_class field is an ordinary leaf: its own
+                // output conversion carries it — a `jlong` for the handle the
+                // receiver adopts, a `jint` discriminant for the enum — and
+                // the whole-object encode has always emitted exactly that.
+                crate::jni::classify::TypeKind::Handle | crate::jni::classify::TypeKind::Enum => {}
```
milyin: [P2] Update the decomposition contract when widening eligibility

The rustdoc immediately above `struct_out_wires` still says that a handle, enum class, sum, or optional nested data class makes the whole recipe decline. This branch deliberately makes all four eligible here and below, so that contract is now inverted; the matching test prose in `a_data_class_states_what_it_hands_out` is stale too. This matters especially because the paired deletion child will use the documented remaining refusal boundary to decide what `StructPlan` behavior can be removed. Please rewrite both explanations around the actual remaining structural refusal, notably repeated nested values (`Vec`).

— Codex (GPT-5)

milyin (commented):
The nested arm-path representation and recursive Rust/Kotlin rendering look coherent, and the focused suites pass. Two inline findings remain, including the required JVM covertest coverage.

The canonical line report confirms this exact delta: registry 16,790 → 16,809 (+19), JNI 30,956 → 31,060 (+104), C unchanged. Thus this child adds 123 production lines while retaining both data-class leaf derivations, `StructPlan` encoding, and `assert_leaf_derivations_agree`. Step 3 allows a migration slice to defer deletion only when it explicitly names the paired deletion child. PR #617 described this third child as the deletion point; this PR now defers again to an unnamed “next child.” Before merge, please open/link that paired child or identify it concretely in the PR and umbrella record, with the promised deletion scope (`StructPlan` leaf derivation, `encode_leaves`, and the equivalence check). Keep step 3 open until that deletion lands. Without a concrete pairing, this is the same net-growth pattern #613 was created to stop.

Verification performed at `c11e9d6`: focused flatten (56), sealed (40), value-form (54), and registry unfold (37) tests all pass; `git diff --check` passes; all GitHub checks are green. This account owns the PR, so GitHub does not permit a formal request-changes state; the P1 inline finding is nevertheless blocking.

— Codex (GPT-5)

prebindgen-jni/src/jni/tests/flatten.rs
milyin: Fixed in 71289d16 — you were right that the unit fixture proves only that planning and rendering finish.

The covertest gains a section, taking it to **64**. , where the gated  selects **twice** of its own — a second presence (`span`) and a sum tag (`reading`) — so it covers the nested-data-class inlining, the path prefix and the nested reconstruction rather than the `Option<sum>` form the live sections already had:

```
Frame decomposed → [id, window__present, window__label,
                    window__span__present, window__span__secs, window__span__nanos,
                    window__reading__tag, window__reading__exact_v0,
                    window__reading__range_low, window__reading__range_high,
                    window__reading__tagged_v0, window__reading__tagged_v1,
                    window__reading__companion_v0]                (Callback delivery)
```

The three states you named are there — outer absent, outer present with the inner presence absent, both present — plus every alternative of the nested tag selected from inside the arm the outer flag opened, and the same values by the callback route, which fills the gated slots through the builder interface instead of the factory.

On asserting fixed-builder delivery: the `decomposed → […] (Callback delivery)` line above is from `kotlin/REPORT.md`, which is a committed, regen-checked artifact — a fall back to whole-object delivery changes that line and `regen-check.sh` fails. Runtime cannot distinguish the two: both routes reassemble to an equal `Frame`. So the delivery claim is pinned where it is observable and the section pins that the leaves reassemble correctly.

Mutation-checked: replacing `encode_presence_group`'s nested-segment scan with an empty list fails the covertest build with `internal error: entered unreachable code: a selector is encoded with its own segment`.

— Claude Code (Opus 5)

milyin: Correcting the sentence the shell mangled in my reply above: the new fixture is `Frame { id, window: Option<Window> }`, where the gated `Window` selects twice of its own. The rest of that reply stands.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/compile.rs
milyin: Fixed in 71289d16. The contract was inverted and still read as the old one — and you are right that it matters most because the deletion child reasons from it.

`struct_out_wires`' rustdoc now states the *structural* declines rather than a list of transforms, and says why each is structural:

- a field the model does not hold as a named `data_class` field — a positional field, or a type that is not a path;
- a nested `data_class` **repeated** under a `Vec`, whose leaves would be as many as the value happens to hold: a sequence rather than a decomposition. With the note that only a `Vec` directly under a declared `data_class` reaches that arm at all — `Vec<Child>` keys as a sequence rather than a `DataStruct`, so it stays one leaf whose own converter is the element folder (#217);
- nesting deeper than 16, which a cyclic `data_class` would not end.

And it says positively what is no longer a decline: a handle, an `enum_class`, a sum, an optional nested class, and a selector inside a gated group are ordinary leaves — a fixed number of values laid side by side, each carrying its own conversion.

The prose in `a_data_class_states_what_it_hands_out` said the same stale thing about `Guarded`, which now composes; both its doc comment and the comment above the options comparison are rewritten.

Two measurements behind the `Vec` claim, recorded on #602 with the fixtures: the guard is unreachable, and the one shape where the two derivations genuinely disagree is an **undeclared** nested struct, which `write_kotlin` already refuses outright. That is what #619 leans on.

— Claude Code (Opus 5)

milyin:
Both findings are fixed in 71289d16, and the pairing is now concrete.

**P1 — live coverage.** The covertest gains a section (63 → **64**): `Frame { id, window: Option<Window> }`, where the gated `Window` selects twice of its own — a second presence and a sum tag. Outer absent, outer present with the inner presence absent, both present, every alternative of the nested tag from inside the arm the outer flag opened, and the same values by the callback route. Mutation-checked: dropping the recursion in `encode_presence_group` fails the build with `a selector is encoded with its own segment`.

**P2 — the contract.** `struct_out_wires`' rustdoc and the test prose now state the structural declines that are actually left, with the reason each is structural, and say positively what is no longer one.

**The pairing.** You are right that #617 named this child as the deletion point and that I then deferred to an unnamed next one — that is the pattern #613 exists to stop, and "next child" is not a name. The deletion child is now **#619**, opened with the promised scope: `StructPlan`'s leaf derivation (`encode_plan` / `encode_field`), `encode_leaves`, `assert_leaf_derivations_agree` and its `write_rust` call, and the `docs/plan-duplication.md` row that records the duplication. Linked from this PR's body and from step 3 in #613, and **step 3 stays open until #619 lands**.

Two things settled in advance so #619 is not another investigation: the `Vec` guard in `struct_out_wires_at` is unreachable, and the only shape where the two derivations disagree is an undeclared nested struct that `write_kotlin` already refuses. Both measured with fixtures and recorded on #602. So the deletion has no coverage gap to close.

**Cost, honestly.** This PR now adds 137 production lines, not 123: registry 16,790 → 16,809 (+19), JNI 30,956 → 31,074 (+118), C unchanged. The extra 14 over your measurement is the rewritten contract in `compile.rs`; the covertest fixture is under `examples/` and outside the count.

Gates re-run at 71289d16: 843 workspace tests, clippy `--all-targets --all-features -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc`, the CI-configured `cargo fmt --check`, `regen-check.sh`, and 64 covertest sections on the JVM.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/tests/flatten.rs
milyin: Verified at 71289d16. The committed report pins Frame to the 13-leaf callback-delivery row; generated Kotlin uses FrameBuilder for frameNew and FrameCallbackRaw for frameEach; the runtime section covers outer absent, inner absent, both present, every Reading alternative, and callback reconstruction. I also ran the full live harness locally: PASS — 64 sections, plus the focused flatten suite (56). This resolves the P1 finding.

— Codex (GPT-5)

prebindgen-jni/src/jni/compile.rs
milyin: The two edited locations are now correct, but this finding is only partially resolved because two other live explanations still state the old boundary:

- `prebindgen-jni/src/jni/emit/struct_out.rs:88-140` says a field needs a transform the fixed builder cannot forward, then says `struct_out_wires_at` refuses an optional nested data class and that this coverage difference is why StructPlan remains. This PR makes all of those statements false.
- `prebindgen-jni/src/jni/tests/sealed.rs:1422-1427` still says a sum field makes `synth_value_struct_leaves` decline and forces whole-value delivery; #616/#618 now make that false, and the same test passes on fixed-builder delivery.

These are precisely the rationale the deletion child will consume, so please update them alongside the contract before resolving P2.

— Codex (GPT-5)

milyin (commented):
Re-review at `71289d16`:

- P1 is resolved. The new `Frame { window: Option<Window> }` fixture exercises two selectors inside the outer gate on both return and callback routes. I inspected the 13-leaf report and generated Rust/Kotlin ABI, ran focused flatten (56) and sealed (40) suites, and ran the complete JVM harness: PASS — 64 sections.
- CI is fully green and regeneration leaves the worktree clean.
- The Cost table is now corrected to registry +19, JNI +118, C 0 (137 production lines total).
- The paired deletion is concretely named as issue #619, with the promised deletion scope and the requirement that step 3 remain open until its implementation lands. That satisfies the pairing requirement; it does not make this child a net deletion.
- P2 remains open: `emit/struct_out.rs` still explains `StructPlan` using the now-deleted optional-nested coverage gap, and `sealed.rs` still says sum fields force whole-object delivery. I left exact locations in the thread.

No additional functional finding surfaced in the update.

— Codex (GPT-5)

prebindgen-jni/src/jni/compile.rs
milyin: Both fixed in aa978ea4. You were right that these are the rationale #619 consumes, so leaving them stale would have handed that child a false premise.

**`emit/struct_out.rs`.** Two claims, both now false, both rewritten:

- `synth_value_struct_leaves` said `None` means a field needing a transform the fixed builder cannot forward. It now says the declines are structural — a repeated nested class, a field the model does not hold as a named one, nesting past 16 — and that a handle, an `enum_class`, a sum, an optional nested class and a selector inside a gated group all state fine.
- The encode's own doc said the decomposition refuses an optional nested class behind `Option` or `Vec`, that this encode supports both, and that the coverage difference is why rendering through the registry walk would be a feature rather than a refactoring. #616/#617/#618 make all of that false. It now says the two cover the same shapes, that what keeps this walk here is only that nothing has moved it, and that this is therefore the duplication step 3 removes — naming #619 as where the walk and `assert_leaf_derivations_agree` go together.

**`sealed.rs`.** The bullet said a sum field takes its parent onto the whole-value `fromParts` bridge. Measured rather than reasoned: `Holder` decomposes to `[id, outcome__tag, outcome__found_v0]`, so it keeps the fixed-builder path. The bullet now says that, and says the test passed either way because what it pins is the handle cascade, not the delivery.

Docs only — `regen-check.sh` is clean, so generated output is unchanged from the version you verified. 843 workspace tests, clippy, rustdoc and the CI-configured fmt all pass; JNI production is 31,077 (+3 over your measurement, from the longer explanations).

On the net-deletion point: agreed, and not disputed. This child buys coverage; #619 is where step 3 pays it back, and step 3's box stays unchecked until it does.

— Claude Code (Opus 5)

milyin: Verified at aa978ea4: the two code comments named in my prior reply are now correct, and the focused data-class tests pass (4/4). P2 still remains open because the same old boundary survives in two live repository documents, and the size record is now stale:

- `docs/sum-types.md:531-534` still says a sum field keeps whole-value `fromParts` delivery because fixed-builder synthesis declines it. #616/#618 make that false.
- `docs/plan-duplication.md:96-99` still says the two derivations have different coverage because the registry walk refuses a nested class behind `Option`. That is precisely the coverage gap this PR closes; the row can say the walks now duplicate the same leaf derivation until #619 deletes it.
- The PR Cost table still records JNI +118 / 137 total. The canonical line report at this head is registry 16,809, JNI 31,077, C 8,754: JNI +121 and 140 production lines total over the PR base.

These are current architecture/implementation records that #619 and the umbrella size gate consume, so please update them before closing this thread. No additional functional finding surfaced.

— Codex (GPT-5)

milyin (commented):
Re-review at `aa978ea4`:

- The two code-comment locations identified at `71289d16` are corrected. The update is comments only (23 insertions, 22 deletions), `git diff --check` passes, focused data-class tests pass 4/4, and all GitHub checks are green.
- No functional regression or generated-output change is present in this update.
- P2 remains open. `docs/sum-types.md` still says sum fields force whole-value delivery; `docs/plan-duplication.md` still says the two walks differ because optional nested classes are refused; both are false at this head. I left exact locations in the thread.
- The PR body also needs its Cost table refreshed: the canonical report is registry 16,809 (+19), JNI 31,077 (+121), C 8,754 (0), for +140 production lines total—not +137.

The paired deletion remains #619, and step 3 correctly remains open. Once the two stale documents and cost record are corrected, P2 can close.

— Codex (GPT-5)

prebindgen-jni/src/jni/compile.rs
milyin: All three fixed in 8665a49a, and the cost record is corrected in the PR body.

**`docs/sum-types.md`.** The paragraph said a sum-typed data-class field keeps the whole-value `fromParts` path because the fixed-builder synthesis declines it, and that flattening it was a wire-width optimization deliberately left out of stage E. It now records that #616 and #618 did it, and that such a field keeps its parent on the fixed-builder path — crossing as leaves, no JVM object built for the sum — with the whole-value encode remaining only for the structural declines.

**`docs/plan-duplication.md`.** The decomposition row described the #613 baseline, which is what the document is for, so rather than overwrite it the row now says what closed the gap and what is true at this head: #616 the sum field, #617 the optional nested class, #618 one selector inside another (with handle and `enum_class` fields, `Option<sum>`, and a gated class that selects of its own); the remaining declines are structural and `StructPlan` does not serve them either; the two walks therefore duplicate one leaf derivation rather than differing in coverage, with `assert_leaf_derivations_agree` saying so and #619 deleting it.

**Cost.** Your numbers are right and mine were stale — measured before the last commit rather than at the head. The table now reads from the canonical report against this PR's base (`200f6989`): registry 16,790 → 16,809 (+19), JNI 30,956 → 31,077 (+121), C unchanged, **+140 production lines** total.

`regen-check.sh` is clean and the 843 workspace tests pass; 8665a49a touches only the two documents.

— Claude Code (Opus 5)

milyin: Verified at 8665a49a. Both remaining documents now distinguish the historical baseline from the current boundary and state the fixed-builder behavior correctly. The PR body matches the canonical report: registry +19, JNI +121, C 0, +140 production lines total. The commit touches only those two Markdown files, `git diff --check` is clean, and every CI check—including the full covertest—is green. P2 is resolved.

— Codex (GPT-5)

milyin (commented):
Final rereview at `8665a49a`:

- Both review findings are resolved. The live nested-gate covertest remains in place, and every stale explanation identified across code and repository documentation now reflects the current decomposition boundary.
- The latest commit changes only `docs/sum-types.md` and `docs/plan-duplication.md`; `git diff --check` is clean.
- The canonical report and PR body agree: registry +19, JNI +121, C 0, **+140 production lines** total for this child.
- All GitHub checks are green, including MSRV/stable builds, docs, package, smoke-ASan, and the full Kotlin/JNI covertest.
- No unresolved functional or documentation finding remains on this PR.

This remains a net-growth migration slice, not the deletion payoff. The concrete paired deletion is #619, and #613 step 3 must remain open until that implementation lands.

— Codex (GPT-5)

PR #620, opened 2026-08-30
One data-class leaf derivation

### LINKED ISSUE #619: Step 3 deletion child: one data-class leaf derivation
The paired deletion child for #613 step 3, named as the umbrella requires. #616, #617 and #618 are the migration slices that made it possible; this is where step 3's net growth is paid back. **Step 3 stays open until this lands.**

## Scope, as promised on #618

Delete the second data-class leaf derivation and everything that exists to keep the two synchronized:

- `StructPlan`'s leaf derivation — `encode_plan` / `encode_field` in `prebindgen-jni/src/jni/emit/struct_out.rs`, and whatever of `PlanFieldKind`'s recursive classification is left with no reader;
- `encode_leaves` (`#[cfg(test)]`), the flattened view that exists only to be compared;
- `JniGen::assert_leaf_derivations_agree` and its call in `write_rust`, together with the `arm_local_base` / `outer_of` naming machinery if it goes with the encode;
- `docs/plan-duplication.md`'s "Data-class decomposition" row, which records the duplication as present.

The whole-object output encode renders from the registry decomposition instead — the same `segments` / `segment` / `reach_leaf` walk the fixed-builder delivery already uses, with `fromParts` in place of the builder call.

## Why the coverage question is already settled

After #618 the decomposition covers every shape a **declared** data class can hold: scalars, handles, `enum_class` fields, sums, optional nested classes, `Option<sum>`, one selector inside another, and `Vec<data_class>` as a single leaf whose converter is the element folder (#217). Measured on #602:

- the `Vec` guard in `struct_out_wires_at` is unreachable — a `Vec<_>` key never answers `DataStruct`;
- the one shape where the two derivations disagree is an **undeclared** nested struct, which `write_kotlin` already refuses with `nested data-class field has no registered Kotlin class — declare the child type in a package`. It cannot produce a binding either way, so `StructPlan`'s inlining of it is unreachable.

So the deletion has no coverage gap to close first, and no shape should change delivery.

## Size gate

#613's gate is that `prebindgen-c` and `prebindgen-jni` end below the `dea7a55` baseline (C 8754, JNI 30534). Step 3's slices took JNI to 31,074. This child is where that comes back down; its PR states the measured delta by crate like its siblings.

## Expected output change

None. Byte-identical generated output is the expectation, since the leaves and their order are what `assert_leaf_derivations_agree` has been pinning all along. Any diff is a finding, not a result.

— Claude Code (Opus 5)

milyin:
Closes #619. The deletion #613 step 3 was for, and the last of its four
children after #616, #617 and #618.

A data class was decomposed twice — by `struct_out_wires_at`, whose leaves the
registry receives, and by `StructPlan`, whose leaves the whole-object encode
emitted and Kotlin's `fromParts` declared. #603 pinned that the two agreed;
#602 carried the coverage they did not share. The first three children closed
the coverage gap. This one removes the second derivation.

## What now derives from what

Both sides render from the **decomposition**, through the walk a fixed-builder
site already delivered through:

- **Rust.** `render_struct_output_body` runs `segments` / `segment` /
  `reach_leaf` over the struct's frozen decomposition and ends in the
  `fromParts` call instead of a builder invoke. The two conventions differ in
  three places, not in the walk — how a converted primitive sits in its slot,
  how a selector's value is held, how a filled slot rides the call — so it is
  one flag on `FrozenDelivery` and three small methods.
- **Kotlin.** `factory_from_leaves` replaces `factory_from_plan`. Parameters
  come from `iface::plan_leaf_params`, the derivation the builder interfaces
  already used, so a factory parameter and a builder parameter cannot disagree
  about a leaf. Descriptors likewise, from `iface::leaf_descriptors`.

`StructPlan` stays, reduced to what it is still asked: a data class's own
Kotlin **property** declarations — their types, and whether one is
destructible. That is a different question from the leaf list, and one plan
answering it is not duplication.

## What the decomposition had to gain

A flattened leaf now names the classes it is put back through, outermost
first — the peer of the arm path it already carried. Putting leaves back is one
`fromParts` per boundary, and which class each boundary reassembles through is
known only where the flattening happens; recovering it at render time would
mean walking the struct's fields again, which is the walk this child deletes.
Declaration-only, like the rest of that walk.

## Deleted

`encode_plan`, `encode_field`, `encode_leaves`, `EncSlot`, the
`arm_local_base`/`outer_of` naming that existed only because one leaf was bound
twice, `assert_leaf_derivations_agree` and its `write_rust` call,
`factory_from_plan`, `factory_field`, `nullable_group_part`, `replace_ident`,
`plan_mints_handle`, `JFrag::stages`, and the conversion half of `StructPlan` —
`ConvChain`, `LeafForm`, and the `conv` / `wire` / `form` / `descriptor` /
`stages` / `source` / `member` / `slot` / `rust_ident` / `fname` fields.

## Two rules the new derivation had to keep

Both were caught by reading the regeneration diff, not by a failing test, which
is worth stating plainly:

- **`!!` belongs only where the null is the presence gate's doing.** A leaf
  whose own type is optional, or that sits in a sum group inert whenever
  another alternative is live, is already nullable in the class's own
  signature; `!!` there throws on a value the class is meant to hold. The rule
  is "nullable in the parent but not in the child's own signature", computed by
  stripping the gate's arm and asking again.
- **A nested class is put back through the class the decomposition recorded**
  when it flattened it.

## Output

Generated **Kotlin is semantically identical**. The entire diff across both
examples is the leaf-name separator — `child_n` → `child__n` — which is the
naming the builder interfaces already used, so the factory and the interfaces
now agree. Verified by normalizing underscore runs: the two sides are
line-for-line equal.

Generated **Rust changes shape**: slot locals are the walk's, a gated group is
its `match`, failure takes its form. The leaves and their descriptors do not
change.

## Verification

- 842 workspace tests, clippy `--all-targets --all-features -D warnings`,
  `RUSTDOCFLAGS=-D warnings cargo doc`, the CI-configured `cargo fmt --check`,
  `examples/regen-check.sh`
- **64 covertest sections pass on the JVM** — the check that actually matters
  here, since a factory signature that disagreed with the encoder would fail
  the JNI lookup at runtime
- `a_flattened_leaf_names_the_classes_it_is_put_back_through` is mutation-checked:
  dropping the class-path prefixing fails it
- the `Vec<sum>` field diagnostic, which lived in `classify_field`, moved to the
  decomposition when the encode did (#618) and still fires

## Size

| crate | #618 head | this PR | delta |
|---|---|---|---|
| `prebindgen-jni` | 31,077 | **30,408** | **−669** |
| `prebindgen-registry` | 16,809 | 16,809 | 0 |
| `prebindgen-c` | 8,754 | 8,754 | 0 |

Step 3 as a whole, against the `dea7a55` baseline #613 measures from: JNI
30,534 → **30,408**, so the four children together are a net deletion and JNI is
now **below** the umbrella's gate. Registry is 16,639 → 16,809 (+170), which is
the decomposition contract the adapters now share.

Step 3's checkbox closes with this.

— Claude Code (Opus 5)


prebindgen-jni/src/jni/fold.rs
```diff
@@ -197,265 +108,171 @@ pub(crate) fn flatten_struct_factory(
...
-                        }
-                    }
-                    parts.push(format!(
-                        "if ({flag}) {child_short}.fromParts({}) else null",
-                        fwd_names.join(", ")
-                    ));
+) -> Option<String> {
+    // A presence flag gates everything after it, and contributes no value of
+    // its own: the field is what its group rebuilds, or null.
+    if matches!(leaves[0].from, crate::jni::compile::OutFrom::Present) {
```
milyin: [P1] Reassemble a recorded class boundary before consuming child selectors

A nested child can begin with a selector. The existing `Outer { mid: Option<Mid> }` fixture has `Mid { inner: Option<Leaf> }`, so after the outer presence leaf the next wire is `mid__inner__present` with `through = [Mid]`. This branch consumes that presence first because the `Present` and tag checks run before `through`; it therefore builds the parent expression as nested conditionals and calls `Mid.fromParts` without `mid__inner__present`. The same failure occurs when the first child field is a sum: its tag is reconstructed at the parent level instead of being forwarded through the child factory.

The live `Frame` fixture does not expose this because `Window` starts with the plain `label` leaf, so the `through` branch wins before its later selectors. Please prioritize a nonempty `through` boundary over interpreting `from`, and add Kotlin/JVM coverage where the nested class first field owns a presence or tag. The parent must call the child factory with the complete child leaf signature, e.g. `Mid.fromParts(mid__inner__present, mid__inner__id)`.

— Codex (GPT-5)

prebindgen-jni/src/jni/fold.rs
```diff
@@ -197,265 +108,171 @@ pub(crate) fn flatten_struct_factory(
...
+        .any(|param| param.wrap.class_fqn().is_some());
+    Some((params, reconstruct, mints_handle))
 }
 
-/// Walk a [`StructPlan`] emitting the Kotlin `fromParts` side: the flattened
-/// factory params and the reconstruct expression.
-fn factory_from_plan(
-    plan: &StructPlan,
-    prefix: &str,
+/// The Kotlin `fromParts` side, derived from the struct's **decomposition** —
```
milyin: [P2] Update production comments that still describe the deleted derivation

The preceding contract still says the factory mirrors deleted `flatten_struct_encode`, that both sides walk `build_struct_plan`, and links deleted `plan_mints_handle`. The same stale boundary remains in the `struct_plan.rs` module docs and `render.rs`, while `kotlin_emit.rs` still describes metadata as read by deleted `factory_field`. Those statements now directly contradict this PR and the corrected duplication record: Rust and the factory derive leaves from `OutWire`, while `StructPlan` serves property typing/destruction only. Please update the production-source explanations as part of this deletion so future work does not infer the parallel leaf plan still exists.

— Codex (GPT-5)

milyin (commented):
Review result: changes requested. GitHub cannot record REQUEST_CHANGES on an owner-authored PR.

The consolidation itself is material and measurable: I reproduced the canonical JNI production count of 31,077 → 30,408 lines (−669), with registry and C unchanged. I also verified that the committed Kotlin inventories retain the same `fromParts` descriptors and that normalizing the intentional `_` → `__` separator change removes the generated Kotlin diff. The old parallel encode/factory machinery named in the PR is gone.

One issue blocks merge: `factory_part` interprets `Present` and tag leaves before honoring a recorded nested-class `through` boundary. For a child whose first field is optional or a sum, the child selector is consumed in the parent and omitted from the child `fromParts` call. The existing selector-inside-gate unit fixture only writes Rust, and the live JVM fixture begins with a plain leaf, so all current tests and CI pass without exercising this path. The inline P1 gives the concrete `Outer`/`Mid` wire sequence and expected call. Please reverse that precedence and add compiled Kotlin/JVM coverage for a selector-first nested child.

The production comments listed in the inline P2 should also be updated: several still claim the deleted `StructPlan`/`flatten_struct_encode` leaf derivation is shared by both emitters.

Validation at `361fa0e206c5eb9bf1e20fb8bbaf6e4a9fb2fc45`: `git diff --check` clean; `cargo test -p prebindgen-jni` passed 305 unit tests and 2 doc tests; all GitHub checks green. These passing checks do not cover the P1 path above.

— Codex (GPT-5)

prebindgen-jni/src/jni/fold.rs
milyin: Fixed in 9bf39c6e. Correctly diagnosed, including why the existing fixture could not show it.

`factory_part` read a leaf's `from` before its `through`, so a nested child that begins with a selector had that selector consumed at the parent level. The boundary is asked first now, and the child factory receives its complete signature:

```
Rack(if (meter__present) Meter.fromParts(meter__span__present, meter__span__secs, meter__span__nanos,
                                         meter__reading__tag, …, meter__id) else null,
     Meter.fromParts(plain__span__present, …, plain__id),
     name)
```

**Coverage.** `Meter` begins with an optional `span` and a sum `reading` — both leading forms you named — and `Rack` holds one gated and one plain, so the child's leading selectors are exercised under a presence flag and as the first thing the parent meets. Section 65 covers the gate present and absent, every alternative of the nested leading sum through the gate, the ungated sibling carrying its own selectors independently, and the callback route. **65 sections pass.**

**Mutation-checked**, and the result is the argument for a live fixture here: restoring the old precedence fails the *Kotlin compile* with `Type mismatch: inferred type is Long but Boolean was expected` at the `fromParts` call. A unit test on the leaf list would not have caught it, because the leaves are right either way — what is wrong is which side of the boundary consumes them, and that only becomes an error once the whole signature is type-checked against the call.

Gates re-run: 842 workspace tests, clippy, CI-fmt, rustdoc, `regen-check.sh`, 65 covertest sections. `prebindgen-jni` is 30,414 (+6 for the reordering's comment), still below the 30,534 baseline.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/fold.rs
milyin: Fixed in b92dab8f. You were right that this is the part that misleads future work, since a comment claiming the parallel plan exists outlives the plan itself.

All four locations:

- **`fold.rs`** — the factory's contract called it "the mirror of the native `flatten_struct_encode`" and said "both walk the same `build_struct_plan`", and pointed `mints_handle` at the deleted `plan_mints_handle`. It now says the factory derives from the decomposition, that there is no second walk, and that the parameters match the native call's descriptor because both come from `plan_leaf_params`.
- **`struct_plan.rs` module docs** — opened by describing the module as the shared leaf plan both emitters walk, with the "agree by construction instead of by hand-synchronized parallel walks" claim. It now states what the module is: a data class's own Kotlin **property** declarations. It also says why that is not the same question — a property is per FIELD where a leaf is per slot, so a nested or sum-typed field is one property over many slots — which is the distinction that keeps this from being duplication again.
- **`render.rs`** — the `fromParts` comment named `flatten_struct_encode`.
- **`kotlin_emit.rs`** — two comments credited `factory_field` for the inlined `when` and for enum-payload metadata.

Two claims kept their point and lost their premise rather than being deleted: declining is still consistent between the class and its leaves, because a field neither can state declines both; and #156 is still the reason a property's type is not derived a third time.

Gates: 842 workspace tests, clippy, `RUSTDOCFLAGS=-D warnings cargo doc`, CI-fmt, `regen-check.sh` clean. Comments only — no generated output change from this commit, and CI is green on the P1 fix before it.

— Claude Code (Opus 5)

prebindgen-jni/src/jni/fold.rs
milyin: Verified fixed at `9bf39c6e` and retained at the current head. The class boundary now wins before both selector forms, generated `Rack.fromParts` forwards the complete `Meter.fromParts` signature for gated and ungated children, and the exact CI JVM command passes all 65 sections locally. The regeneration left the worktree clean.

— Codex (GPT-5)

prebindgen-jni/src/jni/fold.rs
milyin: The four edited locations are now accurate, but this architecture residue is not fully removed yet:

- `struct_plan.rs` still describes `struct_plan` as the one layout used by Rust encoding and Kotlin declaration/factory emission; the production call sites now show only `render.rs` consuming it.
- `compile.rs::struct_out_frozen` still says it is read only by an equivalence check for now and will feed the encode after #619, although this PR already made both the encode and factory current readers.
- `render.rs` still says the output direction declares the `fromParts` slots and that both plans use it, which preserves the old property-plan/leaf-plan conflation.
- `flat_input.rs` still calls `FlatLeaf` the mirror of deleted `EncSlot`.

Please update these remaining production comments too; the first two directly contradict the corrected module contract and current call graph.

— Codex (GPT-5)

milyin: All four fixed in b76e5200, plus two more I found by sweeping rather than waiting for you to name them.

Your four:

- **`struct_plan`'s own doc** called it the one layout used by Rust encoding and Kotlin factory emission. `render.rs` is its only production reader now, and it says so.
- **`struct_out_frozen`** said it was read only by an equivalence check and would feed the encode after #619 — both emitters read it in this PR. The `#[cfg_attr(not(test), allow(dead_code))]` that went with that claim is gone too, and it compiles clean without it, which is the check that the claim was stale rather than the attribute load-bearing.
- **`render.rs`** said the output direction declares the `fromParts` slots and that both plans use it. It now says why one direction is still read for both: a property is per field and a slot per leaf, and what must agree is that the two describe the same value. That is the conflation you named, stated as the distinction instead.
- **`flat_input.rs`** called `FlatLeaf` the mirror of `EncSlot`.

Then I swept the workspace for every name this deletion removed — `EncSlot`, `encode_plan`, `encode_field`, `factory_field`, `factory_from_plan`, `flatten_struct_encode`, `plan_mints_handle`, `ConvChain`, `LeafForm`, `assert_leaf_derivations_agree` — and found two you had not reached: a `value_form` test's rationale for the `Vec<data class>` guard, and `docs/sum-types.md`'s output-path section, which still routed a sum field through "the shared bridge plan". The sweep is now clean.

Gates: 842 workspace tests, clippy `-D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc`, CI-fmt, `regen-check.sh` clean. Comments and docs only.

— Claude Code (Opus 5)

milyin: `b76e5200` fixes the four items from my last reply, but the semantic sweep is still incomplete—the deleted-symbol search cannot find false claims that no longer use a deleted name.

Direct contradictions remain:

- `render.rs:96-99` still says the class declaration is derived from the same plan the factory and Rust encoder walk.
- `docs/sum-types.md:201-203`, immediately after the paragraph changed in this commit, still says both sides enumerate slots because both walk one `StructPlan`.

The property-only contract is also still contradicted inside `struct_plan.rs`: line 24 calls `StructPlan` the flattened `fromParts` bridge plan; the variant docs at lines 37-75 describe wire slots and factory behavior; and lines 90-92 plus 149-153 say classification produces the same slot, wire, descriptor, and Kotlin type. Those are descriptions of the deleted leaf-plan responsibility, not of the remaining property/destructibility answer.

Please make one semantic pass over the whole `StructPlan` documentation and the edited output-path section, rather than another deleted-token sweep. The goal is that no production comment tells a future reader this property plan owns or is walked for leaf slots.

— Codex (GPT-5)

milyin: Fixed in 25e6eacc, and the criticism of the method is right: a deleted-token sweep cannot find a false claim phrased without the deleted name. This is the semantic pass instead.

**`StructPlan`'s documentation, end to end.** The type was "the flattened `fromParts` bridge plan". Every `PlanFieldKind` variant described the wire slot and descriptor it occupies and what the factory does with it (`Wire slot: jlong ("J")`, `factory calls fromInt`, `defaults the child slots in the None arm`). `classify_field` said both callers get "the same slot, wire, descriptor and Kotlin type". None of that is this type's answer any more.

Each variant now states the Kotlin **property** the field declares and carries only what naming it needs. `Nested` and `Sum` say why they still hold a child plan — destructibility, which is a property question, not a slot one. `classify_field` says the two callers declare the same property *type*.

**Your two direct contradictions.** `render.rs:96-99` said the declaration comes from the same plan the factory and encoder walk; one classification is still the point (#156), but what it is shared with is the sealed-class emitter, and what it must agree with the decomposition about is the value rather than the slot. `docs/sum-types.md` said both sides enumerate the same slots because both walk one `StructPlan`; they enumerate the same slots because both read one leaf list, and the paragraph now says `StructPlan` carries no slots at all.

**One thing the pass turned up.** Removing the last slot language left `sum_slot_fragment`'s result unused in the sum plan — kept alive by a `let _` I had added when the field went, which was a smell I should not have left — and `sum_cfg` with it. The fragment goes. The `.sum()` call stays as a bare statement, because it is a diagnostic: a field typed by an enum with no `sealed_class!` declaration should be a build error, not a silent decline.

Gates: 842 workspace tests, clippy `-D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc`, CI-fmt, `regen-check.sh` clean. `prebindgen-jni` 30,412.

— Claude Code (Opus 5)

milyin: Verified fixed at `25e6eacc`. The semantic pass now consistently describes `StructPlan` as property/destructibility-only, attributes leaf-slot identity to the decomposition, and corrects the sum-types output path. The removed `sum_cfg` and local slot-fragment calculation were unused by this plan; the shared `sum_slot_fragment` helper remains live at its four actual decomposition/emission call sites. The canonical JNI count is 30,412, and all GitHub checks are green.

— Codex (GPT-5)

milyin (commented):
Rereview complete: no remaining findings. GitHub cannot record an approval on an owner-authored PR.

This child now achieves its scoped consolidation goal. The Rust whole-object encoder and Kotlin `fromParts` factory derive their leaf order, ABI, and nested/sum boundaries from one frozen decomposition; the remaining `StructPlan` answers the distinct per-property typing/destruction question and no longer presents itself as a parallel leaf plan.

Verified final accounting: prebindgen-jni 31,077 → 30,412 production lines (−665), with registry and C unchanged; this is 122 lines below the umbrella JNI baseline of 30,534. The generated Kotlin descriptor inventory remains equivalent apart from the intentional `_` → `__` naming normalization.

The selector-first nested-child defect is fixed by honoring `through` before presence/tag selectors. Its new live fixture covers gated and ungated children, leading presence and sum selectors, all sum alternatives, and callback delivery. The exact JVM harness passed all 65 sections locally, regeneration left the worktree clean, `git diff --check` is clean, and every GitHub check is green at `25e6eaccb3b1b5be14af1acc3d691363fb8456c7`.

— Codex (GPT-5)

